PR #87, opened 2013-02-26
New Plugin: numconvert

kugel-:
The new plugin is probably not useful for 1.23, but the multiterm fix might be.


frlan:
After face2face discussion with @kugel-  I'm going to close this PR.

PR #89, opened 2013-03-04
Markdown: Remove embedded Discount lib and other stuff

codebrainz:
This pull requests removes the embedded [Discount markdown library](http://www.pell.portland.or.us/~orc/Code/discount/) code because its license is questionably a violation of the GPL when it is being distributed as it was in the Markdown plugin.

To replace it, the [peg-markdown](https://github.com/jgm/peg-markdown) library is embedded into the Markdown plugin source code for when Discount is not available on the system. The default (for Autotools) is to look for Discount library and header and use it if available otherwise fall back to the bundled peg-markdown library. The bundled library can also be selected by using the `--enable-peg-markdown` option for the `configure` script. The peg-markdown library is quite a bit slower for this purpose of real-time Markdown->HTML on key-press, which is why Discount is still used by default. The peg-markdown library is also not widely available in Linux distros, which is why it's embedded rather than just using system one as a fallback.

I have no idea how to use Waf, so (hopefully) it just checks for Discount library and fails to build it if not available. It makes no attempt to fallback to the bundled peg-markdown library if Discount is not found.

Although not tested, when linked against either library it should build/run fine on Windows but like Webhelper and Devhelp, depends on WebKit, so probably not going to be in the Win32 release. That's fine.

Please test and merge if possible.

PS. Sorry for requesting this pull so close to release, it's kind of important in order to avoid any GPL violations in Geany-Plugins releases.

Thanks

### DIFF
--- a/markdown/wscript_build
+++ b/markdown/wscript_build
@@ -24,31 +24,12 @@ from build.wafutils import build_plugin
 from waflib.Utils import subst_vars
 
 name = 'Markdown'
-includes = [ 'src', 'discount' ]
+includes = [ 'src' ]
 defines = [ subst_vars('MARKDOWN_DATA_DIR="${PKGDATADIR}/markdown"', bld.env),
             subst_vars('MARKDOWN_DOC_DIR="${DOCDIR}/markdown"', bld.env),
             subst_vars('MARKDOWN_HELP_FILE="${DOCDIR}/markdown/html/help.html"', bld.env) ]
-libraries = [ 'GTK', 'GTHREAD', 'WEBKIT' ]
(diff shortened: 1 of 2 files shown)


GrantM11235:
Since version 2.1.6, Discount has been released under the 3-Clause BSD License which [is GPL compatible](https://www.gnu.org/licenses/license-list.en.html#ModifiedBSD)

[2.1.6 release notes](http://www.pell.portland.or.us/~orc/Code/discount/#v2.1.6)

[Current Discount COPYRIGHT file](https://github.com/Orc/discount/blob/master/COPYRIGHT)

PR #92, opened 2013-03-20
Devhelp: Add support for GTK3

codebrainz:
This PR includes Autotools build system adjustments as well as new libdevhelp API and other miscellaneous deprecated or obsolete GTK3(-stack) changes.

When using GTK+3 finally we can depend on Debian-based (and other) distros providing a needed libdevhelp again so don't use the embedded version, rather using the system's `libdevhelp-3.0` when compiling against GTK3 Geany.

TODO: Waf build system support for GTK3

Please test, improve and/or approve.


frlan:
@codebrainz  I will close the MR as it was open for quiet some time and it's not mergeable anyway.

PR #95, opened 2013-03-20
MultiTerm: Add support for GTK3

codebrainz:
This PR adds support for GTK3 by adjusting the build system and updating the GTK3(-stack) API used.

It includes an ugly hack for the `geany.vapi` and `geany.deps` files by simply copying them to new names, replacing and/or commenting-out some incompatible stuff and using the appropriate one decided at build time. I think it's OK for now since we'll be improving/sharing Vala bindings support for plugins soon enough and this part won't be needed once the GTK3 support is done correctly in the future shared bindings.

TODO: Waf build system support for GTK3 MultiTerm plugin.

### DIFF
--- a/multiterm/src/plugin.vala
+++ b/multiterm/src/plugin.vala
@@ -109,4 +109,6 @@ public void plugin_cleanup ()
 {
 	foreach (Widget wid in toplevel_widgets)
 		wid.destroy();
+
+	toplevel_widgets = null;
 }


LarsGit223:
@codebrainz: how about finishing this gtk3 stuff?

codebrainz:
@LarsGit223 eventually... I don't use or like GTK+3 so it hasn't been a priority.

If you want to work on it, I'd be more than happy to add you as a contributor on my GP fork so you can push to it.

LarsGit223:
@codebrainz: I would prefer to team up with someone, helping with the gtk3 porting. All the warnings on compiling for gtk3 are quite annoying. I ported scope some months ago and already got new gtk3 deprecation warnings. Also, this PR looks pretty complete - what is missing?

LarsGit223:
@codebrainz: ok, I changed my mind. Add me and I will see what I can do.

codebrainz:
@LarsGit223 done, I also checked the "Allow edits by maintainers" which I think does something like allowing you to push to this PR or something like that.

LarsGit223:
@codebrainz: ok, thanks.

multiterm/src/context-menu.vala
```diff
@@ -68,6 +68,20 @@ namespace MultiTerm
 			item.show();
 		}
 
+		private Gtk.MenuItem add_image_menu_item(string label_text, string icon_name)
```
codebrainz: I guess this should be guarded with `#if MULTITERM_GTK3` too? IIRC Vala compiler warns about unused private functions.

multiterm/src/notebook.vala
```diff
@@ -149,7 +153,11 @@ namespace MultiTerm
 				context_menu.new_window_activate.connect(on_new_window_activate);
 				context_menu.move_to_location_activate.connect(on_move_to_location);
 			}
+#if MULTITERM_GTK3
+			context_menu.popup_at_pointer(event);
```
codebrainz: This requires GTK+ 3.22 but I think the build system only checks for GTK+ >= 3.0, so if someone built it with any version of GTK+ between 3.0 and 3.22, it would cause a build error.

multiterm/src/context-menu.vala
LarsGit223: Done. Even if not it shouldn't be callable on gtk2.

multiterm/src/notebook.vala
LarsGit223: Can I use normal gtk C-style version checks in valac?

LarsGit223:
@codebrainz: Hmmm...I can now at least build your repo but if I open the plugin manager I do not see the Multiterm plugin (already did a ```ldconfig /usr/local/lib```).

multiterm/src/context-menu.vala
codebrainz: Yeah, just prevents some warnings from the Vala compiler because it would be unused.

multiterm/src/notebook.vala
codebrainz: Not for compile-time versions (ie. `GTK_CHECK_VERSION` macros), Vala's preprocessor is too wimpy. You can pass along a macro from the build system using `-D` option if it's really needed.

multiterm/src/plugin.vala
```diff
@@ -53,6 +53,8 @@ public void plugin_init(Geany.Data data)
 	Alignment align;
 	MultiTerm.Notebook notebook;
 
+	toplevel_widgets = new List<Widget>();
```
codebrainz: This probably needs to be assigned `null` in the cleanup function to avoid a leak in the GC and so there's no dangling junk left in it.

LarsGit223: Done.

multiterm/src/notebook.vala
LarsGit223: Ok, I tried defining something like ```MULTITERM_GTK_3_22``` but failed at changing the makefiles properly. So I leave this as is for now.

PR #163, opened 2014-10-15
webhelper: remove generated files from version control

techee:
These two get always regenerated on build on my machine with changes in the comments and are marked as modified by git, which is annoying.

Unless I miss anything, as they are regenerated, they don't have to be tracked by VCS.


b4n:
I found a simpler fix: just don't put full paths in the generated output.  It's not perfect in theory, but in practice it should fix all possible issues arising from this (but if *glib-mkenums*' output changes, but it's quite unlikely).

PR #164, opened 2014-10-16
Less recursive Autotools

b4n:
**_Do NOT merge this**_ yet.  This at least doesn't port all plugins.

---

I wanted to use the `subdir-objects` Automake option in the GP build system, because some plugins build sources in sub-directories (_debugger_, _scope_, and _geanyvc_ and _geanyprj_ in their unit tests), and because it seems to be the future of Automake (e.g. someday it'll be impossible not to use it).  Also, it's pretty neat (in theory).

The problem is that it has [a bug](http://debbugs.gnu.org/cgi/bugreport.cgi?bug=17354) when a subdir object is in a directory that also builds objects.  This is problematic for us in the case of [_geanyprj_](/geany/geany-plugins/blob/master/geanyprj/tests/Makefile.am#L5) and [_geanyvc_](/geany/geany-plugins/blob/master/geanyvc/tests/Makefile.am#L5) usage.

Now, there are three options:
1. don't use `subdir-objects`, and deal with it the day it becomes unavoidable (maybe hoping the aforementioned bug gets fixed by then).
2. don't _build_ the subdir objects in a directory that already builds objects.  This basically mean we have to `#include "../src/utils.c"` in _geanyprj_ and _geanyvc_ unit tests -- not the worst thing ever in a unit test, but not the best either.
3. build all the objects in a particular directory from the same Makefile -- in practice, make things less recursive.

This PR gives a shot at option 3, and basically merges each plugin's _Makefile.am_-s together in _[plugin]/Makefile.am_.  This leans towards the ["non-recursive Automake"](https://autotools.io/automake/nonrecursive.html) pattern some people advertize.

Pros I can think of:
- Works around the aforementioned Automake bug so allows us to use `subdir-objects` -- but that's not so important.
- Less recursion is supposed to be a little faster.
- All rules for a particular plugin (source, data, docs, …) are in the same place: more concise, provides an overview and ease modifications that touch several aspects at once.

Cons:
- All rules for a particular plugin (source, data, docs, …) are in the same place: more bloated and not split by aspect.

Of course, to simply allow us to use `subdir-objects`, only _geanyprj_ and _geanyvc_ build systems have to be tuned, but consistency over GP is good.

Enough chattering, let's see: what do you guys thing about this?  Does what this PR start look like a good path to you or do you think it's a bad idea? @sardemff7 @hyperair @frlan @codebrainz @ntrel @elextr?


kugel-:
https://github.com/geany/geany-plugins/pull/277 implements an IMO superior solution to less recursive automake (using %reldir% as mentioned here before, e.g. in https://github.com/geany/geany-plugins/pull/164#issuecomment-59335147)

kugel-:
@b4n are you still working on this?

b4n:
@kugel- Not now, but I wish.  Actually I'm not working on much Geany/GP stuff these days :(

frlan:
@b4n What we are going to do with this WiP-PR?

PR #172, opened 2014-10-21
Enable Travis CI

b4n:
***This has to be discussed a little before merging***

Here it is a bit more complex than [on Geany](/geany/geany/pull/360), but it's also probably more important.  Same concern applies than on Geany, so I won't repeat myself here and concentrate on technical questions.

1. We need a recent Geany (we can't simply install the Ubuntu 12.04 package).  Here I made the setup rules build it from latest Git, but it might not be so great:
  1. Any build issue in Geany git will break GP.
  2. It means we rebuild from source each time.  Maybe a better solution would be to [install](http://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-from-a-custom-APT-repository) from the [nightly builds .deb](http://nightly.geany.org/debian/)s but unfortunately the [Wheezy ones aren't compatible with Ubuntu 12.04](https://travis-ci.org/b4n/geany-plugins/jobs/38556973#L43).
2. As you can see with the additional commit, I couldn't build GeanyPy because [Ubuntu 12.04 don't seem to have libpython-dev](http://packages.ubuntu.com/search?keywords=libpython-dev&searchon=names&suite=precise&section=all), which is quite weird.
3. As you can also see, I had [a fairly weird](https://travis-ci.org/b4n/geany-plugins/jobs/38544571#L2945) problem with cppcheck that I can't reproduce on my machine (which has cppcheck 1.66).  If anyone can tell me what's going on here, please go ahead (especially as other cppcheck calls worked fine, see above in the log).

Finally, you might see that the [GTK2 build currently fails](https://travis-ci.org/b4n/geany-plugins/builds/38554414), but that's legitimate, see #171 (so that's a good point for Travis CI as I found the error when setting it up :).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,61 @@
+language: c
+compiler:
+  - gcc
+env:
+  - GTK3=no
+  - GTK3=yes
+before_install:
+  - sudo apt-get update -qq
+install:
+  - sudo apt-get install -y libgtk2.0-dev libgtk-3-dev intltool libtool
+  - sudo apt-get install -y python-docutils
+  - sudo apt-get install -y cppcheck check
+  # debugger
+  - sudo apt-get install -y libvte-dev
+  # devhelp
+  - sudo apt-get install -y libwebkitgtk-dev libwnck-dev libgconf2-dev zlib1g-dev
+  # geanygendoc
+  - sudo apt-get install -y libctpl-dev
+  # geanylua
+  - sudo apt-get install -y liblua5.1-0-dev
+  # geanypg
+  - sudo apt-get install -y libgpgme11-dev
+  # geanypy
+  - sudo apt-get install -y libpython-dev
+  # geanyvc
+  - sudo apt-get install -y libgtkspell-dev
+  # geaniuspaste/updatechecker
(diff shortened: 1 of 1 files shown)


eht16:
To wake this one up: regarding the question how to get an up2date version of Geany into the build environment: I'd prefer the nightly builds way.
In the meantime, Travis offers Ubuntu 14.04 as beta environment (https://docs.travis-ci.com/user/ci-environment). Maybe we can check this, it might also help solving the geanypy problems.
The Debian Wheezy nightly builds install fine in Ubuntu 14.04.
Alternatively, we could also setup a nightly build for Ubuntu 12.04. However, this seems a bit overkill just for Travis because I guess normal users have already upgraded their Ubuntus to more recent versions.

And in theory the nightly builds approach has the disadvantage if we change some API in Geany and on the same day use this new API in G-P, the nightly builds will be too old.
Though I think this won't happen too often and/or we could maybe shorten the nightly interval from 24h to 12h hours or even less (they won't rebuild if there were no changes in GIT master).

b4n:
Obsoleted by #394

PR #174, opened 2014-10-22
debugger: Fix crash with GDB 7.7

b4n:
GDB 7.7 doesn't understand `&` as an argument to `-exec-run`.

Moreover, the code handling errors from GDB has an issue with this
error and leads to a crash.

Closes https://sourceforge.net/p/geany-plugins/bugs/112/

---
@cesspit Are you OK with this?  I think it's important to do something as it leads to a crash when using GDB 7.7 (version is according to your findings).  I also think that even if it prevents asynchronous debugging, it's still worth to fix until a proper asynchronous fix is released.

### DIFF
--- a/debugger/src/dbm_gdb.c
+++ b/debugger/src/dbm_gdb.c
@@ -369,7 +369,7 @@ static gboolean on_read_async_output(GIOChannel * src, GIOCondition cond, gpoint
 				update_files();
 
 				/* -exec-run */
-				exec_async_command("-exec-run &");
+				exec_async_command("-exec-run");
 			}
 		}
 		else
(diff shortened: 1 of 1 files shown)


frlan:
Is this ready for merging?

b4n:
@frlan as @cesspit didn't answer (I understand he's quite busy with other things), I think we can go ahead and decide to merge this.  It's easy enough to revert later if it requires a different fix at some point anyway.

b4n:
(BTW, I have other work to clean up all the GDB parsing code, which while quite a huge set of changes should avoid most possibilities of crashes in this area when GDB gives unexpected answers)

frlan:
OK. Going to merge and we will see ;)

PR #182, opened 2014-12-30
scope, debugger: use util.h instead of pty.h on OS X

techee:
pty.h doesn't exist on OS X, it's contents is present in util.h. This fixes plugins compilation problem on OS X (I tried just the basic ones with minimal dependencies).


techee:
Nobody complains, assuming it's fine to apply.

PR #188, opened 2015-02-12
geanylua: Do not accept Lua5.2 as the plugin doesn't build with it

b4n:
Make sure the found Lua package is version 5.1 and not newer, as 5.2
breaks compilation.

The issue with 5.2 was reported by an user on IRC.

### DIFF
--- a/build/geanylua.m4
+++ b/build/geanylua.m4
@@ -14,8 +14,10 @@ AC_DEFUN([GP_CHECK_GEANYLUA],
         done])
 
     LUA_VERSION=5.1
+    LUA_VERSION_BOUNDARY=5.2
     GP_CHECK_PLUGIN_DEPS([GeanyLua], [LUA],
-                         [${LUA_PKG_NAME} >= ${LUA_VERSION}])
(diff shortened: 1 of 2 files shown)


frlan:
Can you add also something to wscript?

b4n:
@frlan food remark, fixed. It works but it might be quite ugly as I don't know much about Waf (e.g. I didn't find a better way to check for `< 5.2`, but admittedly I couldn't find the canonical docs for `conf.check_cfg()`).

eht16:
pkg-config also knows --max-version but for lua-5.1 we probably would need to set --max-version 5.1.99 or something. So I actually think your current version constraint isn't that ugly.

b4n:
@eht16 yeah I found out about `max_version` but as you say it's a `<=` check, so not what we need here.
And ok, if you can't tell me there's a better way to do it there mustn't be any I guess :)

eht16:
Yes, I don't know a better solution. Though I don't agree that this is the same as there is no better solution :).

Let's merge it.

PR #192, opened 2015-02-18
Waf: Allow to explicitly set plugin name

b4n:
Also use the new feature to fix *git-changebar* plugin name and installation directory for common files.

### DIFF
--- a/build/wafutils.py
+++ b/build/wafutils.py
@@ -77,7 +77,7 @@ def add_to_env_and_define(conf, key, value, quote=False):
     conf.env[key] = value
 
 
-def build_plugin(ctx, name, sources=None, includes=None, defines=None, libraries=None, features=None):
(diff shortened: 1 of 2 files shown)


eht16:
I'm ok with this change.
Just being curious, why do you want to change the internal name from "gitchangebar" to "git-changebar"?

b4n:
> Just being curious, why do you want to change the internal name from "gitchangebar" to "git-changebar"?

Just because it's the name used with Autotools and wanted the build systems to agree, and also to some extent to match the directory name in the tree.

However, another solution to keep consistency without this would be just stripping the dash from everywhere, I don't mind either way.

eht16:
Me neither :).
Having this change might help in the future for other plugins, so why not.

b4n:
Alternatively I could also simply set the `name` to "Git-ChangeBar" and not mind the combination of CamelCase and dashes (which I actually don't) -- or event not use camel-case in the log domain.

PR #194, opened 2015-02-20
Properly test valac existence

techee:
According to the documentation

http://www.gnu.org/software/automake/manual/html_node/Vala-Support.html

$VALAC contains 'valac' when valac was not found. I kept the empty
string test just in case it is returned under some special circumstances.

This fixes autogen on OS X.

### DIFF
--- a/build/multiterm.m4
+++ b/build/multiterm.m4
@@ -7,7 +7,7 @@ AC_DEFUN([GP_CHECK_MULTITERM],
         dnl FIXME: if the C sources are present (e.g. in a release tarball),
         dnl        we don't actually need valac
         AM_PROG_VALAC([0.7.0])
-        AS_IF([test "$VALAC" = ""],
+        AS_IF([test "$VALAC" = "valac"],
               [AS_IF([test "$enable_multiterm" = "auto"],
                      [enable_multiterm=no],
                      [AC_MSG_ERROR([valac not found])])])


techee:
Merging - the previous test doesn't work on Linux either, I've just tested. As the empty string test really seems to be wrong, I've kept just the "valac" test.

b4n:
Hum, doesn't this prevent using valac when it's present?  does the macro really always give a full path when it found something?

b4n:
OK, according to the doc it does.  Though, wouldn't it be more sensible to use the third argument of `AM_PROG_VALAC` instead?

b4n:
> OK, according to the doc it does.  Though, wouldn't it be more sensible to use the third argument of `AM_PROG_VALAC` instead?

Answering myself: no, because the 2nd and 3rd parameters are new in 1.12.5 and we depend on 1.8 (and probably depending on > 1.11 is not a great idea for the moment).

b4n:
Hey, it's me again :D

So, Automake's NEWS for 1.12.5 suggest your change, while OK, will only work with valac >= 1.12.5:

>   - If no proper Vala compiler is found at configure runtime, AM_PROG_VALAC
    will set the AC_SUBST'd variable 'VALAC' to 'valac' rather than to ':'.
    This is a better default, because with it a triggered makefile rule
    invoking a Vala compilation will clearly fail with an informative error
    message like "valac: command not found", rather than silently, with
    the error possibly going unnoticed or triggering harder-to-diagnose
    fallout failures in later steps.

So possibly we should extend the test to include both values (and was it empty rather than ":"?)

techee:
Do I understand it correctly there should be

```
AS_IF([test "$VALAC" = "valac" || test "$VALAC" = ":"]
```

Is the colon there correct? OMG how much I hate Automake!

codebrainz:
I think it's just Automake's Vala support that really sucks. Might it be better to use `AM_PATH_PROG` or something and not use the buggy builtin-support? Aside: It's always really bothered me how the Vala support in Automake pollutes the source directory with built files.

b4n:
> Is the colon there correct?

I guess so, yeah.  `:` is a noop shell command.  But I will check tomorrow on a real Automake 1.11 system just to be sure.

> OMG how much I hate Automake!

Well… I won't get into the trollish rant, but know that we are trying to fight Automake's Vala support because it has a feature we consider a bug :)  Also, 1.12.5 (Nov 2012) fixes the particular issue of detecting the actual presence of the compiler my having proper 2nd and 3rd argument.

> Might it be better to use `AM_PATH_PROG` or something

I'm not sure what [`AM_PROG_VALAC`](https://www.gnu.org/software/automake/manual/automake.html#index-AM_005fPROG_005fVALAC) does in addition to locating the compiler and knowing how to check its version, so maybe yeah we could use [`AC_PATH_PROG`](https://www.gnu.org/software/automake/manual/autoconf.html#AC_005fPATH_005fPROG) either instead or in addition to `AM_PROG_VALAC`, yeah.

techee:
If anyone (Colomban, Matthew) of you knows what the "right" way is, would you please fix it for me? Sorry for adding extra work but I'm afraid whatever I ever touch in autoconf/automake goes always wrong...

b4n:
see #195

PR #195, opened 2015-02-26
Autotools Vala checks

b4n:
See #194


codebrainz:
The usage looks great to me, can't comment much on the macro definitions

b4n:
Successfully tested on Debian oldstable (Automake 1.11, no Vala), Ubutnu 14.04 (Automake 1.14, no Vala) and my Debian unstable (Automake 1.14, Vala 0.26), so it's now merged :)

PR #200, opened 2015-03-09
pairtaghighlighter: Fix clearing previous indicators

b4n:
The second argument for SCI_INDICATORCLEARRANGE is the length to clear,
not the end position.

This used to work on earlier versions of Scintilla for some reason,
probably because it didn't check length, but current versions do nothing
with an invalid length.


b4n:
Closes https://github.com/vmkononenko/pair-tag-highlighter/issues/1

PR #202, opened 2015-03-16
debugger: Safer GDB/MI parsing and other fixes

b4n:
This Introduces a proper GDB/MI parser to replace the dangerous string manipulations.  This part of the patchset is large and might have introduced issues I overlooked.  It probably requires extensive testing.
Note that I only replaced the GDB/MI parsing code with something more generic and robust, the logic is (should be) still the same.

The rest of the changes are more trivial and should be straightforward.

If needed or requested, I can split the changes and/or squash some of the commits together.  I did not do it because I'm not certain it's needed, and I prefer to do it last in case this PR leads to additional changes.

**Beware** that I'm not a real user of the debugger plugin, so I only performed limited testing -- I tried to use the most features I could find, both before and after my changes, but didn't use the plugin in any kind of daily basis.


frlan:
I will postpone this again. Now to 1.27 due to lag of review.

b4n:
@frlan yeah, thanks.  Though, in practice I'll probably just merge is as-is after 1.26, as the previous maintainer dropped and was alright with it, and I got positive response from a few people having tested it in the context of other reports.

This said, if anyone wishes to take maintenance of this plugin and/or review this and/or test this is more than welcome.  I'm not planning to maintain it for real in the long term.

frlan:
I'm fine with it. With having Devyn looking into complete debugging stack I've got a good feeling its not ending in a huge mess.

PR #203, opened 2015-03-21
webhelper: replace depricated glib (>=2.32) functions

vogon-poetic:
The glib functions `g_value_set_char` and `g_thread_init`
are depricated, throwing a GCC warning on their compilation.
This commit fixes the GCC warnings by removing the old
function (`g_thread_init`), and changes a block in gwh-plugin.c 
to use the newer (`g_value_set_schar`).

### DIFF
--- a/webhelper/src/gwh-plugin.c
+++ b/webhelper/src/gwh-plugin.c
@@ -419,11 +419,14 @@ plugin_init (GeanyData *data)
    * (g_quark_from_static_string() for example) so it's not safe to remove it */
   plugin_module_make_resident (geany_plugin);
   
-  /* webkit uses threads but don't initialize the thread system */
+  /* Deprecated function g_thread_init() is not needed. after glib 
+   * versions >= 2.32 
+  
   if (! g_thread_supported ()) {
     g_thread_init (NULL);
   }
   
+   */
   load_config ();
   gwh_keybindings_init ();
   
--- a/webhelper/src/gwh-settings.c
+++ b/webhelper/src/gwh-settings.c
@@ -196,7 +196,6 @@ gwh_settings_install_property (GwhSettings *self,
         break;
     
     HANDLE_BASIC_TYPE (BOOLEAN, Boolean, boolean)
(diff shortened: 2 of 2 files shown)


b4n:
No.  I won't drop support for GLib < 2.32 for no good reason, and I don't think these are good reasons.  If you really mind the deprecation warnings, I see two acceptable (IMO) approaches:

1. don't mind deprecation warnings, possibly using `-Wno-deprecated-declarations`.  If you build the GTK3 version of GP you'll see a *lot* of deprecation warnings, and most are just not really reasonably fixable without a *lot* of code addition just for this.

2. I could accept a patch that switch to the non-deprecated variants on 2.32 onwards (e.g. guard with `#if GLIB_CHECK_VERSION`).  I don't really think it matters, but if you really care, well, okay.

webhelper/src/gwh-plugin.c
```diff
@@ -419,11 +419,14 @@ plugin_init (GeanyData *data)
...
   
-  /* webkit uses threads but don't initialize the thread system */
+  /* Deprecated function g_thread_init() is not needed. after glib 
+   * versions >= 2.32 
+  
   if (! g_thread_supported ()) {
     g_thread_init (NULL);
   }
   
+   */
```
b4n: don't comment out stuff, just keep it or remove it.  we use a VCS anyway, it can be recovered if it really need to

webhelper/src/gwh-settings.c
```diff
@@ -211,7 +210,12 @@ gwh_settings_install_property (GwhSettings *self,
     HANDLE_BASIC_TYPE (STRING,  String,  string)
     
     #undef HANDLE_BASIC_TYPE
-    
+
+    /* remove depricated function g_value_set_char and replace w/ this. */
+    case G_TYPE_CHAR:
+    	g_value_set_schar(value, (gint8)((GParamSpecChar*)pspec)->default_value);
+        break;
+
```
b4n: Should use `HANDLE_BASIC_TYPE (CHAR, Char, schar)`

PR #204, opened 2015-03-21
MultiTerm: Minor updates of the README file

codebrainz:


multiterm/README
```diff
@@ -22,45 +22,57 @@ to compile the MultiTerm plugin you will need the development files for
...
+generates triggers a few C compiler warnings in most configurations. These
+warnings are (almost always) harmless, and are outside of our control.
+There's a bug open with the Vala compiler working on fixing this:
+
+https://bugzilla.gnome.org/show_bug.cgi?id=732531
 
-For detailed instructions, see the INSTALL file of the Geany-Plugins package
-this plugin is part of.
+If you want to hide these warnings, add the appropriate `-Wno-XXX` compiler
+flags to the ``MULTITERM_CFLAGS`` environment variable when configuring the
```
b4n: won't that conflict with the pkg-config values?  if MULTITERM_CFLAGS is a pkg-config thing I believe setting it will override the pkg-config value.

codebrainz: I dunno, I just looked at `configure --help`, would it be better to say `VALAFLAGS` or something else (and using `-Xfoo` to pass `foo` to `CC`?

codebrainz: @b4n is there something plugin specific to avoid saying to add those `-Wno-XXX` flags to `CFLAGS` env. var. and applying to all plugins?

cwendling: `--disable-extra-c-warnings` for everyone.  Nothing plugin-specific.

b4n: `--disable-extra-c-warnings` for everyone. Nothing plugin-specific.

codebrainz: Maybe I can add the `#pragma` to ignore it and ensure that file gets included in all the generated C sources?

PR #206, opened 2015-04-01
Misc build related fixes

kugel-:
Needed to compile against Geany's linkage-cleanup branch but they
are good changes regardless.

### DIFF
--- a/codenav/src/utils.c
+++ b/codenav/src/utils.c
@@ -19,6 +19,8 @@
  *  along with this program; if not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <geanyplugin.h>
+
 #include "utils.h"
 
  /**
@@ -107,4 +109,3 @@ strrpos(const gchar *haystack, const gchar *needle)
 	  return p - haystack;
    return -1;   // not found
 }
-
--- a/devhelp/src/Makefile.am
+++ b/devhelp/src/Makefile.am
@@ -25,6 +25,7 @@ devhelp_la_CFLAGS = \
 
 devhelp_la_LIBADD = \
 	$(DEVHELP_LIBS) \
+	$(COMMONLIBS) \
 	$(top_builddir)/devhelp/devhelp/libdevhelp-2.la
 
 include $(top_srcdir)/build/cppcheck.mk
--- a/geanyminiscript/src/gms_gui.c
+++ b/geanyminiscript/src/gms_gui.c
@@ -26,7 +26,7 @@
  * \brief it is the graphical user interface of the geany miniscript plugin
  */
 #include    "config.h"
-#include    "geany.h"
+#include    "geanyplugin.h"
 
 #include    <glib/gstdio.h>
 #include    <glib/gprintf.h>
@@ -42,15 +42,6 @@
 #ifdef HAVE_LOCALE_H
 #include    <locale.h>
 #endif
-
-/* geany headers */
-#include    "support.h"
-#include    "plugindata.h"
-#include    "editor.h"
-#include    "document.h"
-#include    "prefs.h"
-#include    "utils.h"
-#include    "ui_utils.h"
 
 /* user header */
 #include    "gms_debug.h"
--- a/shiftcolumn/src/shiftcolumn.c
+++ b/shiftcolumn/src/shiftcolumn.c
@@ -23,19 +23,11 @@
 # include "config.h"
 #endif
 
-#include "geany.h"
-#include "support.h"
+#include "geanyplugin.h"
 
 #ifdef HAVE_LOCALE_H
 # include <locale.h>
 #endif
-
-#include "ui_utils.h"
-
-#include "document.h"
-#include "keybindings.h"
-#include "plugindata.h"
-#include "geanyfunctions.h"
 
 #include <glib.h>
 #include <glib/gprintf.h>


codenav/src/utils.c
```diff
@@ -19,7 +19,7 @@
  *  along with this program; if not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "utils.h"
```
b4n: don't remove this, it's the include for the plugin's utils.h, aka this file's header

codenav/src/utils.c
```diff
@@ -19,7 +19,7 @@
...
+#include "geanyplugin.h"
```
b4n: hum, why would we need this if we didn't before?

devhelp/src/Makefile.am
```diff
@@ -25,6 +25,7 @@ devhelp_la_CFLAGS = \
 
 devhelp_la_LIBADD = \
 	$(DEVHELP_LIBS) \
+	$(COMMONLIBS) \
```
b4n: OK

b4n:
OK, technically most are good, but why are they needed (nut the `COMMONLIBS` thing, which I understand)

codenav/src/utils.c
kugel-: I got this warning, and the change fixed it for me. I didn't realize there is also a local utils.h
  CC       utils.lo
utils.c: In Funktion »compare_strings«:
utils.c:92:2: Warnung: Implizite Deklaration der Funktion »utils_str_equal« [-Wimplicit-function-declaration]
  return (gint)(!utils_str_equal(a, b));

It's bad that that codenav has a header called util.h, it'll always clash. The convention should have been to #include <geany/geanyplugin.h> but that's a lot of changes.

geanyminiscript/src/gms_gui.c
```diff
@@ -26,7 +26,7 @@
  * \brief it is the graphical user interface of the geany miniscript plugin
  */
 #include    "config.h"
-#include    "geany.h"
+#include    "geanyplugin.h"
```
kugel-: This change fixes
  CC       gms_gui.lo
In file included from gms_gui.c:57:0:
gms.h:30:8: error: unknown type name 'GeanyFunctions'
 extern GeanyFunctions   *geany_functions;

+#include "geanyfunctions.h" would also work but geanyplugin.h is recommended anyway

shiftcolumn/src/shiftcolumn.c
```diff
@@ -23,20 +23,12 @@
...
 #endif
 
-#include "ui_utils.h"
-
-#include "document.h"
-#include "keybindings.h"
-#include "plugindata.h"
-#include "geanyfunctions.h"
-
 #include <glib.h>
```
kugel-: This change fixes

  CC       shiftcolumn.lo
shiftcolumn.c: In function 'shift_left_cb':
shiftcolumn.c:95:4: warning: implicit declaration of function 'sci_has_selection' [-Wimplicit-function-declaration]
    if (sci_has_selection(sci)){
    ^
#include "sciwrappers.h" would also work but geanyplugin.h is recommended anyway

codenav/src/utils.c
```diff
@@ -19,6 +19,8 @@
  *  along with this program; if not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <geanyplugin.h>
+
 #include "utils.h"
```
kugel-: As an alternative one could replace the inappropriate use of utils_str_equal with g_strcmp/0 but i dont like to potentially change the behavior now

frlan:
I'm planning to merge the changes only if referenced Geany changes are getting merged.

kugel-:
These changes don't depend on the Geany changes. They should be applied either way

devhelp/src/Makefile.am
b4n: I now fixed this part in 2d1dee0d9a135f2d6d29c4eead6c6e13e5d10dcd (and the markdown one in 856b3d9a0d065866c8117b031fe82087543b3f6f)

geanyminiscript/src/gms_gui.c
```diff
@@ -26,7 +26,7 @@
...
```
b4n: ideally should use `<>`s

shiftcolumn/src/shiftcolumn.c
```diff
@@ -23,20 +23,12 @@
 # include "config.h"
 #endif
 
-#include "geany.h"
-#include "support.h"
+#include "geanyplugin.h"
```
b4n: `<>`s here too

b4n:
OK, so now I ignored this one when I committed 2d1dee0d9a135f2d6d29c4eead6c6e13e5d10dcd and 856b3d9a0d065866c8117b031fe82087543b3f6f, this "only" "fixes" plugins "incorrectly" including something else than `<geanyplugin.h>`.  But do we have a real guideline on that?  I agree we should and plugins probably should only include `<geanyplugin.h>`, but [the docs](http://www.geany.org/manual/reference/howto.html#helloworld) don't seem to be really clear on this.

And if we do that, maybe there are more plugins to fix? :)

codenav/src/utils.c
LarsGit223: I am not getting that warning if I compile ```codenav/src/utils.c``` from current master.

LarsGit223:
I fixed all usages of ```#include "geanyplugin.h"``` in PR #882.

shiftcolumn/src/shiftcolumn.c
LarsGit223: This seems to have been fixed separately in f2ab51090839b2a93ec8cc3ed0e1e4874ff9b3cf.

geanyminiscript/src/gms_gui.c
LarsGit223: I do not get that error if I build from current master.

LarsGit223:
IMHO we could close this unmerged if PR #882 gets merged because all issues then seem to have been fixed in other PRs.

@b4n, @kugel-: what do you think?

kugel-:
Makes sense, feel free.

LarsGit223:
Closing in favor of all changes done separately in other PRs.

PR #208, opened 2015-04-06
Build: per plugin version check

b4n:
Introduce a macro for a particular plugin to depend on a Geany version.

The goal would be to replace the global Geany version check in GP with a per-plugin one, allowing to build some bleeding-edge plugins against older Geany if it so pleases the user -- or the other way around, allow plugins to depend on bleeding-edge Geany.

Ideally there would be a similar macro to check for a specific Geany *API* version, to allow seamless enabling/disabling of plugins depending on some development version of Geany (imagine someone building dev-GP with a slightly out-of-date dev-Geany).  But this requires finding out Geany API version, which is currently non-trivial.

For now this PR is mostly a suggestion, and is not so useful without the API version check, as it has a release version granularity, which is not sufficient in all cases.

What do you think?  Are the benefits of avoiding build failures with unmatched versions, and allowing building some plugins with older versions of Geany enough to warrant the added complexity of maintaining a per-plugin version check?


b4n:
OK, I added a macro to get the API version (relatively tricky, but it should work well), and macros to easily make a plugin depend on a specific one.

PR #209, opened 2015-04-06
projectorganizer: Fix API version check

b4n:
The API version check was rolled back as part of c890cbc3 probably by
mistake, as the plugin really depends on 221 as it uses the new
TagManager API.

---

@techee is that OK?


techee:
Oh, of course. Thanks for noticing.

PR #210, opened 2015-04-06
Fix checks for openpty() on FreeBSD

b4n:
FreeBSD has openpty() in *libutil.h*.

I PR this because while the Waf changes work, they might be deemed ugly.  @eht16 if you have a better version for this it'd be welcome :)  Otherwise they are OK enough with me.

### DIFF
--- a/build/wafutils.py
+++ b/build/wafutils.py
@@ -141,6 +141,28 @@ def check_c_header_cached(conf, **kw):
     return result
 
 
+def check_c_func_in_headers_cached(conf, header_names, mandatory=True, **kw):
+    """
+    Check for @function_name in any of @header_names.
+    Use like a conf.check_cc() for a lib, but pass a list of headers in
+    @header_names.
+
+    This checks for the headers firs with check_c_header_cached(), then
+    if this succeeded calls conf.check_cc() on it.
+    """
+
+    exception = None
+    for header in header_names:
+        try:
+            check_c_header_cached(conf, header_name=header)
+            return conf.check_cc(**dict(kw, header_name=header, mandatory=True))
+        except Exception as e:
+            exception = e
+    if mandatory:
+        raise exception
+    return None
+
+
 def check_cfg_cached(conf, **kw):
     # Works as conf.check_cfg() but tries to cache already checked packages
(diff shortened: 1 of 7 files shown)


techee:
Just tested on OS X - works fine here.

build/wafutils.py
```diff
@@ -141,6 +141,31 @@ def check_c_header_cached(conf, **kw):
...
+    Check for @function_name in any of @header_name.
+    Use like a conf.check_cc() for a lib, but you can pass a list of
+    headers in @header_name.
+
+    This checks for the headers firs with check_c_header_cached(), then
+    if this succeeded calls conf.check_cc() on it.
+    """
+    headers = kw.get('header_name')
+    mandatory = kw.get('mandatory', True)
+
```
eht16: Since both kwargs are passed explicitly to check_cc() below, we probably should .pop() them from the dict to not pass them twice to check_cc()

build/wafutils.py
```diff
@@ -141,6 +141,31 @@ def check_c_header_cached(conf, **kw):
...
+    exception = None
+    for header in [headers] if type(headers) is str else headers:
```
eht16: Since this whole function only makes sense if more than one header name is passed in, I would rename the argument header_name into header_names and so intimate that a sequence has to be passed in.
This would also make the for-loop more easy to read because we don't need to check for input type.

Also the mandatory and header_names arguments could be listed explicitly in the function definition instead of 'hiding' them in **kw.

eht16:
Except the both code comments this looks fine to me. Why did you think it would be that ugly? :smile:

b4n:
@eht16

I updated the commit to address your remarks.  Looks way better now indeed :)

> Why did you think it would be that ugly?

Well, a combination of what your remarks fixed, and that I don't find it particularly clean to `try…except` in the loop just to re-raise the exception if the loop exited.  But I couldn't really find another way to do this, and it works, so it might be good enough ^^

eht16:
I agree on the part that re-raising a remembered exception is not perfectly nice. Still it is ok I guess.
If anyone comes up with a cleaner solution later, we can improve it then as well.

PR #227, opened 2015-05-04
overview: Support Cairo < 1.12

### LINKED ISSUE #222: Overview: Not compiling against cairo 1.10 (e.g. on Windows)
The Overview plugin uses CAIRO_ANTIALIAS_GOOD which is only available since Cairo 1.12.
But on Windows, the latest available binary build of Cairo (at least on gtk.org) is 1.10 and so currently the plugin cannot be compiled (see http://nightly.geany.org/win32/build_win32_plugins_stderr.log).

We could either:
a) require Cairo 1.12 for this particular plugin
b) adjust the code to only use CAIRO_ANTIALIAS_GOOD if available
c) require Cairo 1.12 generally in G-P

c) would actually be possible, I found some binaries on http://sourceforge.net/projects/dia-installer/files/cairo/1.12.16/cairo-devel-1.12.16.zip/download.

b4n:
Closes #222.


codebrainz:
I need to add a note to the README file (I only mentioned it in initial PR description), but ideally PRs/changes could be made against my upstream repo as it uses it's own build system so it can be free-standing as well, and there's a script to update the fork in Geany-Plugins. In any case, go ahead and merge if you want, since it's breaking the nightly build, and there's already been changes on the fork I'll have to port back to upstream.

IMO, it'd be nicer to not redefine the meaning of that constant though, and rather just guard out the single call to set anti-aliasing mode, but it's no big deal, and I can easily change it later.

b4n:
If you fix it soon-ish I really don't mind, and you can do it the way you prefer it.

And indeed if the call is not strictly required it's probably nicer to just not perform it than redefine GOOD.

codebrainz:
I should really check, because the `_DEFAULT` constant name might be misleading and AA might be off by default or something. Also, thanks for the patches upstream!

codebrainz:
Closed in 1f08a3668cf86abde9942c1d11437e05a9a48438

PR #228, opened 2015-05-08
Add plugin for multiple kinds of Scintilla markers 

pik:
Hello, 
I was playing around with the whole Scintilla plugin thing - and made this. It's actually not as silly as it seems, particularly having multiple markers is quite useful when reading or refactoring a large or confusing section of code. 

On the other-hand I don't know how comfortable I am with adding additional UI elements (Geany is already kind of heavy on those compared to other IDE's maybe) for functionality that's infrequently useful. 

Maybe a better or at least more versatile approach would be to expose some of the Search/Markup api (either by changes to Geany search.h) or via. a plugin to GeanPy? Which might not be gui friendly but could be a lot less limiting than any particular plugin, and could give access to other (search/replace) functions as well.


frlan:
Before digging deeper, can you please add yourself to MAINTAINERS as well as add maybe file to be translated to po/POTFILES.in?

pik:
Of course, I'm actually not sure this is worth pulling in though - in terms of user experience.

frlan:
Well at least w/o MAINTAINERS I'm not going to merge it ;)

pik:
OK - but please don't merge it for now :)

frlan:
Sorry for late feedback. When running autogen.sh there seems to miss a folder: 

    autoreconf: running: automake --add-missing --copy --force-missing
    moremarkup/Makefile.am:2: error: required directory moremarkup/data does not exist

moremarkup/src/moremarkup.c
```diff
@@ -0,0 +1,764 @@
...
+#   include "config.h"
+#endif
+
+#include "geanyplugin.h"
+#include <string.h>
+#include <gdk/gdk.h>
+#include "marker.h"
+#include "moremarkup.h"
+
+PLUGIN_VERSION_CHECK(GEANY_API_VERSION)
```
frlan: I think it would be better to add a real hard coded number here

PR #234, opened 2015-06-05
Overview: install prefs.ui when using the Waf build system

eht16:
This was missing so far causing the file data/prefs.ui not being installed when using the Waf build system.
I guess this part was just forgotten.


codebrainz:
Thanks

PR #236, opened 2015-06-06
autotools: Add a flag to enable or disable all plugins

b4n:
This allows to change the default for whether every plugins are enabled or disabled, effectively allowing to easily select either all, or only a specific set of plugins.

To force enable all plugins:
	`configure --enable-all-plugins`

To force enable all plugins, but disable Addons:
	`configure --enable-all-plugins --disable-addons`

To force enable all plugins, but Addons which is checked automatically:
	`configure --enable-all-plugins --enable-addons=auto`

To disable all plugins, but force enable Addons:
	`configure --disable-all-plugins --enable-addons`

### DIFF
--- a/README
+++ b/README
@@ -82,6 +82,15 @@ Example:
 This will force force both geanylua and spellcheck plugins to be enabled even
 if some dependencies are missing.
 
+You can also use the special option ``all-plugins`` to change the default for all
+plugins.  For example, using ``--enable-all-plugins`` will forcefully enable all
+plugins as if you passed each plugin-specific ``--enable-<option>``.  This can be
+combined with specific plugin's ``--enable-<option>`` to only select a specific
+subset of the plugins without listing them all.  For example, you can use this
+to easily enable one single plugin, using e.g. ``--disable-all-plugins
+--enable-<option>``, or to forcefully enable all plugins but a few specific
+ones, using e.g. ``--enable-all-plugins --enable-<option>=auto``.
+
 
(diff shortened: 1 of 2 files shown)


codebrainz:
It's not clear (to me) from the M4 whether `--enable-all-plugins=auto` is accepted or the default (you mention "force" so it sounds like not). It might be useful and seems to be supported in the `_GP_ARG_DISABLE_ALL` part.

Other than that, I think this is a really good idea.

codebrainz:
I think someone said this is already supported by Waf build system, but it might be useful to add `all` as a plugin name that can be used with `--list-plugins`, `--enable-plugins`, and `--skip-plugins` to make it explicit?

b4n:
> It's not clear (to me) from the M4 whether `--enable-all-plugins=auto` is accepted or the default (you mention "force" so it sounds like not). It might be useful and seems to be supported in the `_GP_ARG_DISABLE_ALL` part.

You can pass `auto` and it's the default, which doesn't change the current behavior.  Passing auto however is pretty much useless as it has virtually no effect.

frlan:
Shall we put this into 1.25?

b4n:
@frlan I think we should, but I wished for some comments here from the people complaining about the missing feature.  Without them, I can't really know if they appreciate this solution, or if they would rather have some other interface.

@zhekov do you have an opinion on this?

b4n:
Added some docs

PR #238, opened 2015-06-19
addons: support for URI's containing ~ and +

glitsj16:
Better current word selection for URI's containing ~ and/or +. By adding ~ and + to GEANY_WORDCHARS URI's from launchpad.net (and others) get selected (and opened) correctly. Otherwise these get cut at the tilde character.

Example: https://launchpad.net/~geany-dev/+archive/ubuntu/ppa gets selected as https://launchpad.net/

PR #239, opened 2015-06-19
addons: support for URI's containing ~ and +

glitsj16:
Current word selection fails on URI's containing ~ and/or + (like launchpad PPA's). Adding ~ and + to GEANY_WORDCHARS fixes this.


b4n:
LGTM

Replaces #238 I guess :)

glitsj16:
This indeed replaces #238, after the mess I created it seemed best to start afresh. Many thanks for your patience & the tips, appreciated :-)

PR #243, opened 2015-06-22
[1.26] projectorganizer: Add ellipsis to menu items requiring user input

b4n:
This is consistent with Geany and follows various common HIG.

https://developer.gnome.org/hig-book/stable/menus-design.html.en#menu-item-type-command

----

**WARNING**: do *not* merge before 1.25 is out, as we froze strings and this modifies some.


techee:
@b4n Thanks for noticing these, will merge it once 1.25 is out.

PR #246, opened 2015-06-24
Replace g_timeout_add with g_idle_add

scriptum:
`g_idle_add` does not make a big performance impact but improves usability
by removing annoying delay


b4n:
LGTM

PR #254, opened 2015-06-25
Multiterm: make strings translatable

### LINKED ISSUE #247: multiterm plugin has no translation strings
Please add them

b4n:
Closes #247.

---

I'm not quite sure whether https://github.com/geany/geany-plugins/blob/master/multiterm/src/shell-config.vala#L48 should be translated or not, but I left it alone.

Also, being able to translate the [default profile name](https://github.com/geany/geany-plugins/blob/master/multiterm/src/defconf.vala#L97) would be nice.  This would be quite easy if this was a real file that would get installed instead of a string template.  Or the template could be built with a translation in the middle.

### DIFF
--- a/multiterm/src/Makefile.am
+++ b/multiterm/src/Makefile.am
@@ -24,6 +24,7 @@ multiterm_la_SOURCES = \
 	terminal.vala
 
 multiterm_la_CFLAGS = \
+	-DGETTEXT_PACKAGE='"$(GETTEXT_PACKAGE)"' \
 	$(AM_CFLAGS) \
 	$(MULTITERM_CFLAGS)
 
--- a/multiterm/src/config.vala
+++ b/multiterm/src/config.vala
@@ -44,7 +44,7 @@ namespace MultiTerm
 			}
 			catch (FileError err)
 			{
-				warning("Unable to save config file %s: %s", _filename, err.message);
+				warning(_("Unable to save config file %s: %s"), _filename, err.message);
 				return true;
 			}
 		}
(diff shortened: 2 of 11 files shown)


multiterm/src/Makefile.am
```diff
@@ -24,6 +24,8 @@ multiterm_la_SOURCES = \
 	terminal.vala
 
 multiterm_la_CFLAGS = \
+	-DGETTEXT_PACKAGE='"$(GETTEXT_PACKAGE)"' \
+	-DLOCALEDIR='"$(localedir)"' \
```
b4n: Alternatively you could probably have a config.h VAPI module, something like this:
```vala
[CCode(cheader="config.h", lower_case_cprefix="")]
namespace Config {
  const string GETTEXT_PACKAGE;
  const string LOCALEDIR;
}
```

codebrainz: I think I like that way better, and IIRC I've done it before on some project. If you feel like doing it go ahead, otherwise this way is fine too, at least the globals are constant.

Would it make more sense to have these in the Geany .vapi?

codebrainz:
Looks good to me, though I haven't had a chance to test Autotools compiling with the changes yet.

multiterm/src/Makefile.am
b4n: > If you feel like doing it go ahead, otherwise this way is fine too, at least the globals are constant.

I won't do it today, but you can ;)

> Would it make more sense to have these in the Geany .vapi?

I don't think so no, because Geany doesn't provide it, nor depend on it.  Well, `GETTEXT_PACKAGE` a little as if it's not defined we will alias `_` to a no-op, but that's about it.

b4n:
> Looks good to me, though I haven't had a chance to test Autotools compiling with the changes yet.

Herm, makes me realize I didn't test the Waf build system, and it's likely not to work.

multiterm/src/Makefile.am
b4n: Hum, actually I just tried and it's not possible for `LOCALEDIR` as it has to be changeable when running Make itself, so it has to be passed through there.
So we could have this for `GETTEXT_PACKAGE`, but I doubt it's very sensible to use it only for this.

b4n: The again, I'm a tiny be stupid as `LOCALEDIR` is already passed through `AM_CFLAGS` from *vars.build.mk`, so it's redundant here.

b4n:
Waf is fixed (to the extent of the changes here)

PR #262, opened 2015-06-30
fix double "_" entree in GEANY_WORDCHARS

glitsj16:
The double entree is indeed superfluous, took the added "_" out as recommended. Thanks for that. Looking good now :-)


b4n:
not very important as duplicate would do no much harm, but that's cleaner. LGTM.

PR #270, opened 2015-07-12
GeanyPy: Unload Python plugins when GeanyPy is deactivated

codebrainz:
This prevents crashes and/or lockups when the GeanyPy C plugin
is reloaded.

FIXME: Because it calls the normal unloading path, it updates
the list of loaded plugins to remove each plugin. The net effect
is that when the C GeanyPy plugin is unloaded, it forgets all
of the Python plugins that were loaded at the time.

### DIFF
--- a/geanypy/geany/manager.py
+++ b/geanypy/geany/manager.py
@@ -93,6 +93,9 @@ class PluginManager(gtk.Dialog):
 	def deactivate_plugin(self, filename):
 		self.loader.unload_plugin(filename)
 
+	def deactivate_all_plugins(self):
+		self.response(gtk.RESPONSE_CLOSE)
+		self.loader.unload_all_plugins()
 
 	def load_plugins_list(self):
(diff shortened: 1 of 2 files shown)


codebrainz:
Feel free to leave out of release, but it might save some bug reports.

b4n:
This seem to mostly work (no more crashes after manually unloading the plugin), although it's super annoying as this means on every geany run (well, whenever it quits, which unloads the plugin) all Python plugins need to be re-enabled.
I'm afraid this might not please users.

frlan:
Postponed to post 1.25 due to the massiv effect on restart of Geany.

codebrainz:
Related: https://github.com/codebrainz/geanypy/pull/31

frlan:
@kugel-  As you are currently maintainer of this part of geanypy, can you please have a look whether it's still a point?

kugel-:
Geanypy is a proxy plugin these days. These cannot be deactivated (through the GUI) without deactivating each and every subplugin. This can be closed, thanks for the reminder.

PR #271, opened 2015-07-12
Replace deprecated GTK2 API to fix compilation against GTK3

eht16:
This really fixes only compilation of the plugin and
makes the document list in the toolbar working.

There are still plenty of deprecation warnings across the whole
plugin and I don't know which to fix and which to accept, e.g.
GTK 3.10 deprecated stock items, some later version
deprecated the whole GtkStatusIcon.


kugel-:
Seems trivial enough to land in 1.25, so add-ons could be part if a unofficial gtk3 installer

kugel-:
Please merge :)

PR #273, opened 2015-07-16
Add send_text_message

scriptum:
Geany crashes if Python string passed to SSM. Use this function if last argument of SSM is string.


codebrainz:
This PR should probably be on upstream repo, otherwise it will cause the maintainer of the plugin in GP grief next time they update from upstream and this part conflicts.

scriptum:
https://github.com/codebrainz/geanypy/pull/33

PR #278, opened 2015-08-26
Ported treebrowser plugin to GTK+ >= 3.10.0

paolostivanin:
now the treebrowser plugin uses GTK+ API compatible with version 3.10.0 and above. No more warnings during compilation :)

Tested on Arch x64, Ubuntu 14.04 x86, Ubuntu 15.04 amd64.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -127,10 +127,13 @@ W: http://plugins.geany.org/geanynumberedbookmarks.html
 S: Maintained
 
 geanypg
-P: Hans Alves <alves.h88@gmail.com>
-M: Hans Alves <alves.h88@gmail.com>
+P1: Hans Alves <alves.h88@gmail.com>
+P2: Paolo Stivanin <info@paolostivanin.com>
+M1: Hans Alves <alves.h88@gmail.com>
+M2: Paolo Stivanin <info@paolostivanin.com>
 W: http://plugins.geany.org/geanypg.html
-S: Odd Fixes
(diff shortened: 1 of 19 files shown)


b4n:
I don't like this.  It adds a GTK 3.16 dependency for no good reason IMO.
Also, dropping all icons is questionable, but well, that's another topic.

b4n:
At the very least the changes on the various plugins should be separate PRs.

paolostivanin:
I didn't add GTK+ 3.16 functions :) I also think is not a good idea to use that version because *buntu 14.04 LTS users wouldn't be able to use Geany GTK+3 ;)

paolostivanin:
Ok. So, can I close this pull request and open 3 new ones?

b4n:
> I didn't add GTK+ 3.16 functions :)

You use [`GtkLabel::xalign`](https://developer.gnome.org/gtk3/stable/GtkLabel.html#GtkLabel--xalign) which was introduced by 3.16.  Of course, you set it dynamically so it's not *that* bad, but it's kinda bad to set an non-existing property.

> Ok. So, can I close this pull request and open 3 new ones?

Yes, but I myself won't merge them.  As those are plugins not part of Geany core I won't oppose to them, but you'll need to convince someone else.

paolostivanin:
I've already remove that code, it was my mistake :) I'll push it with the new prs I'm gonna create!
BTW, thanks for your time @b4n !

PR #283, opened 2015-09-08
git-changebar: Fix compilation against libgit2 >= 0.23

eht16:
libgit2 changed its API and added a new argument to git_diff_buffers().

The new argument is a callback for binary files. This should not be relevant here because Geany probably won't be able to handle binary files in the near future.

I tested compilation with this patch against libgit2 0.22 *and* 0.23.

However, I could not test the plugin against 0.23 because of weird unresolved symbols in libgit2. I assume this is a temporary problem with my distro (ArchLinux).

@b4n not sure if you like the style, I broke the function arguments into one per line to make them more readable. Feel free to drop this PR and do it in another way.

### DIFF
--- a/git-changebar/src/gcb-plugin.c
+++ b/git-changebar/src/gcb-plugin.c
@@ -655,8 +655,21 @@ diff_buf_to_doc (const git_buf   *old_buf,
   opts.context_lines = 0;
   opts.flags = GIT_DIFF_FORCE_TEXT;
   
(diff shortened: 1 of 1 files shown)


eht16:
Update: I got to test the patch with libgit2 0.23 on a Debian Unstable system and it works as expected.

b4n:
I fixed it differently to keep compatibility code together on top.

Thanks a lot for noticing and providing a fix :)

PR #284, opened 2015-09-16
Addons bookmark refresh fix

eht16:
Update bookmark list also on any line changes in the document

In addition to updating the bookmark list on document notebook tab change
and adding/removing line markers, also rebuild the bookmark list
when the text in the document changes with lines added or deleted.

Should fix SF bugs 129 and 39.

This implements some ideas mentioned in the above bug reports like updating the bookmark list
in an idle callback. Additionally, the list store is detach from the GtkTreeView to increase performance on big bookmark lists.

Maybe it is still worth to add a refresh button (or rather a context menu item to refresh) but I would try it without at first.

There is one issue left: if the `Transpose current line` feature is used, the bookmarks don't update. But this is not a problem in the plugin but rather in either Geany or Scintilla as the marker also stays at the "old" line. This should be managed separately.

### DIFF
--- a/addons/src/ao_bookmarklist.c
+++ b/addons/src/ao_bookmarklist.c
@@ -62,7 +62,15 @@ struct _AoBookmarkListPrivate
 
 	gint		 search_line;
 	GtkTreeIter	*search_iter;
+
+	guint refresh_idle_source_id;
 };
+
+typedef struct
+{
+	AoBookmarkList *bm;
+	GeanyDocument *document;
+} AoBookmarkListRefreshContainer;
 
 enum
 {
@@ -401,14 +409,17 @@ void ao_bookmark_list_activate(AoBookmarkList *bm)
 }
 
 
-void ao_bookmark_list_update(AoBookmarkList *bm, GeanyDocument *doc)
+static gboolean update_bookmark_list_delayed(gpointer data)
 {
 	gint line_nr = 0;
 	gint mask = 1 << 1;
-	ScintillaObject *sci = doc->editor->sci;
+	AoBookmarkListRefreshContainer *container = data;
+	AoBookmarkList *bm = container->bm;
 	AoBookmarkListPrivate *priv = AO_BOOKMARK_LIST_GET_PRIVATE(bm);
+	GeanyDocument *doc = container->document;
+	ScintillaObject *sci = doc->editor->sci;
 
-	if (priv->enable_bookmarklist)
+	if (priv->enable_bookmarklist && DOC_VALID(doc))
 	{
 		gtk_list_store_clear(priv->store);
 		while ((line_nr = scintilla_send_message(sci, SCI_MARKERNEXT, line_nr, mask)) != -1)
@@ -416,6 +427,26 @@ void ao_bookmark_list_update(AoBookmarkList *bm, GeanyDocument *doc)
 			add_line(bm, sci, line_nr);
 			line_nr++;
 		}
+	}
+
+	g_free(container);
+	priv->refresh_idle_source_id = 0;
+	return FALSE;
+}
+
+
+void ao_bookmark_list_update(AoBookmarkList *bm, GeanyDocument *doc)
+{
+	AoBookmarkListPrivate *priv = AO_BOOKMARK_LIST_GET_PRIVATE(bm);
+	if (priv->refresh_idle_source_id == 0)
+	{
+		AoBookmarkListRefreshContainer *container = g_new0(AoBookmarkListRefreshContainer, 1);
+		container->bm = bm;
+		container->document = doc;
+		priv->refresh_idle_source_id = plugin_idle_add(
+			geany_plugin,
+			update_bookmark_list_delayed,
+			container);
 	}
 }
 
@@ -444,6 +475,7 @@ static void ao_bookmark_list_init(AoBookmarkList *self)
 	AoBookmarkListPrivate *priv = AO_BOOKMARK_LIST_GET_PRIVATE(self);
 
 	priv->page = NULL;
+	priv->refresh_idle_source_id = 0;
 }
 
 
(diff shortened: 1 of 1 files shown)


codebrainz:
LGTM, though I didn't test.

addons/src/ao_bookmarklist.c
```diff
@@ -62,8 +62,16 @@ struct _AoBookmarkListPrivate
...
 	gint		 search_line;
 	GtkTreeIter	*search_iter;
+
+	guint refresh_idle_source_id;
 };
 
+typedef struct
+{
+	AoBookmarkList *bm;
+	GeanyDocument *document;
```
b4n: you might want to use Document IDs as you delay the reference to the document.  It probably doesn't matter much in practice though, it's highly unlikely a document is reused during the delay.

b4n:
Otherwise, as @codebrainz, I didn't test but it LGTM

eht16:
Thanks for the feedback.

I changed the document referencing to IDs as suggested.

Refreshing the treeview with about 5000 markers set takes about 40-50ms with de- and reattching the model. Without detaching, refreshing takes about 60-70ms, on my machine.
So it is actually not that big difference and I doubt users will click that many markers. I cheated a bit by programatically setting them on startup :).

So, as the difference is actually not really big, I removed commit 0d130be42b3e7968c597c85a1782b1e34d69d5a8. However, I could not notice any flickering, with or without the commit.

addons/src/ao_bookmarklist.c
```diff
@@ -401,14 +409,17 @@ void ao_bookmark_list_activate(AoBookmarkList *bm)
...
+static gboolean update_bookmark_list_delayed(gpointer data)
 {
 	gint line_nr = 0;
 	gint mask = 1 << 1;
-	ScintillaObject *sci = doc->editor->sci;
+	AoBookmarkListRefreshContainer *container = data;
+	AoBookmarkList *bm = container->bm;
 	AoBookmarkListPrivate *priv = AO_BOOKMARK_LIST_GET_PRIVATE(bm);
+	GeanyDocument *doc = document_find_by_id(container->document_id);
+	ScintillaObject *sci = doc->editor->sci;
```
b4n: logic error: you dereference `doc` here, but check whether it's NULL on the next line.

db0451: I assume the fix is this?

GeanyDocument *doc = document_find_by_id(container->document_id);

	if (doc != NULL && priv->enable_bookmarklist)
	{
		ScintillaObject *sci = doc->editor->sci;
// etc.

b4n: yes

eht16: Oops. Fixed in 8ab3ee28baca96e683b92a585b9ab226196b170f.

addons/src/ao_bookmarklist.c
```diff
@@ -401,15 +409,18 @@ void ao_bookmark_list_activate(AoBookmarkList *bm)
...
-	ScintillaObject *sci = doc->editor->sci;
+	AoBookmarkListRefreshContainer *container = data;
+	AoBookmarkList *bm = container->bm;
 	AoBookmarkListPrivate *priv = AO_BOOKMARK_LIST_GET_PRIVATE(bm);
+	GeanyDocument *doc = document_find_by_id(container->document_id);
 
-	if (priv->enable_bookmarklist)
+	if (priv->enable_bookmarklist && doc != NULL)
 	{
+		ScintillaObject *sci = doc->editor->sci;
```
b4n: style probably asks for a blank line here (really no biggie, and your call actually)

addons/src/ao_bookmarklist.c
```diff
@@ -417,6 +428,26 @@ void ao_bookmark_list_update(AoBookmarkList *bm, GeanyDocument *doc)
...
+
+	g_free(container);
+	priv->refresh_idle_source_id = 0;
+	return FALSE;
+}
+
+
+void ao_bookmark_list_update(AoBookmarkList *bm, GeanyDocument *doc)
+{
+	AoBookmarkListPrivate *priv = AO_BOOKMARK_LIST_GET_PRIVATE(bm);
```
b4n: here too

addons/src/ao_bookmarklist.c
```diff
@@ -417,6 +428,26 @@ void ao_bookmark_list_update(AoBookmarkList *bm, GeanyDocument *doc)
...
+	if (priv->refresh_idle_source_id == 0)
+	{
+		AoBookmarkListRefreshContainer *container = g_new0(AoBookmarkListRefreshContainer, 1);
```
b4n: and here

b4n:
LGTM

eht16:
I added https://github.com/eht16/geany-plugins/commit/e8c0dccee3a0a1688325939f1640c455498e03b0 to add the newlines and squashed unimportant commits together.

eht16:
And it seems I pulled too many commits into the rebase. Can this get problematic when merging?

b4n:
yes, it very well might.  But you should be able to fix that fairly easily to fix: check master out, pull (fast-forward, `pull --ff-only`) from upstream, and rebase your branch on top of that.  Or you can also just `rebase -i 8234f7c^` and remove the two offending/duplicate commits, only keeping yours.  As it's a separate branch, there won't be no problem.

b4n:
@eht16 if you're interested, I just did the second solution as you didn't seem to update this already: https://github.com/b4n/geany-plugins/commits/eht16/addons_bookmark_refresh_fix ;)

eht16:
Thanks again for your help and sorry for unnecessary mistakes in code and GIT usage :(.

b4n:
np, that's what annoying reviews are for :)

PR #289, opened 2015-10-11
geanypy: Try to locate the real dereferenced Python DSO

b4n:
Try to pre-load the Python library with its real, fully dereferenced
path.

This is necessary because some installations don't have the symlink
from the generic (.so) to the versioned one (.so.X.Y) when the Python
development files aren't installed (i.e. libpython2.7-dev on Debian).

Fortunately for us (at least under Debian) Python has the name of the
final installed library in its `INSTSONAME` sysconfig's config var, so
use this instead of `LDLIBRARY` when it's available.

Closes https://github.com/codebrainz/geanypy/issues/20


b4n:
PR'd on the upstream repo too: https://github.com/codebrainz/geanypy/pull/36

PR #298, opened 2015-11-08
Removing word 'plugin' from plugin-description on GeanyLaTeX, GeanyLipsum, GeanySendMail, TabbleConvert and UpdateChecker

frlan:


tableconvert/src/tableconvert.c
```diff
@@ -29,7 +29,7 @@ PLUGIN_VERSION_CHECK(200)
 
 PLUGIN_SET_TRANSLATABLE_INFO(
     LOCALEDIR, GETTEXT_PACKAGE, _("Tableconvert"),
-    _("A little plugin to convert lists into tables"),
+    _("Convert lists into tables for different filetypes"),
```
b4n: Probably should be "Converts" to better fit in the style of the other

b4n:
LGTM

PR #306, opened 2015-11-12
geanynumberedbookmarks: Pass document pointer around where needed

### LINKED ISSUE #301: plugin Numbered Bookmarks activated causes geany to crash at start-up
When the plugin Numbered Bookmarks is switched on, Geany crashes instantly on start-up. Running GDB shows the following warning:

Program received signal SIGSEGV, Segmentation fault.
0x00007fffd153f2f4 in SetMarker (sci=0x9fa4a0, bookmarkNumber=1, markerNumber=2, line=247) at geanynumberedbookmarks.c:885
885		fd=GetFileData(document_get_current()->file_name);

I assume that at start-up the call <b>document_get_current()</b> will return a NULL pointer? And asking a <b>file_name</b> member will cause a segfault?

Tested on geany-plugins commit 4c1137373f78bd13df9ee46c10876a3142675573 of Nov. 8th, 2015.

b4n:
Pass the document pointer instead of always getting the current document, which may or may not be the one we're working on.

Closes #301.

---

@peter-scholtens Could you please try this and see if it actually fixes the issue you've been having?

### DIFF
--- a/geanynumberedbookmarks/src/geanynumberedbookmarks.c
+++ b/geanynumberedbookmarks/src/geanynumberedbookmarks.c
@@ -746,11 +746,12 @@ static guint32 * GetMarkersUsed(ScintillaObject* sci)
 
 
 /* get next free marker number */
-static gint NextFreeMarker(ScintillaObject* sci)
+static gint NextFreeMarker(GeanyDocument* doc)
 {
 	gint i,l,m,k;
 	guint32 *markers;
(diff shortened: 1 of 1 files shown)


peter-scholtens:
I tested the fix and it does not crash. The keys still didn't work, so I created an pull request which shouls now let them work (#307). Did anyone actually tested this plug-in lately?

PR #317, opened 2015-12-07
Changes from Wolfgang Schwarz <wschwarz@online.de>

andrewjanke:
Note that I don't use geany much anymore so can't comment on the veracity of these changes.

Here are the notes from Wolfgang regarding the two bugs this fixes.

I made 2 changes to your geany-plugin Shiftcolumn, see the code attached.

-fixed a bug when one selects text from right to left and then shifts it
(see comment "in case selection_start was after selection_end" in my code)

-When no text is selected, now the current word is exchanged with the
next/previous, which I find quite useful.

### DIFF
--- a/shiftcolumn/src/shiftcolumn.c
+++ b/shiftcolumn/src/shiftcolumn.c
@@ -18,16 +18,15 @@
  *  MA 02110-1301, USA.
  */
 
-
 #ifdef HAVE_CONFIG_H
-# include "config.h"
+#include "config.h"
 #endif
 
 #include "geany.h"
 #include "support.h"
 
 #ifdef HAVE_LOCALE_H
-# include <locale.h>
+#include <locale.h>
 #endif
 
 #include "ui_utils.h"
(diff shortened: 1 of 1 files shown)


LarsGit223:
> -fixed a bug when one selects text from right to left and then shifts it
> (see comment "in case selection_start was after selection_end" in my code)

This seems to already have been fixed by @b4n in dc0af51b96d4b98ada550d464b0cd17b24268419. I cannot reproduce any issue on selecting text from the right and shifting it. Works fine.

> -When no text is selected, now the current word is exchanged with the
> next/previous, which I find quite useful.

IMHO this does not fit into this plugin. If someone needs it then a new PR to a different plugin where it fits in or a new plugin is welcome.

I vote for closing this un-merged. What do the others think?

elextr:
Since it has conflicts and the plugin is orphaned and nobody has done anything for some years I'd say its fair enough to just close.

codebrainz:
Closing due to previous comments.

PR #322, opened 2016-01-02
Improve detection of English contractions and other use of single quotes

### LINKED ISSUE #320: Doesn't and isn't are not proper English?
Both US & UK English spell-checks highlight these two words as incorrect. But I happen to know that they are correct. However, I'm not sure this is a bug in this geany-plugin or the source of these language packs itself in `ubuntu 14.04`. Please provide any suggestions.

eht16:
Temporarily add ' (single quote) to WORDCHARS to be able to parse
whole words containing a single quote (like doesn't and similar in English).

Closes #320.

### DIFF
--- a/spellcheck/src/speller.c
+++ b/spellcheck/src/speller.c
@@ -203,12 +203,27 @@ gint sc_speller_process_line(GeanyDocument *doc, gint line_number, const gchar *
 {
 	gint pos_start, pos_end;
 	gint wstart, wend;
-	GString *str;
 	gint suggestions_found = 0;
+	gint wordchars_len;
+	gchar *wordchars_orig;
+	GString *str;
 
 	g_return_val_if_fail(sc_speller_dict != NULL, 0);
 	g_return_val_if_fail(doc != NULL, 0);
 	g_return_val_if_fail(line != NULL, 0);
+
+	/* add ' (single quote) temporarily to wordchars
+	 * to be able to check for "doesn't", "isn't" and similar */
+	wordchars_len = scintilla_send_message(doc->editor->sci, SCI_GETWORDCHARS, 0, 0);
+	wordchars_orig = g_malloc0(wordchars_len + 1);
+	scintilla_send_message(doc->editor->sci, SCI_GETWORDCHARS, 0, (sptr_t)wordchars_orig);
+	if (! strchr(wordchars_orig, '\''))
+	{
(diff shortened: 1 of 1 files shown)


spellcheck/src/speller.c
```diff
@@ -203,13 +203,28 @@ gint sc_speller_process_line(GeanyDocument *doc, gint line_number, const gchar *
...
 	g_return_val_if_fail(line != NULL, 0);
 
+	/* add ' (single quote) temporarily to wordchars
+	 * to be able to check for "doesn't", "isn't" and similar */
+	wordchars_len = scintilla_send_message(doc->editor->sci, SCI_GETWORDCHARS, 0, 0);
+	wordchars_orig = g_malloc0(wordchars_len + 1);
+	scintilla_send_message(doc->editor->sci, SCI_GETWORDCHARS, 0, (sptr_t)wordchars_orig);
+	if (! strchr(wordchars_orig, '\''))
+	{
+		GString *wordchars_new = g_string_new(wordchars_orig);
```
b4n: depending on whether it's a hot spot, you could also simply add the `'` to the orig string and truncate the last byte afterward (instead of copying the string)

b4n:
LGTM, but if people use `’` and stuff like that.  But I guess fully correct word boundary recognition requires a clever library like Pango or whatever, and it's tricky; so I guess adding `'` is good enough at least for now.

spellcheck/src/speller.c
eht16: Yeah, nice idea. What exactly do you mean by "whether it's a hot spot"?

eht16:
Yes, when anything else is used for contractions/short forms than a single quote, it will fail. Though, to my knowledge, this also would not correct English then anymore.
Also, in English as well as in German and maybe other languages as well, the single quote is used for things like "John Doe's shoes" and I guess using any other character than a single quote is not valid then.

spellcheck/src/speller.c
b4n: whether the code is "hot" (called often), so basically whether optimizing this is worth the (trivial) added subtlety.

b4n:
`’` is the typographic [apostrophe](https://en.wikipedia.org/wiki/Apostrophe), and is technically valid (and even more accurate) than the typewriter one `'`.  But in practice it's less likely to be encountered in Geany I guess, because `'` is more common outside typesetting programs.

PR #324, opened 2016-01-08
Add new "Lineoperations" plugin

smostertdev:
LineOperations is a plugin that gives a bunch of simple functions that can be applied to a document.
(remove duplicate lines, empty lines, lines with only whitespace, sort lines...)

It is intended to be a convenient way to use these functions for non-technical users.
I am thinking of adding some more useful functions (e.g. similar to TextFX in Notepad++).

Take a look and let me know what you think, I'm open to suggestions for improvements.


smostertdev:
Alright, this should now compile with autotools. (Thanks for your help @frlan )
The README file was also re-written in reStructuredText. bd2293f

frlan:
Ca you please rebase your pr against current master and repush it?

lineoperations/src/linefunctions.c
```diff
@@ -0,0 +1,401 @@
...
+
+	totalnumchars = sci_get_length(doc->editor->sci);
+	totalnumlines = sci_get_line_count(doc->editor->sci);
+	lines         = g_malloc(sizeof(gchar *) * totalnumlines);
+	newfile       = g_malloc(sizeof(gchar) * (totalnumchars+1));
+	numlines      = 0;
+	nfposn        = 0;
+	k             = 0;
+
+	if(newfile && lines)    // verify memory allocation
```
b4n: `g_malloc()` aborts on failure.  If you really mean to support allocation failure you need to use `g_try_malloc()` instead (no need unless you allocate really large data, as on very low memory situations something in GLib/GTK/Geany will most likely abort anyway).

lineoperations/src/linefunctions.c
```diff
@@ -0,0 +1,401 @@
...
+	{
+		// copy *all* lines into **lines array
+		for(i = 0; i < totalnumlines; i++)
+			lines[numlines++] = sci_get_line(doc->editor->sci, i);
+
+		qsort(lines, numlines, sizeof(gchar *), compareasc);
+
+		if(numlines > 0)	// copy the first line into *newfile
+			for(j = 0; lines[0][j] != '\0'; j++)
+				newfile[nfposn++] = lines[0][j];
```
b4n: you should probably rather use `memcpy()` (or even `strcpy()`)

lineoperations/src/linefunctions.c
```diff
@@ -0,0 +1,401 @@
...
+		sci_set_text(doc->editor->sci, newfile);	// set new document
+
+	}
+
+
+	// free used memory
+	for(i = 0; i < numlines; i++)
+		if(lines[i]) g_free(lines[i]);
+	if(lines)   g_free(lines);
+	if(newfile) g_free(newfile);
```
b4n: `g_free()` (like `free()`) is `NULL`-safe, no need for the test

lineoperations/src/linefunctions.c
```diff
@@ -0,0 +1,401 @@
...
+				for(j = 0; line[j] != '\0'; j++)
+					newfile[nfposn++] = line[j];
+			}
+		}
+		newfile[nfposn] = '\0';
+		sci_set_text(doc->editor->sci, newfile);	// set new document
+
+
+		g_free(newfile);
+	}
```
b4n: I'd probably rather implement this one using only direct Scintilla manipulation:
```C
gint line_count;

sci_start_undo_action(sci);

line_count = sci_get_line_count(sci);
for (gint line = 0; line < line_count; line++) {
  if (sci_get_position_from_line(sci, line) == sci_get_line_end_position(sci, line)) {
    scintilla_send_message(sci, SCI_DELETERANGE,
                           sci_get_position_from_line(sci, line),
                           sci_get_line_length(sci, line));
    line_count--;
  }
}

sci_end_undo_action(sci);
```
or similar

lineoperations/src/linefunctions.c
```diff
@@ -0,0 +1,401 @@
...
+					for(j = 0; line[j] != '\0'; j++)
+						newfile[nfposn++] = line[j];
+			}
+
+		}
+		newfile[nfposn] = '\0';
+		sci_set_text(doc->editor->sci, newfile);	// set new document
+
+		g_free(newfile);
+	}
```
b4n: ditto, but using `sci_get_line_indent_position()`

lineoperations/src/linefunctions.c
```diff
@@ -0,0 +1,401 @@
...
+
+#include "linefunctions.h"
+
+
+// isspace()
+gboolean issp(gchar c)
+{
+	return (c == ' ' || c == '\t' || c == '\f' ||
+			c == '\v' || c == '\r' || c == '\n');
+}
```
b4n: [`g_ascii_isspace()`](https://developer.gnome.org/glib/unstable/glib-String-Utility-Functions.html#g-ascii-isspace)

lineoperations/src/linefunctions.c
```diff
@@ -0,0 +1,401 @@
...
+	{
+		for(i = 0; i < totalnumlines; i++)    // loop through opened doc
+		{
+			linelen = sci_get_line_length(doc->editor->sci, i);
+
+			if(linelen == 2)
+			{
+				line = sci_get_line(doc->editor->sci, i);
+
+				if(line[0] != '\r')
```
b4n: as `iswhitespaceline()` will recognize `\r` as a space, this code path is redundant with the `else if` below

lineoperations/src/linefunctions.c
```diff
@@ -0,0 +1,401 @@
...
+ *      with this program; if not, write to the Free Software Foundation, Inc.,
+ *      51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+
+#include "linefunctions.h"
+
+
+// isspace()
+gboolean issp(gchar c)
```
b4n: only used in this file, should be static

lineoperations/src/linefunctions.c
```diff
@@ -0,0 +1,401 @@
...
+{
+	return (c == ' ' || c == '\t' || c == '\f' ||
+			c == '\v' || c == '\r' || c == '\n');
+}
+
+
+// iswhitespaceline(gchar* line) checks if line is a whitespace line
+gboolean iswhitespaceline(gchar* line)
```
b4n: ditto

lineoperations/src/lineoperations.c
```diff
@@ -0,0 +1,193 @@
...
+
+GeanyPlugin		*geany_plugin;
+GeanyData		*geany_data;
+
+PLUGIN_VERSION_CHECK(224)
+
+PLUGIN_SET_INFO(_("LineOperations"),
+				 _("Line operations, remove duplicate lines, empty lines, lines with only whitespace, sort lines."),
+				 "0.1",
+				 "Sylvan Mostert")
```
b4n: Unless you try to support older versions of Geany on purposed, I'd recommend making use of the [new plugin API](http://www.geany.org/manual/reference/legacy.html)

lineoperations/src/lineoperations.c
```diff
@@ -0,0 +1,193 @@
...
+#include "Scintilla.h"
+#include "linefunctions.h"
+
+
+GeanyPlugin		*geany_plugin;
+GeanyData		*geany_data;
+
+PLUGIN_VERSION_CHECK(224)
+
+PLUGIN_SET_INFO(_("LineOperations"),
```
b4n: This should generally be a descriptive title in plain English (e.g. add the space between the two words).

lineoperations/src/lineoperations.c
```diff
@@ -0,0 +1,193 @@
...
+	gtk_widget_show(submenu);
+	sep1 = gtk_separator_menu_item_new();
+	sep2 = gtk_separator_menu_item_new();
+	rmdupst_item  = gtk_menu_item_new_with_mnemonic(_("Remove Duplicate Lines, Sorted"));
+	rmdupln_item  = gtk_menu_item_new_with_mnemonic(_("Remove Duplicate Lines, Ordered"));
+	rmunqln_item  = gtk_menu_item_new_with_mnemonic(_("Remove Unique Lines"));
+	rmemtyln_item = gtk_menu_item_new_with_mnemonic(_("Remove Empty Lines"));
+	rmwhspln_item = gtk_menu_item_new_with_mnemonic(_("Remove Whitespace Lines"));
+	sortasc_item  = gtk_menu_item_new_with_mnemonic(_("Sort Lines Ascending"));
+	sortdesc_item = gtk_menu_item_new_with_mnemonic(_("Sort Lines Descending"));
```
b4n: you might want to add actual mnemonics, by putting a `_` in front of the mnemonic letter in the string

build/lineoperations.m4
```diff
@@ -0,0 +1,10 @@
+AC_DEFUN([GP_CHECK_LINEOPERATIONS],
+[
+    GP_ARG_DISABLE([LineOperations], [auto])
+    GP_CHECK_PLUGIN_GTK2_ONLY([LineOperations])
```
b4n: Unless you know why you wouldn't want to support GTK3, you should remove this and try.  Actually, as I read your code nothing should stop you from supporting both GTK2 and GTK3 without a change.
So i'd suggest to drop this GTK2 enforcement

lineoperations/src/linefunctions.c
smostertdev: Thanks, this is now updated. I decided to keep g_malloc and remove the null check since it is unnecessary  as "GLib/GTK/Geany will most likely abort anyways.

lineoperations/src/linefunctions.c
smostertdev: This is now updated. I replaced this with g_stpcpy() for any for-loop string copies. (mostly for joining array of lines into new_file string.)

lineoperations/src/linefunctions.c
smostertdev: Yup my bad, I deleted all NULL tests before g_free();

lineoperations/src/linefunctions.c
smostertdev: Your solution is beautiful, I changed it. Thanks.

lineoperations/src/linefunctions.c
smostertdev: This is now updated too.

lineoperations/src/linefunctions.c
smostertdev: The function `issp()` have been removed. It is no longer needed because of your suggestions with `sci_get_line_indentation` and direct scintilla manipulation.

lineoperations/src/linefunctions.c
smostertdev: The function `iswhitespaceline()` have been removed. It is no longer needed because of your suggestions with `sci_get_line_indentation` and direct scintilla manipulation.

lineoperations/src/linefunctions.c
smostertdev: function removed

lineoperations/src/linefunctions.c
smostertdev: function removed

lineoperations/src/lineoperations.c
smostertdev: space added

lineoperations/src/lineoperations.c
smostertdev: Oops. I put an appropriate _ for each menu item. 
Is it acceptable to put a mnemonic _ for `("_Line Operations")` even though someone else's plugin/menu item uses _L as a mnemonic character -- or should I just leave that out?

build/lineoperations.m4
smostertdev: Thanks for the suggestion, I dropped the GTK2 enforcement.

lineoperations/src/lineoperations.c
codebrainz: It's not ideal, but it won't break mnemonics, it will just make user have to press it twice to get to the second one.

lineoperations/src/linefunctions.c
codebrainz: > mostly for joining array of lines into new_file string

Not sure if it's applicable here (didn't really look at the code), but if you can enforce the last array element is `NULL`, you could maybe use [`g_strjoinv()`](https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strjoinv) and using `\n` as the `separator` argument.

lineoperations/src/lineoperations.c
b4n: Yeah, I guess it's acceptable (and several plugins already do that).  As you said, it can indeed be suboptimal as another plugin might have the same mnemonic as yours, but as @codebrainz said it wouldn't break things too bad either.
So IMO do as you like.  but maybe @frlan has a stronger opinion on that matter?

lineoperations/README
```diff
@@ -0,0 +1,226 @@
...
+* Remove Unique Lines
+* Remove Empty Lines
+* Remove Whitespace Lines
+* Sort Lines Ascending
+* Sort Lines Descending
+
+Usage
+=====
+
+After the plugins has been installed successfully, load the plugin via Geany's plugin manager and a new menu item in the Tools menu will appear. Click on each menu item to apply the operation on whole  file. See descriptions below to see operations for each menu item.
```
b4n: I would wrap long lines at the 72th/80th column myself.  not sure if we require this though, @frlan

lineoperations/README
```diff
@@ -0,0 +1,226 @@
...
+  ::
+
+    Line 2\n
+    Line 1\n
+    \n          #removed
+         \n     #NOT removed (contains spaces)
+        Line 1\n
+    Line 2\n
+
+  The **Remove Unique Lines** will change the file into this:
```
b4n: shouldn"t this rather be "Remove Empty Lines"?

lineoperations/README
```diff
@@ -0,0 +1,226 @@
...
+    Line 2
+    Line 1
+    Line 2
+    Line 1
+    Line 2
+
+Remove Empty Lines
+------------------
+
+Removes all lines that end with a newline character(s) AND do not have any other characters.
```
b4n: Not sure if this is still true with the Scintilla-based implementation.  I think Scintilla will happily find the last line (without the `\n`) to be an empty one

lineoperations/README
```diff
@@ -0,0 +1,226 @@
...
+Removes all lines that have only white space characters.
+
+A **whitespace character** is one of:
+
+1. `' '` : space
+2. `'\t'`: horizontal tab
+3. `'\f'`: form feed
+4. `'\v'`: vertical tab
+5. `'\r'`: cartridge return
+6. `'\n'`: newline character
```
b4n: does this still holds with the Scintilla-based version?

lineoperations/README
```diff
@@ -0,0 +1,226 @@
...
+
+    Line 2\n
+    Line 1\n
+    \n           #removed
+         \n      #removed (contains only whitespace chars)
+     \t \n       #removed (contains only whitespace chars)
+        Line 1\n #NOT removed (contains non whitespace chars)
+    Line 2\n
+
+  The **Remove Unique Lines** will change the file into this:
```
b4n: Here too, reference should probably be "Remove Whitespace Lines"

lineoperations/src/lineoperations.c
b4n: FWIW, this one still holds

b4n: (even more so as you depend on API 225 which provides the new-style plugin API :))

lineoperations/README
frlan: Yes, I'd prefer it too but wouldn't enforce it. So up to you ;)

lineoperations/src/linefunctions.c
smostertdev: I have `g_strjoinv()` in `sortlines()`, but any other operation require some sort of filtering (e.g. remove duplicate sorted fns, sorts lines then takes first occurrences)

lineoperations/src/lineoperations.c
smostertdev: The original API fix I made was not working, so I just kept the old version of this file here (and forgot I bumped to 225).
But good news, this is now changed to the new plugin API in the commits today (assuming I understand the new plugin changes correctly).

lineoperations/README
smostertdev: I checked  Remove Empty Lines and Remove Whitespace Lines, and it looks like they handle these situations fine.

lineoperations/README
smostertdev: no, I removed this (I don't know what characters Geany/Scintilla considers whitespace with `sci_get_line_indent_position()`).

lineoperations/src/lineoperations.c
```diff
@@ -0,0 +1,203 @@
...
+	ui_add_document_sensitive(rmwhspln_item);
+	ui_add_document_sensitive(sortasc_item);
+	ui_add_document_sensitive(sortdesc_item);
+
+	return TRUE;
+}
+
+
+void
+lo_cleanup(void)
```
b4n: should be static, and signature should be `static void lo_cleanup(GeanyPlugin *plugin, gpointer pdata)`

b4n: See http://www.geany.org/manual/reference/structGeanyPluginFuncs.html

b4n: (and http://www.geany.org/manual/reference/howto.html)

lineoperations/src/lineoperations.c
```diff
@@ -0,0 +1,203 @@
...
+}
+
+
+G_MODULE_EXPORT
+void geany_load_module(GeanyPlugin *plugin)
+{
+    plugin->info->name        = _("Line Operations");
+    plugin->info->description = _("Line Operations provides a handful of functions that can be applied to a document such as, removing duplicate lines, removing empty lines, removing lines with only whitespace, and sorting lines.");
+    plugin->info->version     = "0.1";
+    plugin->info->author      = _("Sylvan Mostert <smostert.dev@gmail.com>");
```
b4n: To allow those strings actually be displayed translated, you should call `main_locale_init()` before filling the plugin info (well, before any call to `_()`).  See http://www.geany.org/manual/reference/howto.html#translatable_plugin_information

Note: `main_loacle_init()` only has to be called once per plugin run, so calling it here will be enough for the whole plugin, no need to call it again in `init()` or anything.

lineoperations/src/lineoperations.c
b4n: Great :)  Changes are almost good, commented on the few things to fix.

BTW, this new API's docs are quite new, so you'd be most welcome to report what was unclear or lacking, so we can improve it :)

lineoperations/src/lineoperations.c
smostertdev: Awesome, it should be all up-to-date now.

(Both 2c2cfac and 0241373 are the mostly the same as before. I updated 2c2cfac, so they're showing that they are new. (but they're not really))

lineoperations/src/lineoperations.c
smostertdev: Yeah, I didn't find too many issues with it, but I'll read the page through again and see if there is anything to improve.

By the way, thank you very much for all of the feedback on this plugin, much appreciated. If you have any more tips (even if they are minor) I'll be glad to hear them.

lineoperations/src/lineoperations.c
```diff
@@ -0,0 +1,205 @@
...
+	gtk_menu_item_set_submenu(GTK_MENU_ITEM(main_menu_item), submenu);
+
+	gtk_container_add(GTK_CONTAINER(geany->main_widgets->tools_menu), main_menu_item);
+
+
+	g_signal_connect(rmdupst_item, "activate", G_CALLBACK(action_rmdupst_item), NULL);
+	g_signal_connect(rmdupln_item, "activate", G_CALLBACK(action_rmdupln_item), NULL);
+	g_signal_connect(rmunqln_item, "activate", G_CALLBACK(action_rmunqln_item), NULL);
+	g_signal_connect(rmemtyln_item, "activate", G_CALLBACK(action_rmemtyln_item), NULL);
+	g_signal_connect(rmwhspln_item, "activate", G_CALLBACK(action_rmwhspln_item), NULL);
```
b4n: you could lower duplication of the wrapper callbacks by having a single callback that calls the appropriate function (either passed as parameter [1] or referenced as an ID or something), as they all have they all behave the same (but the sort ones).
You may or may not want this, it's merely a suggestion.  And it's even a bad one if you expect to alter the various functions to receive different parameters in the future.

[1] technically in plain C you're not supposed to convert a function pointer to a data pointer.  However, GTK depends on this and it effectively works on all supported platforms (if not all platforms in general).

lineoperations/src/lineoperations.c
```diff
@@ -0,0 +1,205 @@
...
+	g_signal_connect(sortasc_item, "activate", G_CALLBACK(action_sortasc_item), NULL);
+	g_signal_connect(sortdesc_item, "activate", G_CALLBACK(action_sortdesc_item), NULL);
```
b4n: Similarly (but more straightforward), you could pass the flag whether to sort ascending or descending through the user data parameter.

lineoperations/src/lineoperations.c
```diff
@@ -0,0 +1,205 @@
...
+	gtk_widget_show(submenu);
+	sep1 = gtk_separator_menu_item_new();
+	sep2 = gtk_separator_menu_item_new();
+	rmdupst_item  = gtk_menu_item_new_with_mnemonic(_("Remove Duplicate Lines, _Sorted"));
+	rmdupln_item  = gtk_menu_item_new_with_mnemonic(_("Remove Duplicate Lines, _Ordered"));
+	rmunqln_item  = gtk_menu_item_new_with_mnemonic(_("Remove _Unique Lines"));
+	rmemtyln_item = gtk_menu_item_new_with_mnemonic(_("Remove _Empty Lines"));
+	rmwhspln_item = gtk_menu_item_new_with_mnemonic(_("Remove _Whitespace Lines"));
+	sortasc_item  = gtk_menu_item_new_with_mnemonic(_("Sort Lines _Ascending"));
+	sortdesc_item = gtk_menu_item_new_with_mnemonic(_("Sort Lines _Descending"));
```
b4n: similarly, there's quite a lot of duplication in the menu building code, while it's basically always exactly the same.  you could easily lower duplication with e.g. representing the items to add like that:
```C

	struct
	{
		const gchar *label;
		GCallback cb_activate;
		gpointer cb_data;
	}
	menu_items[] =
	{
		{ N_("Remove Duplicate Lines, _Sorted"),
		  G_CALLBACK(action_func_item), rmdupst },
		{ N_("Remove Duplicate Lines, _Ordered"),
		  G_CALLBACK(action_func_item), rmdupln },
		{ N_("Remove _Unique Lines"),
		  G_CALLBACK(action_func_item), rmunqln },
		{ NULL },
		{ N_("Remove _Empty Lines"),
		  G_CALLBACK(action_func_item), rmemtyln },
		{ N_("Remove _Whitespace Lines"),
		  G_CALLBACK(action_func_item), rmwhspln },
		{ NULL },
		{ N_("Sort Lines _Ascending"),
		  G_CALLBACK(action_sort_item), GINT_TO_POINTER(1) },
		{ N_("Sort Lines _Descending"),
		  G_CALLBACK(action_sort_item), GINT_TO_POINTER(0) }
	};
```
and adding them in a loop
```C
for (i = 0; i < G_N_ELEMENTS(menu_items); i++)
	{
		GtkWidget *item;

		if (! menu_items[i].label) /* separator */
			item = gtk_separator_menu_item_new();
		else
		{
			item = gtk_menu_item_new_with_mnemonic(_(menu_items[i].label));
			g_signal_connect(item, "activate", menu_items[i].cb_activate, menu_items[i].cb_data);
			ui_add_document_sensitive(item);
		}

		gtk_widget_show(item);
		gtk_menu_shell_append(GTK_MENU_SHELL(submenu), item);
	}
```

lineoperations/src/lineoperations.c
```diff
@@ -0,0 +1,205 @@
...
+
+    plugin->info->name        = _("Line Operations");
+    plugin->info->description = _("Line Operations provides a handful of functions that can be applied to a document such as, removing duplicate lines, removing empty lines, removing lines with only whitespace, and sorting lines.");
+    plugin->info->version     = "0.1";
+    plugin->info->author      = _("Sylvan Mostert <smostert.dev@gmail.com>");
+
+    plugin->funcs->init       = lo_init;
+    plugin->funcs->cleanup    = lo_cleanup;
+
+    GEANY_PLUGIN_REGISTER(plugin, 225);
```
b4n: indentation is odd (4 spaces here instead of 1 tab everywhere else)

lineoperations/src/lineoperations.c
b4n: Done in https://github.com/b4n/geany-plugins/commit/de106a6b044d8296858e1559a7b402fa8424903f if you're interested.  It doesn't depend on altering the callbacks so can be applied directly.

lineoperations/src/lineoperations.c
b4n: Done in https://github.com/b4n/geany-plugins/commit/f05a05b67e8451d619ec90aa41ebb7c54f6b0039 if you're interested.  The patch however depends on the simplified menu building, so you might want to apply it afterward, or just use this as inspiration (or not at all).

lineoperations/src/linefunctions.c
```diff
@@ -0,0 +1,328 @@
...
+void sortlines(GeanyDocument *doc, gboolean asc) {
+	gint  total_num_chars;  /* number of characters in the document */
+	gint  total_num_lines;  /* number of lines in the document */
+	gchar **lines;          /* array to hold all lines in the document */
+	gchar *new_file;        /* *final* string to replace current document */
+	gint  i;                /* iterator */
+
+	total_num_chars = sci_get_length(doc->editor->sci);
+	total_num_lines = sci_get_line_count(doc->editor->sci);
+	lines           = g_malloc(sizeof(gchar *) * total_num_lines+1);
```
b4n: there's a subtle bug here: the +1 is not applied correctly as `*` has priority over `+`.  Wrap in parentheses. (thanks to Valgrind for finding this)

b4n: Fix available at https://github.com/b4n/geany-plugins/commit/ee354655993507b7b998a3c13e4417ea0b1286af

lineoperations/src/lineoperations.c
smostertdev: Nice, I added these commits. It definitely makes it look a lot cleaner.

lineoperations/src/lineoperations.c
```diff
@@ -0,0 +1,134 @@
...
+
+G_MODULE_EXPORT
+void geany_load_module(GeanyPlugin *plugin)
+{
+	main_locale_init(LOCALEDIR, GETTEXT_PACKAGE);
+
+	plugin->info->name        = _("Line Operations");
+	plugin->info->description = _("Line Operations provides a handful of functions that can be applied to a document such as, removing duplicate lines, removing empty lines, removing lines with only whitespace, and sorting lines.");
+	plugin->info->version     = "0.1";
+	plugin->info->author      = _("Sylvan Mostert <smostert.dev@gmail.com>");
```
b4n: there probably is no need to translate your name :)

b4n:
LGTM

Issue #312, opened 2015-11-23
[GeniusPaste] possibility to select private pastebin service


Komosa:
I will be nice to have possibility to use own pastebin service, such as https://github.com/seejohnrun/haste-server

If haste-server API is different that any of supported ones, it still be very usefull to have possibility to write own wrapper and provide URI of that wrapper to plugin.

PR #338, opened 2016-01-16
[GeniusPaste] Rewrite to make pastebins customizable

b4n:
Rewrite the paste logic to use customizable and configurable pastebin services through configuration files, fixing #312.  As a showcase, this PR adds support for http://fpaste.org quite painlessly in a7bf3a73b85aa8b136aa6905c2ec0905d111f76f.

**Disclaimer:** this has not yet been tested under Windows, and I suspect that looking up the path to the system configuration directory holding the pastebin service configurations shipped with the plugin might not work.  This should be tested.


b4n:
Fixed a few details and added a neat feature while trying to add support for http://paste.debian.net/ (a459a192f62ad21b02f353fb6a7e8d2e5111b94c): support for using the pastebin service's HTTP redirection response as paste URL (649939a4490034e51ff69e02d5380f8b7664d882).  It's a lot neater than grepping some more or less random web page when using a non-API pastebin (e.g. when simulating a browser rather than using a well defined API).

geniuspaste/data/tinypaste.com.conf
```diff
@@ -0,0 +1,20 @@
+# FIXME: outdated, this leads to a 404 on http://pasted.co/
```
b4n: Yes, this doens't work.  It didn't before my changes either, I guess the service changed since it was implemented.
I couldn't find how to use it, and their link to a supposed API documentation leads to a 404.

I'd delete this pastebin configuration altogether if I was the only caller, but for now I kept it in the hope maybe someone knows how to fix it.

b4n: So in the end, it's merely ported from the previous hard-coded logic, with zero testing :)

frlan:
Looks fine to me. Once @Enrix835  is giving ok, I'd like to merge it.

PR #340, opened 2016-01-18
lineoperations: Fix translations

b4n:
Fix translations by using the correct text domain.

### DIFF
--- a/lineoperations/src/lineoperations.c
+++ b/lineoperations/src/lineoperations.c
@@ -19,6 +19,8 @@
 */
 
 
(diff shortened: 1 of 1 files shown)


b4n:
@smostertdev

smostertdev:
Yeah looks good to me.
Could you also quickly remove the translation from `plugin->info->author      = _("Sylvan Mostert <smostert.dev@gmail.com>");`.
Then every thing should be perfect.

frlan:
Chery-picked and translation make removed.

PR #341, opened 2016-01-18
GeanyCtags: Find Tag dialog UI fixes

b4n:
Add mnemonics to two labels in the Find Tag dialog that already have a mnemonic widget but no actual mnemonics.

Also, fix the match type labels to be translatable.


frlan:
Looks good ;)

techee:
Great, thanks!

PR #343, opened 2016-01-25
Updated translations

pmralbuquerque:


pmralbuquerque:
Like in Geany, I think it's rebased.

frlan:
I needed to revert the changes as it also was corrupt..

PR #351, opened 2016-02-17
projectorganizer: Use the word "symbol" instead of "tag"

techee:
Do the same for ProjectOrganizer as for Geany in https://github.com/geany/geany/pull/582

### DIFF
--- a/projectorganizer/src/prjorg-project.c
+++ b/projectorganizer/src/prjorg-project.c
@@ -620,16 +620,16 @@ gint prjorg_project_add_properties_tab(GtkWidget *notebook)
 	gtk_entry_set_text(GTK_ENTRY(e->ignored_dirs_patterns), str);
 	g_free(str);
 
-	label = gtk_label_new(_("Generate tags for all project files:"));
+	label = gtk_label_new(_("Index all project files:"));
 	gtk_misc_set_alignment(GTK_MISC(label), 0, 0);
 	e->generate_tag_prefs = gtk_combo_box_text_new();
-	gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(e->generate_tag_prefs), _("Auto (generate if less than 300 files)"));
+	gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(e->generate_tag_prefs), _("Auto (index if less than 300 files)"));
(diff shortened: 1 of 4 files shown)


projectorganizer/src/prjorg-project.c
```diff
@@ -620,7 +620,7 @@ gint prjorg_project_add_properties_tab(GtkWidget *notebook)
 	gtk_entry_set_text(GTK_ENTRY(e->ignored_dirs_patterns), str);
 	g_free(str);
 
-	label = gtk_label_new(_("Generate tags for all project files:"));
+	label = gtk_label_new(_("Parse symbols for all project files:"));
```
b4n: I'm not sure if the new sentence is correct.  I'd rather say _Parse symbols **in** all project files_ or _**from**_ rather than _**for**_. @elextr ?

b4n:
Otherwise, LGTM

projectorganizer/src/prjorg-project.c
techee: Yeah, and actually we don't want to "parse symbols" but rather files. I'd just simplify it to "Index all project files"

projectorganizer/src/prjorg-project.c
```diff
@@ -620,16 +620,16 @@ gint prjorg_project_add_properties_tab(GtkWidget *notebook)
 	gtk_entry_set_text(GTK_ENTRY(e->ignored_dirs_patterns), str);
 	g_free(str);
 
-	label = gtk_label_new(_("Generate tags for all project files:"));
+	label = gtk_label_new(_("Index all project files:"));
 	gtk_misc_set_alignment(GTK_MISC(label), 0, 0);
 	e->generate_tag_prefs = gtk_combo_box_text_new();
-	gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(e->generate_tag_prefs), _("Auto (generate if less than 300 files)"));
+	gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(e->generate_tag_prefs), _("Auto (index if less than 300 files)"));
```
b4n: BTW, it says 300 here, but 500 in the docs.

techee:
@b4n Thanks, fixed. As the linguistic problem disappeared because I renamed the label, I guess the patch is OK now so merging.

PR #352, opened 2016-02-18
Macro plugin compatibility for both gtk2 and gtk3 [Needs Testing]

sagarchalise:
This is based on a patch. It builds on gtk3 as well as gtk2 but not thoroughly tested.


geanymacro/src/geanymacro.c
```diff
@@ -1387,21 +1386,26 @@ static void EditSearchOptions(GtkTreeModel *model,GtkTreeIter *iter)
 	/* create dialog box */
 	dialog=gtk_dialog_new_with_buttons(_("Search Options:"),
 	                                   GTK_WINDOW(geany->main_widgets->window),
-	                                   GTK_DIALOG_DESTROY_WITH_PARENT,NULL);
+	                                   GTK_DIALOG_DESTROY_WITH_PARENT,
+_("Ok"),
```
b4n: Mnemonic should not be removed (unless actually wanted)

geanymacro/src/geanymacro.c
```diff
@@ -1387,21 +1386,26 @@ static void EditSearchOptions(GtkTreeModel *model,GtkTreeIter *iter)
...
+        GTK_RESPONSE_OK,
+      _("Cancel"),
```
b4n: ditto

geanymacro/src/geanymacro.c
```diff
@@ -1733,9 +1736,12 @@ static void EditMacroElements(Macro *m)
...
 	cTitle=g_strdup_printf(_("Edit: %s"),m->name);
-	dialog=gtk_dialog_new_with_buttons(cTitle,GTK_WINDOW(geany->main_widgets->window),
-		GTK_DIALOG_DESTROY_WITH_PARENT,NULL);
-
+	dialog=gtk_dialog_new();
+    //_with_buttons(cTitle,GTK_WINDOW(geany->main_widgets->window),
+		//GTK_DIALOG_DESTROY_WITH_PARENT,NULL);
+    gtk_window_set_title(GTK_WINDOW(dialog), cTitle);
+    gtk_window_set_transient_for(GTK_WINDOW(dialog), GTK_WINDOW(geany->main_widgets->window));
+    gtk_window_set_destroy_with_parent(GTK_WINDOW(dialog), TRUE);
```
b4n: what was the problem with the `_with_buttons()` call, but with 0 buttons?

b4n:
Indentation has some problems, and some code should be dropped rather than commented out.

Also, did you check how the widgets expand?  `gtk_conatiner_add()` changed behavior to that regard in GTK3, where it doesn't expand by default.

sagarchalise:
@b4n I guess I did unnecessary changes regarding `gtk_dialog`. Also, I have not tested it thoroughly

b4n:
Hum, according to GitHub there are no changes.  And apparently the version in master already supports GTK3 just fine, because of #410

sagarchalise:
This has already been implemented by #410

PR #356, opened 2016-02-20
geanydoc: Drop use of the most obvious deprecated Geany API

b4n:
Not tested, but builds fine.

### DIFF
--- a/geanydoc/src/geanydoc.c
+++ b/geanydoc/src/geanydoc.c
@@ -35,7 +35,7 @@
 #include "geanydoc.h"
 
 /* These items are set by Geany before init() is called. */
-PluginFields *plugin_fields;
+GeanyPlugin *geany_plugin;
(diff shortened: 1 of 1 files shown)


geanydoc/src/geanydoc.c
```diff
@@ -80,24 +78,17 @@ word_check_right(gchar c)
...
 	gint pos;
 	gint cstart, cend;
 	gchar c;
-	gint text_len;
 
 	doc = document_get_current();
 	g_return_val_if_fail(doc != NULL && doc->file_name != NULL, NULL);
 
-	text_len = sci_get_selected_text_length(doc->editor->sci);
-	if (text_len > 1)
```
techee: Technically this checks for 2-character selections and longer but looks more like a bug than a feature.

techee:
LGBI.

PR #357, opened 2016-02-21
geanyinsertnum: Drop use of the most obvious deprecated Geany API

b4n:
Not really tested, but should not be a big deal as it's only `s/ui_widget_set_tooltip_text/gtk_widget_set_tooltip_text/`, which are synonyms nowadays.


techee:
LGBI.

PR #359, opened 2016-02-21
pairtaghighlighter: Drop use of the most obvious deprecated Geany API

b4n:
Not really tested.

@vmkononenko


vmkononenko:
Verified. Provided change does not introduce a regression and can be merged.

PR #360, opened 2016-02-21
geanyminiscript: Drop use of the most obvious deprecated Geany API

b4n:
Not tested.

@earshinov


b4n:
I just added a commit not to access private Geany fields, as pointed out in geany/geany#910

techee:
LGBI.

PR #365, opened 2016-02-21
projectorganizer: Drop use of the most obvious deprecated Geany API

b4n:
Not thoroughly tested but should be quite straightforward


b4n:
@techee geany/geany#911 ;)

PR #366, opened 2016-02-21
Shiftcolumn: Geany deprecation fixes + a simple fix

b4n:
Quickly tested.

Rectangular selection handling is still broken, but nothing different than before the patches.


techee:
LGBI

PR #372, opened 2016-02-21
Use new spawn API to fix broken Win32 ctags.exe command execution

eht16:
Since the new spawn API, which is used by utils_spawn_sync() as well,
passing a whole command line in argv[0] doesn't work any longer.
So use the new spawn API directly which conveniently offers to
pass the command either as command line or as argv.

Tested on Linux and Windows.
I bumped the required API version. It seems unfortunately we didn't bump the API version in Geany when merging the spawn API but the currently used 226 is newer than the spawn API and is already available in the last Geany release, so it should be ok.

### DIFF
--- a/geanyctags/src/geanyctags.c
+++ b/geanyctags/src/geanyctags.c
@@ -46,7 +46,7 @@
 #endif
 
 
-PLUGIN_VERSION_CHECK(211)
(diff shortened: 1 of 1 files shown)


techee:
@eht16 Ah, missed this one. Works fine on linux and I trust you on Windows :-)

PR #371, opened 2016-02-21
Set G_LOG_DOMAIN for proper GLib error reporting/logging

eht16:
This turns messages like:
```
** INFO: User plugins: /home/enrico/tmp/geany_conf/plugins/geanypy/plugins
** INFO: System plugins: /home/enrico/apps/share/geany/geanypy/plugins
```
into
```
GeanyPy-INFO: User plugins: /home/enrico/tmp/geany_conf/plugins/geanypy/plugins
GeanyPy-INFO: System plugins: /home/enrico/apps/share/geany/geanypy/plugins
```

PR #370, opened 2016-02-21
Use GLib error reporting/logging instead of fprintf()

eht16:
This makes logging and error reporting in the plugin more consistent
with other plugins and Geany.
Also set the G_LOG_DOMAIN for the plugin in order to have proper
log messages.

PR #373, opened 2016-02-27
Define G_LOG_DOMAIN for all plugins where it is missing

eht16:
This helps to get proper log messages prefixed with the plugin name
in Geany's Help->Debug Messages view as well as on console
logging.

For some plugins this might be unnecessary if they don't use any `g_message()`, `g_warning()` or similar functions but `G_LOG_DOMAIN` is also used by `g_return_if_fail()` and friends.
Defining it for all plugins at least reduces the chance to miss some plugin, also for the future.

Omitted the plugins GeanyPy and GeanyPG where seperate PRs exist (#371 and #370).

### DIFF
--- a/geanydoc/src/Makefile.am
+++ b/geanydoc/src/Makefile.am
@@ -7,7 +7,7 @@ geanydoc_la_SOURCES = \
 	geanydoc.c \
 	geanydoc.h
 
-geanydoc_la_CFLAGS = $(AM_CFLAGS)
+geanydoc_la_CPPFLAGS = $(AM_CPPFLAGS) -DG_LOG_DOMAIN=\"GeanyDoc\"
 geanydoc_la_LIBADD = $(COMMONLIBS)
 
 include $(top_srcdir)/build/cppcheck.mk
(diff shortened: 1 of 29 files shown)


geanydoc/src/Makefile.am
```diff
@@ -7,7 +7,7 @@ geanydoc_la_SOURCES = \
 	geanydoc.c \
 	geanydoc.h
 
-geanydoc_la_CFLAGS = $(AM_CFLAGS)
+geanydoc_la_CPPFLAGS = $(AM_CPPFLAGS) -DG_LOG_DOMAIN=\"GeanyDoc\"
```
eht16: I replaced `geanydoc_la_CFLAGS` by accident. But I think this line was useless already before because $(AM_CFLAGS) is probably the default. If I'm wrong, I'd repush without that line removed.

b4n: `AM_CFLAGS` is the default for when there is no traget-specific `CFLAGS`, so yeah should be fine.

PR #375, opened 2016-02-28
Add shell script to add plugin dependencies into GTK bundle on MSYS2

eht16:
This is a helper script to easily fetch and extract plugin dependencies
on Windows using the MSYS2 environment.

Inspired and based on @kugel- 's initial script in https://github.com/geany/geany/pull/560.

Currently it installs the dependencies we had already in the past (enchant, lua, libgit2, libsoup, ctpl, libxml2) and a few new ones which we got for free from MSYS2 (webkitgtk, ctags, gnupg/gpgme).
This script works like the one from Geany but only extracts additional dependencies to be merged into the bundle created for Geany.

### DIFF
--- a/build/gtk-bundle-from-msys2.sh
+++ b/build/gtk-bundle-from-msys2.sh
@@ -1,0 +1,289 @@
+#!/bin/sh
+#
+# Fetch and extract Geany-Plugins dependencies for Windows/MSYS2
+# This script will download (or use Pacman's cache) to extract
+# plugin dependencies as defined below. To be run within a MSYS2
+# shell. The extracted files will be placed into the current
+# directory.
+
+ABI=i686
+use_cache="no"
+make_zip="no"
+gtkv="3"
+
+# ctags - binary for GeanyCTags plugin
+# ctpl-git - for GeanyGenDoc plugin
+# enchant, hunspell - for SpellCheck plugin
+# curl, glib-networking, gnutls, icu, sqlite3, webkitgtk2/3 for WebHelper and Markdown plugins
+# lua51 - for GeanyLua plugin
+# gnupg, gpgme - for GeanyPG plugin
+# libsoup - for UpdateChecker plugin
+# libgit2 - for GitChangeBar plugin
+# libxml2 - for PrettyPrinter plugin
+# gtkspell - for GeanyVC plugin
+# the rest is dependency-dependency
+packages="
+ctags
+ctpl-git
+curl
+dbus
+dbus-glib
+enchant
+geoclue
+giflib
+glib-networking
+gmp
+gnupg
+gnutls
+gpgme
+gstreamer
+gst-plugins-base
+gtkspell
+http-parser
+hunspell
+icu
+libassuan
+libgcrypt
+libgpg-error
+libgit2
+libidn
+libjpeg-turbo
+libogg
+libsoup
+libssh2
+libsystre
+libtasn1
+libtheora
+libtiff
+libtre-git
+libvorbis
+libvorbisidec-svn
+libwebp
+libxml2
+libxslt
+lua51
+nettle
+openssl
+orc
+p11-kit
+readline
+rtmpdump-git
+sqlite3
+termcap
+xz
+"
+
+gtk2_dependency_pkgs="
+webkitgtk2
+"
+gtk3_dependency_pkgs="
+webkitgtk3
+"
+
+package_urls=""
+
+
+handle_command_line_options() {
+	for opt in "$@"; do
+		case "$opt" in
+		"-c"|"--cache")
+			use_cache="yes"
+			;;
+		"-z"|"--zip")
+			make_zip="yes"
+			;;
+		"-2")
+			gtkv="2"
+			;;
+		"-3")
+			gtkv="3"
+			;;
+		"-h"|"--help")
+			echo "gtk-bundle-from-msys2.sh [-c] [-h] [-z] [-2 | -3] [CACHEDIR]"
+			echo "      -c Use pacman cache. Otherwise pacman will download"
+			echo "         archive files"
+			echo "      -h Show this help screen"
+			echo "      -z Create a zip afterwards"
+			echo "      -2 Prefer gtk2"
+			echo "      -3 Prefer gtk3"
+			echo "CACHEDIR Directory where to look for cached packages (default: /var/cache/pacman/pkg)"
+			exit 1
+			;;
+		*)
+			cachedir="$opt"
+			;;
+		esac
+	done
+}
+
+initialize() {
+	if [ -z "$cachedir" ]; then
+		cachedir="/var/cache/pacman/pkg"
+	fi
+
+	if [ "$use_cache" = "yes" ] && ! [ -d "$cachedir" ]; then
+		echo "Cache dir \"$cachedir\" not a directory"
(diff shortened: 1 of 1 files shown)


eht16:
As you can see, the list of dependencies is quite long. Probably 80% of the list are for webkitgtk.
Do we want to include webkitgtk into the plugins installer?
I'd vote yes as it brings the Webhelper and Markdown plugins on Windows. Though the uncompressed G-P dependencies then take about 90 MB (I don't know yet the compressed installer size nor the size without webkitgtk and its dependencies).

eht16:
One more note: the whole dependency list is only tested with GTK2.
I can't spend time testing GTK3 so close before the release. Might be done after 1.27.

build/gtk-bundle-from-msys2.sh
```diff
@@ -0,0 +1,259 @@
...
+	echo "Cleanup unnecessary files"
+	# etc: cleanup development files
+	rm -rf etc/bash_completion.d
+	rm -rf etc/dbus-1
+	rm -rf etc/pkcs11
+	rm -rf etc/xml
+	# include: cleanup development files
+	rm -rf include
+	# lib: cleanup development files
+	rm -rf libexec
```
b4n: libexec is not generally dev files, is it?

build/gtk-bundle-from-msys2.sh
```diff
@@ -0,0 +1,259 @@
...
+# ctpl-git - for GeanyGenDoc plugin
+# enchant, hunspell - for SpellCheck plugin
+# curl, glib-networking, gnutls, icu, sqlite3, webkitgtk2/3 for WebHelper and Markdown plugins
+# lua51 - for GeanyLua plugin
+# gnupg, gpgme - for GeanyPG plugin
+# libsoup - for UpdateChecker plugin
+# libgit2 - for GitChangeBar plugin
+# libxml2 - for PrettyPrinter plugin
+# gtkspell - for GeanyVC plugin
+# the rest is dependency-dependency
```
b4n: can't pacman be used to resolve dependencies of deps?  would look more robust if it could

b4n:
> As you can see, the list of dependencies is quite long. Probably 80% of the list are for webkitgtk.
Do we want to include webkitgtk into the plugins installer?
I'd vote yes as it brings the Webhelper and Markdown plugins on Windows. Though the uncompressed G-P dependencies then take about 90 MB (I don't know yet the compressed installer size nor the size without webkitgtk and its dependencies).

Ideally I'd say include it in the installer but allow not to install them with a nice installer option -- e.g. allow disabling specific plugins, or just webkitgtk ones.

build/gtk-bundle-from-msys2.sh
eht16: Yeah, the comment is a bit misleading. Sorry.
```
$ ls -l bundle-plugins/libexec/*
-rwxr-xr-x 1 enrico None  71586 Jul 10  2015 bundle-plugins/libexec/dbus-bash-completion-helper.exe
-rwxr-xr-x 1 enrico None  22546 Jul 10  2015 bundle-plugins/libexec/geoclue-example.exe
-rwxr-xr-x 1 enrico None  26689 Jul 10  2015 bundle-plugins/libexec/geoclue-geonames.exe
-rwxr-xr-x 1 enrico None  54614 Jul 10  2015 bundle-plugins/libexec/geoclue-gsmloc.exe
-rwxr-xr-x 1 enrico None  64544 Jul 10  2015 bundle-plugins/libexec/geoclue-hostip.exe
-rwxr-xr-x 1 enrico None  81757 Jul 10  2015 bundle-plugins/libexec/geoclue-localnet.exe
-rwxr-xr-x 1 enrico None  27452 Jul 10  2015 bundle-plugins/libexec/geoclue-manual.exe
-rwxr-xr-x 1 enrico None 106206 Jul 10  2015 bundle-plugins/libexec/geoclue-master.exe
-rwxr-xr-x 1 enrico None  27775 Jul 10  2015 bundle-plugins/libexec/geoclue-nominatim.exe
-rwxr-xr-x 1 enrico None  78317 Jul 10  2015 bundle-plugins/libexec/geoclue-plazes.exe
-rwxr-xr-x 1 enrico None  78238 Jul 10  2015 bundle-plugins/libexec/geoclue-skyhook.exe
-rwxr-xr-x 1 enrico None  24743 Jul 10  2015 bundle-plugins/libexec/geoclue-yahoo.exe

bundle-plugins/libexec/gstreamer-1.0:
total 20
-rwxr-xr-x 1 enrico None 20137 Feb 10 09:29 gst-plugin-scanner.exe
```
Nothing we need.
Anyway, I'll change the script to let `libexec` alone and instead delete all *.exe. Then the resulting empty directories will be deleted at the end.

build/gtk-bundle-from-msys2.sh
eht16: Yes it can, see the approach in https://github.com/geany/geany/pull/560.
I didn't consider this because at least in Geany's dependency tree it pulled in a lot of things we never need nor depend on. I don't want to imagine what this would pull in for dependency monsters like webkitgtk. The bundle is already quite big...

I don't mind changing this for post-1.27 but for now, we should get it running.

eht16:
> Do we want to include webkitgtk into the plugins installer?
> 
> Ideally I'd say include it in the installer but allow not to install them with a nice installer option -- e.g. allow disabling specific plugins, or just webkitgtk ones.
Sure, would be nice. But also a lot of work in the installer because you
would need to define list of *.dlls and other files which belong to each
plugin. Nothing which seems reasonable for 1.27, IMHO.

b4n:
Yeah I didn't mean it would be easy, that's why I said "ideally" :)  So yeah if it's not easy leave it.
But with today's hard drive sizes, I guess people will be more happy to have additional plugins (we have been asked about WebHelper a few times already) than sad it takes so much place.
Also power users can have fun deleting files they don't need afterward :]

build/gtk-bundle-from-msys2.sh
b4n: fair enough

eht16:
Heh about the `power users` part :).

Btw, I pushed a change regarding handling of `libexec`.

eht16:
Just pushed an update with all command line options and arguments documented and squashed commits together.

build/gtk-bundle-from-msys2.sh
```diff
@@ -0,0 +1,289 @@
...
+}
+
+move_extracted_files() {
+	echo "Move extracted data to destination directory"
+	if [ -d mingw32 ]; then
+		for d in bin etc home include lib libexec locale sbin share ssl var; do
+			if [ -d "mingw32/$d" ]; then
+				rm -rf $d
+				# prevent sporadic 'permission denied' errors on my system, not sure why they happen
+				sleep 0.5
```
b4n: Huhu :]

b4n:
LGBQI (quick inspection ;)

build/gtk-bundle-from-msys2.sh
frlan: Isn't this deleting not only $d?

eht16: As discussed on IRC, it *does* delete only $d which is intended here because mingw32/$d is moved to $d afterwards.

PR #376, opened 2016-02-29
Set docdir on Windows builds

eht16:
And remove unused define PKGLIBDIR and quote paths even.

As continuation of #374 after more testing on Windows.
These should be last bits regarding Waf->Autotools migration for G-P on Windows, so I hope.

As we all know, I'm not an Autotools expert: maybe there is a better way/location to overwrite `docdir` than in `GP_CHECK_MINGW`.

### DIFF
--- a/build/common.m4
+++ b/build/common.m4
@@ -65,7 +65,7 @@ AC_DEFUN([GP_COMMIT_PLUGIN_STATUS],
     GP_STATUS_PLUGIN_ADD([$1], [$m4_tolower(AS_TR_SH(enable_$1))])
 ])
 
-dnl GEANY_CHECK_MINGW
+dnl GP_CHECK_MINGW
 dnl Checks whether we're building for MinGW, and defines appropriate stuff
 dnl if it is the case.
 dnl Most importantly, AM_CODITIONALs MINGW
--- a/build/vars.build.mk
+++ b/build/vars.build.mk
@@ -2,11 +2,11 @@ if MINGW
 LOCAL_AM_CFLAGS = \
 	-DLOCALEDIR=\""share/locale"\" \
 	-DPREFIX=\"\" \
-	-DDOCDIR=\"\" \
-	-DGEANYPLUGINS_DATADIR=\"share\" \
+	-DDOCDIR=\""share/doc/geany-plugins"\" \
+	-DGEANYPLUGINS_DATADIR=\""share"\" \
 	-DPKGDATADIR=\""share/geany-plugins"\" \
 	-DLIBDIR=\""lib"\" \
-	-DPKGLIBDIR=\"\"
+	-DPKGLIBDIR=\""lib/geany-plugins"\"
 else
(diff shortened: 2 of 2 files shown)


eht16:
With https://github.com/eht16/geany-plugins/commit/6389d9e834d1f09f5090859ed060176f3d8ec515 I'm now done with my changes. G-P build fine and installer works properly.

There is still an issue with some plugins not being able to open their local help files in `DOCDIR` but this need to be fixed in the affected plugins.

build/common.m4
```diff
@@ -75,6 +75,7 @@ AC_DEFUN([GP_CHECK_MINGW],
 		*mingw*)
 			AC_DEFINE([WIN32], [1], [we are cross compiling for WIN32])
 			AM_CONDITIONAL([MINGW], true)
+			AC_SUBST([docdir], '${prefix}/doc/geany-plugins')
```
b4n: this dosn't seem to match what `DOCDIR` is set to below, what's up with this?

build/vars.build.mk
```diff
@@ -14,15 +13,14 @@ LOCAL_AM_CFLAGS = \
 	-DDOCDIR=\""$(docdir)"\" \
 	-DGEANYPLUGINS_DATADIR=\""$(datadir)"\" \
 	-DPKGDATADIR=\""$(pkgdatadir)"\" \
-	-DLIBDIR=\""$(libdir)"\" \
-	-DPKGLIBDIR=\""$(pkglibdir)"\"
```
b4n: ok, not used for the moment.  FWIW, would be `"lib/$(PACKAGE)"` on Windows I guess -- or simply `"lib"` as it's already likely to be package-specific

build/common.m4
eht16: Oops, this was just a left-over in this PR. I actually built the provided test binaries without this.
Thanks for noticing it and sorry. Fixed in https://github.com/eht16/geany-plugins/commit/953d478f2cbb1b06d730e421b6673bdfdd4564dd.

build/vars.build.mk
eht16: I have no idea what it would be on Windows as I also don't know what it might be used for on non-Windows.
Do we want to keep it? I couldn't find any reference to it.

eht16:
Btw, I'd like to squash the commits before merging.

build/vars.build.mk
b4n: `pkglibdir` is `$(libdir)/$(PACKAGE)`.

And well, it's a common thing to have but yeah as it's not used I guess we can remove it.

b4n:
Sure.  LGBI now.

build/vars.build.mk
eht16: Ok, even if not used, I changed my mind and re-added PKGLIBDIR, leave it untouched for non-Windows and set it to "lib/geany-plugins" on Windows. Not sure if this is the correct path, we will see once it is used. Then it is least there and visible.

Issue #382, opened 2016-03-05
GeanyFunctions depreciation warnings


techee:
Most of the plugins now complain about GeanyFunctions being deprecated which makes the compilation output rather noisy. I have removed GeanyFunctions from my plugins but this requires bumping API to a post-linkage-cleanup build, which is 224.

To get rid of the warnings there are two options:

1. Remove geany_functions and bump the API requirement for every plugin.
2. Guard the geany_functions declaration with a macro checking the API version.

I'd personally vote for (1) otherwise the deprecated declaration gets never removed. Is it a problem to bump the API requirement? Is it common to build new plugin versions with old Geany versions?

eht16:
I'd also agree for (1).

I've just done this for the Addons and SpellCheck plugins.

Keeping compability with old Geany versions is a nice-to-have I'd say but not really necessary as we try to release Geany-Plugins together with Geany itself. So this should be fine.

b4n:
Agreed, we should rather drop them.  And as it's already in 1.26 even.

techee:
Good, will prepare a patch.

PR #388, opened 2016-03-06
Remove deprecated GeanyFunctions

techee:
I removed the GeanyFunctions *geany_functions declaration and bumped the API for the plugins where the change was trivial. I skipped:

* geanylua (wasn't quite sure if geany_functions isn't exported to lua scripts)
* geanypy (like above)
* multiterm (written in vala and I didn't want to screw things up)

(Mostly) fixes https://github.com/geany/geany-plugins/issues/382


b4n:
LGBI

codenav/src/codenavigation.h
```diff
@@ -42,7 +42,6 @@
 
 /* Last */
 #include "plugindata.h"		/* this defines the plugin API */
-#include "geanyfunctions.h"	/* this wraps geany_functions function pointers */
```
b4n: leads to missing declaration of `utils_str_equal()` #396

shiftcolumn/src/shiftcolumn.c
```diff
@@ -35,7 +35,6 @@
 #include "document.h"
 #include "keybindings.h"
 #include "plugindata.h"
-#include "geanyfunctions.h"
```
b4n: same, with several other functions

codenav/src/codenavigation.h
codebrainz: Should change the code to use `g_strcmp0` or `strcmp` instead of using the plugin API as a redundant utility library :)

b4n: yeah that probably would be better, esp. as it's misused :)

We may want to deprecated `utils_str_equal()` in favor to this at the same time.

PR #389, opened 2016-03-06
Update fr.po

landrok:
`Line 4706` "more easy" should be replaced by "easier". In this case, I think /geanysendmail/src/geanysendmail.c should be updated and other translations files too.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -35,7 +35,7 @@ msgstr "N°"
 #: ../addons/src/ao_bookmarklist.c:350 ../devhelp/src/dhp-object.c:494
 #: ../devhelp/devhelp/dh-window.c:1142
 msgid "Contents"
-msgstr "Contenu"
+msgstr "Contenus"
 
 #: ../addons/src/ao_bookmarklist.c:383 ../treebrowser/src/treebrowser.c:624
 #: ../webhelper/src/gwh-plugin.c:329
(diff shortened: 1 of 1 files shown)


po/fr.po
```diff
@@ -35,7 +35,7 @@ msgstr "N°"
 #: ../addons/src/ao_bookmarklist.c:350 ../devhelp/src/dhp-object.c:494
 #: ../devhelp/devhelp/dh-window.c:1142
 msgid "Contents"
-msgstr "Contenu"
+msgstr "Contenus"
```
b4n: Doubt it's really the plural form in English, but rather the whatever-it-is-named fact this word is always plural when referring to actual stuff.
Would have to triple check in the context of the plugins, but I think singular form in French is the right choice.

po/fr.po
```diff
@@ -4723,8 +4723,8 @@ msgstr "Envoyer par _mail"
 msgid ""
 "Sends the opened file as unzipped attachment by any mailer from your $PATH"
 msgstr ""
-"Envois le fichier ouvert comme pièce jointe non-compressée avec n'importe "
-"quel client."
+"Envoie le fichier ouvert comme pièce jointe non-compressée avec n'importe "
+"quel client depuis votre $PATH"
```
b4n: I think I was the one leaving this part of the sentence because I feel it's heavy and really pointless, but yeah technically it's there.

--

Ceci dit, je driais plutôt "dans votre $PATH", par ce que la question ici est si le client mail est dans un dossier inclu dans la liste `$PATH`, `$PATH` n'est pas le conteneur lui-même.

b4n:
Apart from the remarks, LGTM.  Nice work :)

po/fr.po
landrok: For the DevHelp usage, it is a left panel subtabs label. In this context, twice are possible. So, in the doubt, I rollback this string. Thanks

landrok:
Thanks for the quick and nightly reviews :)

b4n:
LGTM :)

PR #378, opened 2016-02-29
lineoperations: added selection support

smostertdev:
Here are some changes to lineoperations; adding selection support.

If a selection is made in the file, the line operation will be applied to the lines within the selection, rather than the whole file.

Like usual, I'm open to hear any suggestions/improvements.

PR #391, opened 2016-03-07
lineoperations: Fix a memory leak found by cppcheck

b4n:
`new_file` is [assigned again with the return value of `g_strjoinv()`](https://github.com/geany/geany-plugins/pull/391/files#diff-12783b0a2f2a5a415ec6f2505c90e634R315), thus leaking the previous value.

### DIFF
--- a/lineoperations/src/linefunctions.c
+++ b/lineoperations/src/linefunctions.c
@@ -286,17 +286,13 @@ void rmwhspln(GeanyDocument *doc) {
 
 /* Sort Lines Ascending and Descending */
 void sortlines(GeanyDocument *doc, gboolean asc) {
-	gint  total_num_chars;  /* number of characters in the document */
 	gint  total_num_lines;  /* number of lines in the document */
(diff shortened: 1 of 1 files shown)


b4n:
@smostertdev

lineoperations/src/linefunctions.c
```diff
@@ -295,8 +295,6 @@ void sortlines(GeanyDocument *doc, gboolean asc) {
 	total_num_chars = sci_get_length(doc->editor->sci);
```
smostertdev: We can remove `total_num_chars` from this function. (and on line 289)

smostertdev:
Wow, nice catch. :)

Note to self: I think I removed this, but I need to remember to double check in #378.

frlan:
Can you provide an updated patchset?

lineoperations/src/linefunctions.c
b4n: Good catch :) and if I was more careful yesterday I would have seen my compiler told me.  Updated.

PR #393, opened 2016-03-07
spellcheck: Use <> to include Geany headers

b4n:
Apparently some older versions of cppcheck get stupid when passed extra
flags like `-D`, so try and make them happy by showing them those are
library (system-wide) headers rather than application (local) ones.

### DIFF
--- a/spellcheck/src/gui.c
+++ b/spellcheck/src/gui.c
@@ -26,7 +26,7 @@
 # include "config.h"
 #endif
 
-#include "geanyplugin.h"
+#include <geanyplugin.h>
 
 #include <ctype.h>
 #include <string.h>
--- a/spellcheck/src/scplugin.c
+++ b/spellcheck/src/scplugin.c
@@ -27,7 +27,7 @@
 # include "config.h"
 #endif
 
-#include "geanyplugin.h"
+#include <geanyplugin.h>
 
 
 #include "scplugin.h"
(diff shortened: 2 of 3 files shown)


b4n:
Err, there's a typo in the plugin's name, funny for a spellcheck-related fix :)

b4n:
fixed the typo

eht16:
I'm happy with merging this.
Still being curious: where is the typo you mentioned? In the diff? In the title of this PR, then it is not fixed :).

And more related to the changes themselves: shouldn't other plugins which use `#include "geanyplugin.h"` also be affected? Or is there something special in SpellCheck from `cppcheck`'s POV?

b4n:
> Still being curious: where is the typo you mentioned? In the diff? In the title of this PR, then it is not fixed :).

It was the commit message, but I forgot to update the title (which was autogenerated from the commit) :)

> And more related to the changes themselves: shouldn't other plugins which use `#include "geanyplugin.h"` also be affected? Or is there something special in SpellCheck from `cppcheck`'s POV?

Apparently other plugins are not affected, yet I can't really tell you why.  All I can guess is that, as stated in the commit message, when you pass defines options to cppheck (`-Dsomething`) it starts requiring all `""` headers to be found -- maybe it tries running a preprocessor or something in this case?  As SpellCheck is the only plugin passing a `-D` option to cppcheck, it's the only one affected.

PR #396, opened 2016-03-09
Fix implicit function declarations by including appropriate headers

b4n:
Introduced my legacy inclusion removal in 25e3c8851a5113423ce16a6b8008.


techee:
Yeah, of course. Thanks for the fix.

PR #397, opened 2016-03-10
scope: Ignore a false positive from cppcheck

b4n:
http://trac.cppcheck.net/ticket/7243

### DIFF
--- a/scope/src/Makefile.am
+++ b/scope/src/Makefile.am
@@ -59,4 +59,7 @@ scope_la_CFLAGS = $(AM_CFLAGS) $(VTE_CFLAGS) \
 	-DPLUGINHTMLDOCDIR=\"$(plugindocdir)/html\" \
 	-Wno-shadow
 
+# http://trac.cppcheck.net/ticket/7243
+AM_CPPCHECKFLAGS = --suppress='unknownEvaluationOrder:$(srcdir)/store/scptreestore.c:909'
+
 include $(top_srcdir)/build/cppcheck.mk


b4n:
If you want to "fix" the issue, I guess you could do that (untested):
```diff
diff --git a/scope/src/store/scptreestore.c b/scope/src/store/scptreestore.c
index 4ee0f45..c659560 100644
--- a/scope/src/store/scptreestore.c
+++ b/scope/src/store/scptreestore.c
@@ -906,7 +906,7 @@ GtkTreePath *scp_tree_store_get_path(VALIDATE_ONLY ScpTreeStore *store, GtkTreeI
        {
                gtk_tree_path_append_index(path, ITER_INDEX(iter));
 
-               while ((elem = elem->parent), elem->parent)
+               for (elem = elem->parent; elem->parent; elem = elem->parent)
                {
                        gint index = scp_ptr_array_find(elem->parent->children, elem);
 
```
But that's not a but so…

b4n:
FTR, quoting [*ISO/IEC 9899:201x (C11), N1570*](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf):

> ## 6.5.17 Comma operator
### Syntax
```
	expression:
		assignment-expression
		expression , assignment-expression
```
### Semantics
The left operand of a comma operator is evaluated as a void expression; there is a sequence point between its evaluation and that of the right operand. Then the right operand is evaluated; the result has its type and value. <sup>114)</sup>
>
<small>EXAMPLE As indicated by the syntax, the comma operator (as described in this subclause) cannot appear in contexts where a comma is used to separate items in a list (such as arguments to functions or lists of initializers). On the other hand, it can be used within a parenthesized expression or within the second expression of a conditional operator in such contexts. In the function call
```
	f(a, (t=3, t+2), c)
```
the function has three arguments, the second of which has the value 5.</small>

frlan:
I think I will merge this and leave the point open for plugin maintainer.

PR #400, opened 2016-03-11
geanypy: fix loading plugins from user config directory.

kugel-:
Have to add ~/.config/geany/plugins to Python's path otherwise loading plugins from there
fails.

Tested with a relatively complex plugin which also ships extra modules in subdirectory:
https://github.com/sagarchalise/geanypy-reStructured-preview


kugel-:
@frlan @elextr important bug fix for 1.27

frlan:
LGTM

PR #402, opened 2016-03-12
geniuspaste: Add Windows-specific logic for locating system data files

b4n:
This fixes locating pastebin configuration files shipped with the
plugin under Windows (those installed in the plugin's data directory
rather than in the user's configuration directory).

---

Quickly tested and working, but I wouldn't trust it with my life.

---
Probably unrelated, but the plugin crashes when clicking the link in the popup dialog after a successful paste.  This relies on GTK's URL opening (all the plugin does is setting markup using `<a href="...">`in a dialog), and the crash might or might not be related to the fact I cross-built the plugin with a different GTK version (works on my Linux machine).  Once I figure out how to setup everything to actually build it on Windows I could try again -- or someone else could try the patch, who knows @eht16 :)
However, if the plugin is configured to open the URI without asking (and then using `utils_open_browser()`), it works fine.

### DIFF
--- a/geniuspaste/src/geniuspaste.c
+++ b/geniuspaste/src/geniuspaste.c
@@ -94,6 +94,54 @@ PLUGIN_SET_TRANSLATABLE_INFO(LOCALEDIR, GETTEXT_PACKAGE, PLUGIN_NAME,
                              PLUGIN_VERSION, "Enrico Trotta <enrico.trt@gmail.com>")
 
 
+/* Windows compatibility for resolving system plugin data path.
+ * This only works if the plugin's data is installed in the same tree as the
+ * plugin DLL, which might not be true (as the DLL has to be in Geany's libdir
+ * but the plugin's data might be anywhere in theory) */
+#ifdef G_OS_WIN32
+#   include <windows.h>
+
+static HINSTANCE geniuspaste_hinstance = NULL;
+static gchar *geniuspaste_datadir = NULL;
+
+BOOLEAN WINAPI DllMain(IN HINSTANCE hDllHandle,
+                       IN DWORD     nReason,
+                       IN LPVOID    Reserved)
+{
+    switch (nReason)
+    {
+        case DLL_PROCESS_ATTACH:
+            DisableThreadLibraryCalls(hDllHandle);
+
+            geniuspaste_hinstance = hDllHandle;
+            break;
+
+        case DLL_PROCESS_DETACH:
+            g_free(geniuspaste_datadir);
+            geniuspaste_datadir = NULL;
+            break;
+    }
+
+    return TRUE;
+}
+
+static const gchar *get_plugin_data_directory(void)
+{
+    if (! geniuspaste_datadir)
+    {
+        gchar *base = g_win32_get_package_installation_directory_of_module(geniuspaste_hinstance);
+
+        geniuspaste_datadir = g_build_filename(base, PLUGINDATADIR, NULL);
+        g_free(base);
+    }
+
+    return geniuspaste_datadir;
+}
+
+#   undef PLUGINDATADIR
+#   define PLUGINDATADIR (get_plugin_data_directory())
+#endif
+
 static void pastebin_free(Pastebin *pastebin)
 {
     g_key_file_free(pastebin->config);


geniuspaste/src/geniuspaste.c
```diff
@@ -94,6 +94,54 @@ PLUGIN_SET_TRANSLATABLE_INFO(LOCALEDIR, GETTEXT_PACKAGE, PLUGIN_NAME,
                              PLUGIN_VERSION, "Enrico Trotta <enrico.trt@gmail.com>")
 
 
+/* Windows compatibility for resolving system plugin data path.
+ * This only works if the plugin's data is installed in the same tree as the
+ * plugin DLL, which might not be true (as the DLL has to be in Geany's libdir
+ * but the plugin's data might be anywhere in theory) */
```
b4n: How can we solve this better?  I have no idea how we could locate the data files if they are not in the same tree as the plugin's DLL.  I don't know if it's a plausible situation on Windows, but it is on Linux, so it's at least worth asking.

codebrainz: Maybe win32 resources or GResource?

b4n: Yeah, GResource would work, as it'd be packed in.

eht16: Why so complicated?
`g_win32_get_package_installation_directory_of_module(NULL);` works perfectly fine and returns "C:\Program Files (x86)\Geany". We use exactly this also in Geany itself.

I think this would even not break if the plugin DLL is installed somewhere else, according to the GTK docs, if I understood correctly, when passing `NULL` the location of the main process is used, which is in our case Geany.
That would require that the plugin's data still has to be installed in Geany's installation directory.
I would say this is a valid requirement.

Btw, PLUGINDATADIR cannot be used here because it contains the absolute path of the prefix used at `./configure`.
But PKGDATADIR can be used even if it is not per plugin defined, e.g.
```-DPLUGINDATADIR=\"/h/geany_install/share/geany-plugins/geniuspaste\" 
-DPKGDATADIR=\"share/geany-plugins\" ```
(I configure Geany on my system with `--prefix=/h/geany_install` where /h is H:\ which is a RAM disk)

See http://pastebin.geany.org/WOdqZ/ for a simpler patch which works for me, pasted from within Geany using GeanyVC and GeniusPaste :)

eht16: Oh, we should remove the `g_message` call from my patch :)

eht16:
Regarding the link crash: this is unrelated to the cross-compilation, at least I got the same when natively compiled on Windows.
https://github.com/geany/geany-plugins/pull/403 has a workaround, though no fix nor do I know the root cause for the crash :(.

geniuspaste/src/geniuspaste.c
eht16: One more note: an alternative fix for the path issue could be to manipulate `PLUGINDATADIR` in `src/Makefile.am`.

b4n: @codebrainz Actually GRssource is probably not the best idea here, as those are more configuration than data files the plugin requires.  GResource look like a good idea for UI definitions, icons or the like, things that are part of the app itself, but sounds odd for configuration.

@eht16 
> Why so complicated?
`g_win32_get_package_installation_directory_of_module(NULL);` works perfectly fine and returns "C:\Program Files (x86)\Geany". We use exactly this also in Geany itself.

Yeah, that does the same as my more complicated version if both are installed at the same location (which is the installer thing).  And it's probably fine, as my version doesn't handle better the case the data files are not next to the plugin DLL either.
It's slightly worse in theory IMO, as my version should work also if the plugin was installed in a 3rd part directory, and Geany made to look in that tree's *lib/* folder -- as it would look for the files in that tree rather than Geany's.

> Btw, PLUGINDATADIR cannot be used here because it contains the absolute path of the prefix used at `./configure`.

Sad.  We probably should make this one also usable, like others.
Though, I successfully used it by doing `./configure --prefix=/ && make install DESTDIR=/somewhere` -- this "fixes" all paths at once as it removes all prefix.

---

Looking at either of those, I wonder if we shound't provide a `libgeany-plugins` GP plugins can link to and which would provide the path lookups (so has access to GP's configured directories at least on *NIX, which code in Geany couldn't).  Looks like many plugins could need something like this.

b4n: Wait.  I think I just had an epiphany:  if we use the code I have here, but on a `libgeany-plugins` library, which would be installed in GP's own tree, it would be correct no matter where the Geany and GP's tree are far apart.

But maybe it's just being too nice, as maybe indeed we rely on both being next to one another.

eht16: So, what do we want to do?
I'm not sure how relevant supporting any possible installation scenario on Windows really is. I pretty much assume most users just install Geany. Maybe there are two or three developers who actually fight with developing on Windows...

Regarding `libgeany-plugins`: sounds interesting and would ease a lot of things probably. Something we should consider for 1.28.

codebrainz: @b4n GResource is fine for default configuration data, IMO. I have plugins where I just hard-code the default config into a C string, and on first use dump it into a file in the users config dir. Using GResource would be effectively the same. Having user-editable data files in a non-user accessible directory is not a good idea, which is what it sounds like this plugin is doing, instead of copying/writing them into a directory the user can access.

b4n: @codebrainz well, I don't want to pollute the user directory with stuff installed by the plugin that would better be off somewhere else.  If the user wants to edit those, he copies them in his own directory, but he shouldn't need to, it shuold Just Work.  But it's totally optional, I don't need them to function, without them there's just no default pastebins.

b4n: @eht16 
> So, what do we want to do?
I'm not sure how relevant supporting any possible installation scenario on Windows really is. I pretty much assume most users just install Geany. Maybe there are two or three developers who actually fight with developing on Windows...

You already know it, but I opened #405 and #404 as an alternative.

PR #404, opened 2016-03-12
Define plugin paths

b4n:
Define PLUGIN and PLUGIN{DATA,DOC,LIB}DIR for all plugins

This helps for locating data files, and to define those appropriately depending on the platform (Windows).


b4n:
Obsoleted by #406.

PR #405, opened 2016-03-12
geniuspaste: Add Windows-specific logic for locating system data files

b4n:
This fixes locating pastebin configuration files shipped with the
plugin under Windows (those installed in the plugin's data directory
rather than in the user's configuration directory).

---
somewhat simpler version of #402.


b4n:
relies on `PLUGINDATADIR` being defined properly, so either `--prefix=/` on Windows, or #404.

b4n:
Cherry-picked in #406.

PR #407, opened 2016-03-12
overview: Fix finding the interface file for the preferences on Windows

b4n:
Based off #406, so better apply that one first.


codebrainz:
Not tested, but it looks OK.

codebrainz:
@b4n if you don't mind, you could push this upstream too (no need for PR). I use [a script there](https://github.com/codebrainz/overview-plugin/blob/master/gp-update.sh) to generate changes to GP.

b4n:
@codebrainz okay, will try to remember doing that.

PR here is superseeded by #406 though, so it can be closed.

PR #408, opened 2016-03-12
markdown: Remove an unused define

b4n:
Untested, but should be fine

### DIFF
--- a/markdown/src/Makefile.am
+++ b/markdown/src/Makefile.am
@@ -18,7 +18,6 @@ noinst_HEADERS = \
 markdown_la_CFLAGS = \
 	$(AM_CFLAGS) \
 	$(MARKDOWN_CFLAGS) \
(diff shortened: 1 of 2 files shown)


codebrainz:
Seems OK if it's not used and is causing grief.

b4n:
> and is causing grief.

Probably not much, but that's one path less to worry about

codebrainz:
OK by me then. I probably thought of a use for it at some point and just didn't need it or stopped needing it.

geanyadmin:
> OK by me then. I probably thought of a use for it at some point and just
> didn't need it or stopped needing it.

If we merge #406 it'd become the same as PLUGINDATADIR, so you could use
that instead in the future.

frlan:
@b4n Can you rebase it after the 1.27 as I woul direct merge it in

PR #409, opened 2016-03-13
scope: Fix dependencies checks on Windows

b4n:
On Windows Scope doesn't need neither VTE nor pty.h, so drop the incorrect checks there.

### DIFF
--- a/build/scope.m4
+++ b/build/scope.m4
@@ -3,17 +3,22 @@ AC_DEFUN([GP_CHECK_SCOPE],
     GP_ARG_DISABLE([Scope], [auto])
     GP_CHECK_PLUGIN_GTK2_ONLY([Scope])
 
-    GP_CHECK_PLUGIN_DEPS([scope], [VTE],
-                         [vte >= 0.17])
-    AC_CHECK_HEADERS([util.h pty.h libutil.h])
-    GP_COMMIT_PLUGIN_STATUS([Scope])
-
     case "$host_os" in
-        cygwin* | mingw* | win32*) PTY_LIBS="" ;;
-        *) PTY_LIBS="-lutil" ;;
+        cygwin* | mingw* | win32*)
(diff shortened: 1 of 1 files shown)


b4n:
@eht16 @kugel- could you try this on Windows and if it does work, merge it?

eht16:
I think I tried it in March but forgot to give feedback: IIRC the changes are fine and Scope builds. Though the plugin is of limited use on Windows because you cannot define the path to the `gdb` executable, Scope expects to find `gdb.exe` in $PATH.
But having Scope built and distributed at all surely would be better than now.

I will test this in the next days again to get sure but before I need to rebuild everything with GTK2 for Scope (no GTK3 support).

codebrainz:
> Though the plugin is of limited use on Windows because you cannot define the path to the gdb executable,

And unless it's been fixed since I last used it, it doesn't work on non-`C:` drives, which is a drag when all your source code lives on a data drive or network share :(

eht16:
This PR works: Scope can be built fine on Windows after applying the changes.

As mentioned, Scope expects gdb.exe to be available in $PATH and probably also what @codebrainz reported (I didn't test). 

Do we want to ship Scope in the Windows installer anyway?
If so, just merge it.

b4n:
I don't care at all (last time I tried to test Scope on Linux it didn't seem to work, but maybe I was doing something wrong, and I didn't search further).
The PR here wasn't made in a hope to get it so I can use it on Windows, but merely to get things "more right".

IMO if this fixes the build, it should be merged, and whether or not it is included in the installer could be another question.  But I don't have any real opinion actually.

codebrainz:
> Do we want to ship Scope in the Windows installer anyway?

Both issues have simple-enough workarounds. IMO, if it's the only working debugging plugin on Windows, it'd probably be worth leaving it in.

PR #410, opened 2016-03-13
GTK+3 support for more plugins

sardemff7:
Geany requires GTK+ 2.24 minimum, so all the used APIs are there directly.

### DIFF
--- a/geanylatex/src/geanylatex.c
+++ b/geanylatex/src/geanylatex.c
@@ -304,10 +304,10 @@ plugin_configure(GtkDialog * dialog)
 	config_widgets.lower_selection_on_smallcaps = gtk_check_button_new_with_label(
 		_("Lower selection when formatting smallcaps (\\textsc{})"));
 
-	config_widgets.glatex_autocompletion_active = gtk_combo_box_new_text();
-	gtk_combo_box_insert_text(GTK_COMBO_BOX(config_widgets.glatex_autocompletion_active), 0,
+	config_widgets.glatex_autocompletion_active = gtk_combo_box_text_new();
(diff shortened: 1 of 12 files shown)


sardemff7:
(This version requirement is there since Geany 1.25, and Geany Plugins requires Geany 1.26.)

geanylatex/src/geanylatex.c
```diff
@@ -2128,8 +2128,8 @@ add_menu_to_menubar(void)
 	{
 		/* Build up menu for menubar */
 		menu_latex = gtk_menu_item_new_with_mnemonic(_("_LaTeX"));
-		gtk_menu_shell_insert(
-			menubar, menu_latex, g_list_length(menubar->children)-1);
+		gtk_menu_shell_append(
+			menubar, menu_latex);
```
b4n: Is this the same?  I think that was the same change that altered GeanyVC behavior, wasn't it?

geanylatex/src/geanylatex.c
```diff
@@ -2278,8 +2278,8 @@ add_menu_to_menubar(void)
 	if (menu_bibtex == NULL)
 	{
 		menu_bibtex = gtk_menu_item_new_with_mnemonic(_("_BibTeX"));
-		gtk_menu_shell_insert(
-			menubar, menu_bibtex, g_list_length(menubar->children)-1);
+		gtk_menu_shell_append(
+			menubar, menu_bibtex);
```
b4n: idem

b4n:
Not tested, but mostly LGTM

geanylatex/src/geanylatex.c
frlan: It makes a difference yes. I think we should add the same logic as on GeanyVC here.

sardemff7:
Changed to the same logic as GeanyVC.

PR #419, opened 2016-03-29
scope: Fix validate_number()

techee:
The brackets are apparently wrong - the check that should happen is

strcmp(text, "2147483648") < 0

but instead the whole expression

(s - text == 10 && strcmp(text, "2147483648"))

is compared if it's less than zero.

Noticed by LLVM:

```
utils.c:654:50: warning: comparison of constant 0 with boolean expression is always false
      [-Wtautological-constant-out-of-range-compare]
                (s - text == 10 && strcmp(text, "2147483648")) < 0) ? text : NULL;
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~
```

### DIFF
--- a/scope/src/utils.c
+++ b/scope/src/utils.c
@@ -651,7 +651,7 @@ static gchar *validate_number(gchar *text)
 	for (s = text; isdigit(*s); s++);
 	*s = '\0';
 	return *text && (s - text < 10 ||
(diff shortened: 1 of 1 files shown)


b4n:
LGBI

techee:
Merged as the original seemed to LBBI quite clearly.

PR #435, opened 2016-06-09
Resign Maintainership

elextr:
Operation of Geany Plugins is untenable, commits are made without maintainer approval. The plugin has now diverged from upstream.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -138,10 +138,10 @@ W: http://plugins.geany.org/geanyprj.html
 S: Odd Fixes
 
 geanypy
-P: Lex Trotman <elextr@gmail.com>
-M: Lex Trotman <elextr@gmail.com>
+P:
+M:
 W: http://plugins.geany.org/geanypy.html
-S: Maintained
+S: Orphaned, diverged from upstream
 
 geanysendmail
 P: Frank Lanitz <frank@frank.uvena.de>


frlan:
Thinking about whether it would be useful to remove it at all from this distribution as having two plugin doesn't make sense and in future there will be always a gap due to the difference on developing.

elextr:
The problem with that is this is the only distribution for the plugin, upstream is a development repository and requires users compiling the plugin.  But a Python user may not have C development experience.

frlan:
G-p requires compiling too. We only offer the service for Windows users (is geanypy included there anyway?). Everything else is done by distribution workers like @hyperair, @dmaphy, ..

elextr:
On 9 June 2016 at 23:23, Frank Lanitz <notifications@github.com> wrote:

> G-p requires compiling too. We only offer the service for Windows users
> (is geanypy included there anyway?). Everything else is done by
> distribution workers like @hyperair <https://github.com/hyperair>, @dmaphy
> <https://github.com/dmaphy>, ..
>

​Correct, the Linux users can get it from the distros, as I said, there is
no distribution of upstream.​

b4n:
@elextr well, the problem is that

* many plugins aren't really maintained, even by their maintainers.  That's the problem of GP: we want some standard out of it, but maintainers don't necessarily answer.  Another approach could be arbitrarily dropping plugins when they have something we'd like to fix instead of proposing a fix.
* some are happy we do some of the work.  And that's mostly fine, although some should also show more concern :)
* some never answer PRs, neither refusing nor accepting them.  Which is the problem here, you probably should have refused it.  It's fairly complex to remember not to PR something here but other things there, for which part are here (e.g. build system should go on GP, but code changes on upstream, which then should be imported back).
So yeah, it might be annoying, I get that.  But maintaining it might mean triaging reports, or even backport changes -- or revert them.

So well… hard to please us all :S

codebrainz:
One neat thing might be to write a script which gets the list of issues off Github, reads MAINTAINERS, and for each maintainer emails them a list of links to the issues labelled with the plugin they maintain, maybe like once a week or month. It's only a technical solution to @b4n's first and second item above, but at least it might annoy people enough to either deal with the issues, or say "hey stop emailing me, I don't maintain this anymore". Just a thought.

codebrainz:
> Correct, the Linux users can get it from the distros, as I said, there is no distribution of upstream.​

FWIW, I used to provide binaries until Github dropped the Downloads feature and since GeanyPy got forked into another project which conveniently got packagers to package it, I stopped bothering.

techee:
> Operation of Geany Plugins is untenable, commits are made without maintainer approval. The plugin has now diverged from upstream.

Alright, I confess I didn't know there was other upstream than GP for this plugin - I assumed upstream was the GP version. And looks even crazier, you Lex were a maintainer of Matthew's plugin in GP? How did that work? You were watching Matthew's repository for changes and applying them to GP? And if someone made a pull request on GP, you took the patch, made a pull request on Matthew's repo?

IMO it would be best if GP was upstream for every (maintained) plugin - both for the development reasons (like applying patches to all plugins at once when some change is made in Geany, verification that something didn't break for all the plugins etc.) and also user perspective (all plugins at one place). 

Of course everybody is free to decide whether he wants his plugin part of GP or not but if he wants to keep a separate repo for development, he should make sure the two are in sync (and it should be one person taking care for both repos).

elextr:
> Alright, I confess I didn't know there was other upstream than GP for this plugin

Wonder why its called "readme" :wink: 

As was explained in other forums, @codebrainz  did not want/have time to make the G-P build changes, so I had a go (though it needed fixing by @b4n and others) to make geanypy available to non-building users.  It would have been rude to nominate codebrainz as maintainer without his permission, so I made myself maintainer.

There have not been enough changes for a hard and fast policy to be defined but basically changes from upstream were moved to G-P one way or another, although sometimes we got ahead of upstream in the expectation that the change was going to be accepted upstream "soon".  But other than the build system no independent changes were added to Geanpy in G-P.

MAINTAINERS
```diff
@@ -138,10 +138,10 @@ W: http://plugins.geany.org/geanyprj.html
 S: Odd Fixes
 
 geanypy
-P: Lex Trotman <elextr@gmail.com>
-M: Lex Trotman <elextr@gmail.com>
```
techee: Let's get technical and review the patch :-).

I actually agree with the "-" part of the patch - having GP version of the plugin maintained by someone else than the upstream author is less than ideal. IMO it would be best having Matthew Brush at the "+" side of the patch if @codebrainz is willing to maintain it. 

Otherwise I think it doesn't make sense to keep an unmaintained copy of a maintained plugin in GP.

MAINTAINERS
```diff
@@ -138,10 +138,10 @@ W: http://plugins.geany.org/geanyprj.html
...
+P:
+M:
 W: http://plugins.geany.org/geanypy.html
-S: Maintained
```
techee: The "diverged from upstream" part can be easily fixed by a simple "git revert" of the patch I merged or by applying the patch upstream - but this requires some kind of actual maintenance of the plugin which is something that should happen if the plugin should stay in GP (my opinion).

MAINTAINERS
elextr: On 10 June 2016 at 21:36, Jiří Techet <notifications@github.com> wrote:

> In MAINTAINERS
> <https://github.com/geany/geany-plugins/pull/435#discussion_r66599694>:
>
> > @@ -138,10 +138,10 @@ W: http://plugins.geany.org/geanyprj.html
> >  S: Odd Fixes
> >
> >  geanypy
> > -P: Lex Trotman <elextr@gmail.com>
> > -M: Lex Trotman <elextr@gmail.com>
>
> Let's get technical and review the patch :-).
>
> I actually agree with the "-" part of the patch - having GP version of the
> plugin maintained by someone else than the upstream author is less than
> ideal. IMO it would be best having Matthew Brush at the "+" side of the
> patch if @codebrainz <https://github.com/codebrainz> is willing to
> maintain it.
>
> Otherwise I think it doesn't make sense to keep an unmaintained copy of a
> maintained plugin in GP.
>
​In that case you better remove quite a lot of other plugins that are also
unmaintained.

There is nothing to review.  If a later patch adds a maintainer, fine.

techee: > ​In that case you better remove quite a lot of other plugins that are also
unmaintained.

But they don't have an upstream version from which they'd diverge.

codebrainz: >  IMO it would be best having Matthew Brush at the "+" side of the patch if @codebrainz is willing to maintain it. 

I like having separate repo/issues/build system, etc, which is why I kept GeanyPy as external to Geany-Plugins. I don't think every plugin people write should be forced to be in Geany-Plugins. I don't mind formatting patches for changes to upstream so they can be applied to the fork in Geany-Plugins, but it's going to get harder and harder as the fork has diverged significantly from my repo (ex. the proxy plugin stuff, the geany_functions commit, etc).

techee: > I like having separate repo/issues/build system, etc, which is why I kept GeanyPy as external to Geany-Plugins. I don't think every plugin people write should be forced to be in Geany-Plugins.

Yes, definitely not forced but let's say "strongly suggested" because having a plugin in GP is nice both for users and developers. I don't think there's a problem in having a separate development repo - I think even with this it could be quite easy to copy-over the sources from the development repo to GP before a release with a simple commit message like "updates from upstream" (something similar to what happens when Scintilla gets updated in Geany).

The question is what to do when someone makes a pull request to the plugin in GP - one option is to merge it in GP and apply it also upstream, the other is to reject the patch in GP and ask for submitting it upstream.

> I don't mind formatting patches for changes to upstream so they can be applied to the fork in Geany-Plugins, but it's going to get harder and harder as the fork has diverged significantly from my repo (ex. the proxy plugin stuff, the geany_functions commit, etc).

That's real pity it became a real fork - any chance to get the implementations merged so there's just one geanypy? Which one contains what - the GP one contains the proxy plugin stuff and yours not or the other way round?

codebrainz: > The question is what to do when someone makes a pull request to the plugin in GP - one option is to merge it in GP and apply it also upstream, the other is to reject the patch in GP and ask for submitting it upstream.

Unless the GP maintainer intends to have a proper fork, IMO the proper way is to only accept Git-formatted patches from upstream, keeping the authorship and commit messages at least, if possible. Shouldn't allow changes downstream which are not first applied upstream. Something like #440 would enforce this and makes it easier to keep downstream tracking (whatever stable) upstream they want.

> That's real pity it became a real fork - any chance to get the implementations merged so there's just one geanypy?

I don't think it would be really hard to bring them inline, but it would be better to do it after something like #440 so it can't happen again, IMO.

> Which one contains what - the GP one contains the proxy plugin stuff and yours not or the other way round?

Yeah, the main repo doesn't have the proxy plugins stuff merged, mostly because a) it got rebased after I started reviewing/testing b) I wanted to retain compatibility so everybody's plugins didn't break and c) after I gave in and gave @kugel- the OK to merge it, I think he lost interest since it was merged in the fork already.

elextr: > Something like #440 would enforce this and makes it easier to keep downstream tracking (whatever stable) upstream they want.

+1

b4n:
> Removing the maintainer is fine for me so I will merge the patch.

Sure, we can't force the maintainer to stay.

But as it's a forked plugin nobody would maintain, we might want to consider removing it altogether instead of getting an "unmaintained version of a maintained plugin" as @techee rightly put it -- which, if it stays that way long enough, is rather counter-productive as many people would use an outdated version of the plugin while they could use the upstream one (which indeed has less visibility than the already established GP, but could be used/packaged just fine -- or so I hope).  Maybe just list "this plugin can be found at [URL]" in a directory named after the plugin and voila, at least until that #440 thing got sorted out.
Just some ¢2.

elextr:
Yes, if upstream is _DISTRIBUTED_ on both Windows and Linux then the G-P plugin should be pointed at upstream.  But until upstream either itself or persuades others to package it, this remains the only packaged source.

kugel-:
I can take over maintainership. I can push to upstream too if needed to keep the sync.

But I won't be working on new features myself.

PR #437, opened 2016-06-09
Geanygendoc: replace deprecated filetype_id

techee:


techee:
Looks like false positive from travis.

b4n:
Seems to require API 226 as it was introduced here in geany/geany@302b40e9778ff49407902f3ef9f272ed202f4547

b4n:
Apart that I re-ran travis that failed without a reason apparently, and it's now happy.

Also, I don't see this as much required, but I don't mind either.

techee:
> Seems to require API 226 as it was introduced here in geany/geany@302b40e

Right. In fact, it should be 227 because API wasn't bumped for this change. Fixed now.

> Also, I don't see this as much required, but I don't mind either.

Up to you really - I just tried to get less warnings during build.

PR #438, opened 2016-06-09
multiterm: Drop deprecated geany_functions and bump API requirement

techee:
@codebrainz Please check - pretty trivial though so probably correct.

### DIFF
--- a/multiterm/src/plugin.vala
+++ b/multiterm/src/plugin.vala
@@ -25,7 +25,6 @@ using MultiTerm;
 
 public Plugin		geany_plugin;
 public Data			geany_data;
-public Functions	geany_functions;
 
 extern const string LOCALEDIR;
 extern const string GETTEXT_PACKAGE;
@@ -36,7 +35,7 @@ private List<Widget> toplevel_widgets = null;
 /* Geany calls this to determine min. required API/ABI version */
 public int plugin_version_check(int abi_version)
 {
-	return Plugin.version_check(abi_version, 185);
+	return Plugin.version_check(abi_version, 224);
 }
 
 /* Geany calls this to get some info about the plugin */


codebrainz:
Seems kind of pointless to bump the required API just to remove a harmless line of code. If you're adamant, I don't really mind though, as long as the plugin still builds and works.

techee:
The geany_functions declaration was required before the introduction of geany shared library, which corresponds to API 224. When geany_functions gets removed, the plugin won't work on older Geany versions anyway so better to announce it using the API. I did this change in the past for other plugins too, see https://github.com/geany/geany-plugins/pull/388 and the linked bug.

codebrainz:
Yeah, I know what it is (I wrote the original commit removing it), I just mean since it's not removed from Geany, why make the plugin artificially require a way newer Geany version, just so one line of harmless code can be removed?

techee:
Because there are way too many deprecation warnings when compiling GP and it's easy to overlook the important ones (happened to me before). Now it's true multiterm generates quite many warnings anyway because of the vala-generated code (any way to suppress that?) so one extra doesn't matter that much.

I'll leave the decision up to you - IMO not many people use new plugins with old Geany version so the change wouldn't hurt anyone.

codebrainz:
Couldn't we just not use a compiler builtin to nag about deprecations in Geany? It's pretty rare that we ever use these, usually just the API documentation has `@deprecated` added. It just seems weird to churn so many plugins, bumping their API dependency, just to avoid warnings we ourselves cause (annoying, I agree).

Anyway, I don't really care that much. As long as the plugin still builds and functions.

frlan:
So merging yes/no? Let's start the poll ;)

PR #448, opened 2016-06-15
projectorganizer: Fix multiple tabs in Project Properties

### LINKED ISSUE #444: Project Organizer: Keeps adding more tabs to Project prefs dialog
I thought I had opened an issue about this already but I can't find it. The ProjectOrganizer plugin keeps adding tabs to the Project properties dialog each time it's shown and not removing them when it's closed.

geany_po_proj_dialog

techee:
There is a problem when there are multiple tabs added by plugins: if
on_project_dialog_open() is called and Project Organizer stores
the index of the tab and then another plugin adds a tab before the
Project Organizer's tab, then the stored index becomes invalid which
can lead to not correctly removing the tab from the dialog.

Instead store a pointer of the added tab widget and remove the widget
when the properties dialog is closed.

Fixes #444

### DIFF
--- a/projectorganizer/src/prjorg-main.c
+++ b/projectorganizer/src/prjorg-main.c
@@ -42,7 +42,7 @@ PLUGIN_SET_TRANSLATABLE_INFO(
 	VERSION,
 	"Jiri Techet <techet@gmail.com>")
 
-static gint page_index = -1;
+static GtkWidget *properties_tab = NULL;
 
 
 static void on_doc_open(G_GNUC_UNUSED GObject * obj, G_GNUC_UNUSED GeanyDocument * doc,
@@ -96,8 +96,8 @@ static void on_build_start(GObject *obj, gpointer user_data)
 static void on_project_dialog_open(G_GNUC_UNUSED GObject * obj, GtkWidget * notebook,
 		G_GNUC_UNUSED gpointer user_data)
 {
-	if (prj_org && page_index == -1)
-		page_index = prjorg_project_add_properties_tab(notebook);
+	if (prj_org && !properties_tab)
+		properties_tab = prjorg_project_add_properties_tab(notebook);
 }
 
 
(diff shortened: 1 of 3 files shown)


projectorganizer/src/prjorg-main.c
```diff
@@ -115,10 +115,10 @@ static void on_project_dialog_confirmed(G_GNUC_UNUSED GObject * obj, GtkWidget *
 static void on_project_dialog_close(G_GNUC_UNUSED GObject * obj, GtkWidget * notebook,
 		G_GNUC_UNUSED gpointer user_data)
 {
-	if (page_index != -1)
+	if (properties_tab)
 	{
-		gtk_notebook_remove_page(GTK_NOTEBOOK(notebook), page_index);
-		page_index = -1;
+		gtk_container_remove(GTK_CONTAINER(notebook), properties_tab);
```
b4n: you should probably simply destroy the widget, `gtk_widget_destroy(properties_tab)` (see [`gtk_container_remove()` docs](https://developer.gnome.org/gtk3/stable/GtkContainer.html#gtk-container-remove))

codebrainz:
Would it be any use to add a function to the API to facilitate adding a plugin-specific tab to the Project preferences dialog? Or as @b4n mentioned on IRC, we could just re-create and delete the dialog each time.

techee:
> Would it be any use to add a function to the API to facilitate adding a plugin-specific tab to the Project preferences dialog? Or as @b4n mentioned on IRC, we could just re-create and delete the dialog each time.

Less API is better IMO so I'd either vote for recreating the dialog as @b4n suggested or keeping things as they are.

projectorganizer/src/prjorg-main.c
techee: One is too lazy to read the function description when gtk_container_remove() seems so self-descriptive :-). Will be fixed.

PR #449, opened 2016-06-17
addons/tasks: Small improvments

sardemff7:
- Do not extract tokens from non-comments
- Strip the comment closing sequence from the text  
    Allows to use `/* TODO: some stuff */` without the extra `*/` being in the tasks list


eht16:
LGTM and tested.

PR #450, opened 2016-06-17
Refactor panel option setting code into a seperate function

eht16:
This fixes a 'false positive' from cppcheck:
dconfig.c:351: error: va_list 'ap' was opened but not closed by va_end().

Probably cppcheck got confused by the nested while, switch and
break statements.

See also the output of last night's Debian Unstable build using cppcheck 1.74:
http://pastebin.geany.org/isOyz/

### DIFF
--- a/debugger/src/dconfig.c
+++ b/debugger/src/dconfig.c
@@ -333,6 +333,54 @@ void config_set_debug_changed(void)
 /*
  *	set one or several panel config values
  */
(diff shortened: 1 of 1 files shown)


eht16:
Don't merge this.
Scope is affected in a similar way and this is a bug in cppcheck which got already fixed:
http://trac.cppcheck.net/ticket/7533

So we just need to wait for a new cppcheck version.

PR #453, opened 2016-06-22
Fix three plugin names in their READMEs

philippwiesemann:
Thanks!


b4n:
LGBI, good catch!

codebrainz:
:+1: 

Copy/paste strikes again!

@b4n I guess it's a good indication that WebHelper is the best plugin to copy when adding a new plugin to GP :)

PR #463, opened 2016-07-14
Git-Changebar: Mention the tooltip feature in the README

vfaronov:
The feature is non-obvious. It would probably take me quite a while to discover it if I hadn’t decided to go browse the code.


b4n:
LGBI, good point

PR #464, opened 2016-07-14
Commander: Add period to SEPARATORS

vfaronov:
This makes it easier to switch between files. For example, if I have `xyzzy.c` and `xyzzy.h` open, I can switch to `xyzzy.h` by typing “x.h”.


b4n:
Seems good, but I'm trying it out for a bit to see if it doesn't have unexpected side effects on search result on daily usage.

PR #467, opened 2016-07-21
Gtk3 compatibility for geanydoc plugin

sagarchalise:
I have tested this in debian testing and ubuntu 16.04 and works for both gtk2 and gtk3.


geanydoc/src/geanydoc.c
```diff
@@ -284,15 +284,14 @@ on_comboboxType_changed(GtkComboBox * combobox, G_GNUC_UNUSED gpointer user_data
 
 #define GLADE_HOOKUP_OBJECT(component,widget,name) \
   g_object_set_data_full (G_OBJECT (component), name, \
-    gtk_widget_ref (widget), (GDestroyNotify) gtk_widget_unref)
+    g_object_ref (widget), (GDestroyNotify) gtk_widget_destroy)
```
b4n: `gtk_widget_destroy` isn't the same as `gtk_widget_unref`, `g_object_unref` would be a lot closer.  It may or may not change anything, depending on how it's used, but it probably would be "safer" to use `g_object_unref`

geanydoc/src/geanydoc.c
```diff
@@ -336,7 +332,9 @@ create_Configure(void)
 	gtk_window_set_title(GTK_WINDOW(Configure), _("Doc"));
 	gtk_window_set_type_hint(GTK_WINDOW(Configure), GDK_WINDOW_TYPE_HINT_DIALOG);
 
-	dialog_vbox1 = GTK_DIALOG(Configure)->vbox;
+	dialog_vbox1 = gtk_vbox_new(FALSE, 6);
```
b4n: doesn't this `6` change UI layout?

b4n: also, you could assume just fine that `gtk_dialog_get_content_area()` returns a VBox if you wanted

b4n:
LGBI apart my small comments.  But this might require a little checking of the actual widget packing behavior, because the GTK3 defaults are different.

kugel-:
The commits should be cleaned up (possible squash everything into one).

PR #472, opened 2016-08-06
Rename GeanyLipsum to just Lipsum

frlan:
There is no need for Geany prefix on plugins.


lipsum/src/lipsum.c
```diff
@@ -144,12 +148,69 @@ plugin_init(G_GNUC_UNUSED GeanyData *data)
...
 	GeanyKeyGroup *key_group;
 
+
 	config_file = g_strconcat(geany->app->configdir,
 		G_DIR_SEPARATOR_S, "plugins", G_DIR_SEPARATOR_S,
 		"geanylipsum", G_DIR_SEPARATOR_S, "lipsum.conf", NULL);
 
+	#ifndef G_OS_WIN32
+	/* We try only to move if we are on not Windows platform */
+	config_dir_old = g_strconcat(geany->app->configdir, G_DIR_SEPARATOR_S,
```
codebrainz: could use [`g_build_filename()`](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-build-filename) instead for these.

lipsum/src/Makefile.am
```diff
@@ -0,0 +1,10 @@
+include $(top_srcdir)/build/vars.build.mk
+plugin = ylipsum
```
codebrainz: typo

lipsum/README
```diff
@@ -1,12 +1,12 @@
-GeanyLipsum
+Lipsum
 -----------
```
b4n: could change the underline length too (not necessary though)

lipsum/src/lipsum.c
```diff
@@ -36,9 +40,9 @@ PLUGIN_VERSION_CHECK(224)
 PLUGIN_SET_TRANSLATABLE_INFO(
 	LOCALEDIR,
 	GETTEXT_PACKAGE,
-	_("GeanyLipsum"),
+	_("Lipsum"),
 	_("Creating dummy text with Geany"),
-	"0.4.4",
+	VERSION,
```
b4n: so you follow GP versioning now?

lipsum/src/lipsum.c
b4n: yep, would probably be neater

lipsum/src/lipsum.c
```diff
@@ -144,12 +148,69 @@ plugin_init(G_GNUC_UNUSED GeanyData *data)
...
+	if (g_file_test(config_file_old, G_FILE_TEST_EXISTS))
+	{
+		if (dialogs_show_question(
+			_("Renamed plugin detected!\n"
+			  "\n"
+			  "GeanyLipsum has been renamed to Lipsum -- you surely have "
+			  "already recognised it. \n"
+			  "Geany is able to migrate your old plugin configuration by "
+			  "moving the old configuration file to new location.\n"
+			  "Attention: this will not include your keybindings.\n"
```
b4n: should probably read "Warning:", shouldn't it?

lipsum/src/lipsum.c
```diff
@@ -144,12 +148,69 @@ plugin_init(G_GNUC_UNUSED GeanyData *data)
...
+				 * When saving new configuration it will go to
+				 * new folder so migration should
+				 * be implicit. */
+				g_free(config_file);
+				config_file = g_strdup(config_file_old);
+				dialogs_show_msgbox(
+					GTK_MESSAGE_WARNING,
+					_("Your old configuration directory \"%s\" could "
+					  "not be moved to \"%s\" (%s). "
+					  "Please move manually the directory to the new location."),
```
b4n: *please manually move […]* or *please move the directory to the new location manually* I think.   @elextr ?

b4n:
Otherwise LGBI

lipsum/src/lipsum.c
```diff
@@ -144,12 +148,69 @@ plugin_init(G_GNUC_UNUSED GeanyData *data)
...
+	config_file_old = g_strconcat(config_dir_old,  G_DIR_SEPARATOR_S,
+		"lipsum.conf", NULL);
+	config_dir = g_strconcat(geany->app->configdir, G_DIR_SEPARATOR_S,
+		"plugins", G_DIR_SEPARATOR_S, "lipsum", NULL);
+	if (g_file_test(config_file_old, G_FILE_TEST_EXISTS))
+	{
+		if (dialogs_show_question(
+			_("Renamed plugin detected!\n"
+			  "\n"
+			  "GeanyLipsum has been renamed to Lipsum -- you surely have "
```
elextr: as you will have already recognised

lipsum/src/lipsum.c
```diff
@@ -144,12 +148,69 @@ plugin_init(G_GNUC_UNUSED GeanyData *data)
...
+			  "Geany is able to migrate your old plugin configuration by "
+			  "moving the old configuration file to new location.\n"
+			  "Attention: this will not include your keybindings.\n"
+			  "Move now?")))
+		{
+			if (g_rename(config_dir_old, config_dir) == 0)
+			{
+				dialogs_show_msgbox(GTK_MESSAGE_INFO,
+					_("Your configuration directory has been "
+					  "successfully moved from \"%s\" to \"%s\"."),
```
elextr: For translatable strings the %s must be order independent in case translation changes the order, use `n$` to specify the parameter: `Your configuration directory has been successfully moved from \"%1$s\" to \"%2$s\".`

b4n: That's not true, you don't have to use positional format to allow translators to do it.  We'd have a lot more to handle if it was the case.

No, the *translators* might have to do that if they want to swap the arguments, but the developer doesn't have to worry about it so long as (s)he uses proper placeholders and not string building.

elextr: Do translators know about `%n$`?

b4n: They are supposed to and Gettext documents it.

lipsum/src/lipsum.c
frlan: Yes. Doesn't make sense to have own numbering schema here

build/lipsum.m4
```diff
@@ -0,0 +1,9 @@
+AC_DEFUN([GP_CHECK_GEANYLIPSUM],
```
b4n: you could want to rename that (and in the configure.ac too).

lipsum/src/lipsum.c
```diff
@@ -144,12 +148,69 @@ plugin_init(G_GNUC_UNUSED GeanyData *data)
...
+		"lipsum.conf", NULL);
+	config_dir = g_build_filename(geany->app->configdir,
+		"plugins", "lipsum", NULL);
+	if (g_file_test(config_file_old, G_FILE_TEST_EXISTS))
+	{
+		if (dialogs_show_question(
+			_("Renamed plugin detected!\n"
+			  "\n"
+			  "GeanyLipsum has been renamed to Lipsum -- as you will "
+			  "already recognised it. \n"
```
b4n: *as you will **have** already recognised.*  Add *have* and remove *it*.
https://github.com/geany/geany-plugins/pull/472#discussion_r73810656

(though I'd rather say *"as you will already have realized"*, but whatever @elextr says is likely correct)

b4n:
Otherwise LGBI

lipsum/src/lipsum.c
codebrainz: "As you may have already noticed, GeanyLipsum has been renamed to just Lipsum."

codebrainz: Could add another sentence to expand a bit like "As a result, the configuration old plugin configuration file will no longer be read".

b4n: > "As you may have already noticed, GeanyLipsum has been renamed to just Lipsum."

even better

Not sure the additional sentence is needed as suggesting to move the configuration seems clear enough to me, but why not.

PR #478, opened 2016-08-17
treebrowser: Free `renderers` in `on_treeview_renamed()`.

konsolebox:
Because it's not freed.

Also, why not just use the value passed to the function instead?


b4n:
Wowow, this plugins uses GTK internal API!? `_gtk_cell_layout_get_cells()` is not supposed to be used outside GTK itself, it's pure luck it didn't break sooner.

And good question why not use the arg (if it's the correct one), or use a custom variable.

PR #481, opened 2016-08-28
Addons: Implement deselect with single click

### LINKED ISSUE #445: [addons] Mark all only when word selected
In most code editors when you unselect the word, all the highlighted marks are also cleared. I wish the mark all functionality in this plugin would also behave like this. Currently you have to either double-click on a blank line or manually clear marks.

eht16:
For this to work, we need to connect to the "button-press-event" for
each document and then differentiate between single and double
clicks, using a timeout handler which is triggered first after the GTK
double-click-time has passed, i.e. GTK won't recognise the event as part
of a double click.

I didn't find a easier solution to differentiate between single and double click.
Better ideas are welcome.

Closes #445.

### DIFF
--- a/addons/src/addons.c
+++ b/addons/src/addons.c
@@ -77,6 +77,7 @@ typedef struct
 	gboolean enable_systray;
 	gboolean enable_bookmarklist;
 	gboolean enable_markword;
+	gboolean enable_markword_single_click_deselect;
 	gboolean enable_xmltagging;
 	gboolean enable_enclose_words;
 	gboolean enable_enclose_words_auto;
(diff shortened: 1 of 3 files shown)


codebrainz:
Not sure if it's useful, but [Scintilla has a double-click notification](http://www.scintilla.org/ScintillaDoc.html#SCN_DOUBLECLICK).

eht16:
I used `SCN_DOUBLECLICK` before in the code, see https://github.com/geany/geany-plugins/pull/481/commits/d18495a9e5481000670ddd9b0b1a952c67857b34?diff=split#diff-7e319383ec157ada7c430d2e26918d02L119 but there is no single click notification. I couldn't find any other Scintilla notification which is close to a single click event.

I removed the `SCN_DOUBLECLICK` only because we already get the same event notification in the GTK "button-press-event" handler and so do the logic at one place.

eht16:
I actually found a better way: see https://github.com/geany/geany-plugins/pull/481/commits/a0ef5fdb6103589fd9fe22721683e9e09c40640d and its message.

PR #486, opened 2016-10-06
geanyprj/workingdir

donjan:
This implements working directories in the geanyprj plugin.
The motivation is to be able to 'make' via build menu shortcut within Geany when the project is set up with CMake and uses some far away build directory.
It syncs with Geany's own build menu working directories, while giving precedence to the main project system.

Comments/improvements welcome.

### DIFF
--- a/geanyprj/src/geanyprj.c
+++ b/geanyprj/src/geanyprj.c
@@ -74,6 +74,7 @@ static void reload_project(void)
 	{
 		if (g_current_project)
 			xproject_close(TRUE);
+		set_non_ft_build_wdir((gchar*)"");
 		return;
 	}
 
@@ -86,6 +87,7 @@ static void reload_project(void)
 		xproject_close(TRUE);
 		xproject_open(proj);
 	}
+	set_non_ft_build_wdir(g_current_project->working_dir);
 	if (proj)
 		g_free(proj);
 }
--- a/geanyprj/src/menu.c
+++ b/geanyprj/src/menu.c
@@ -177,6 +177,18 @@ static PropertyDialogElements *build_properties_dialog(gboolean properties)
 	gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(e->regenerate), TRUE);
 	ui_table_add_row(GTK_TABLE(table), 3, label, e->regenerate, NULL);
 
+	/* add row for working directory */
+	label = gtk_label_new(_("Working Directory:"));
+	gtk_misc_set_alignment(GTK_MISC(label), 1, 0);
+
+	e->working_dir = gtk_entry_new();
+	gtk_widget_set_tooltip_text(e->working_dir,
+			     _("Working directory for run and build commands."));
+	bbox = ui_path_box_new(_("Choose Project working dirctory."),
+			       GTK_FILE_CHOOSER_ACTION_SELECT_FOLDER, GTK_ENTRY(e->working_dir));
+	gtk_entry_set_text(GTK_ENTRY(e->working_dir), dir);
+
+	ui_table_add_row(GTK_TABLE(table), 4, label, bbox, NULL);
 
 	label = gtk_label_new(_("Type:"));
 	gtk_misc_set_alignment(GTK_MISC(label), 1, 0);
@@ -186,7 +198,7 @@ static PropertyDialogElements *build_properties_dialog(gboolean properties)
 		gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(e->type), project_type_string[i]);
 	gtk_combo_box_set_active(GTK_COMBO_BOX(e->type), 0);
 
-	ui_table_add_row(GTK_TABLE(table), 4, label, e->type, NULL);
+	ui_table_add_row(GTK_TABLE(table), 5, label, e->type, NULL);
 
 	gtk_container_add(GTK_CONTAINER(vbox), table);
 	g_free(dir);
--- a/geanyprj/src/menu.c
+++ b/geanyprj/src/menu.c
@@ -50,6 +50,7 @@ typedef struct _PropertyDialogElements
 	GtkWidget *description;
 	GtkWidget *file_name;
 	GtkWidget *base_path;
+	GtkWidget *working_dir;
 	GtkWidget *make_in_base_path;
 	GtkWidget *run_cmd;
 	GtkWidget *regenerate;
@@ -226,6 +227,7 @@ void on_new_project(G_GNUC_UNUSED GtkMenuItem *menuitem, G_GNUC_UNUSED gpointer 
 		geany_project_set_name(prj, gtk_entry_get_text(GTK_ENTRY(e->name)));
 		geany_project_set_description(prj, "");
 		geany_project_set_run_cmd(prj, "");
+		geany_project_set_working_dir(prj, gtk_entry_get_text(GTK_ENTRY(e->working_dir)));
 		geany_project_set_type_int(prj, gtk_combo_box_get_active(GTK_COMBO_BOX(e->type)));
 		geany_project_set_regenerate(prj,
 					     gtk_toggle_button_get_active(GTK_TOGGLE_BUTTON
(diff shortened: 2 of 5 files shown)


geanyprj/src/geanyprj.c
```diff
@@ -74,6 +74,7 @@ static void reload_project(void)
 	{
 		if (g_current_project)
 			xproject_close(TRUE);
+		set_non_ft_build_wdir((gchar*)"");
```
b4n: should take a `const gchar *` as it doesn't need modifying the string anyway, and would remove the need for this odd cast

geanyprj/src/menu.c
```diff
@@ -269,6 +284,7 @@ void on_preferences(G_GNUC_UNUSED GtkMenuItem *menuitem, G_GNUC_UNUSED gpointer
 		geany_project_set_name(g_current_project, gtk_entry_get_text(GTK_ENTRY(e->name)));
 		geany_project_set_description(g_current_project, "");
 		geany_project_set_run_cmd(g_current_project, "");
+		geany_project_set_working_dir(g_current_project,  gtk_entry_get_text(GTK_ENTRY(e->working_dir)));
```
b4n: double space after the comma

geanyprj/src/utils.c
```diff
@@ -275,3 +275,13 @@ GSList *get_file_list(const gchar *path, guint * length, gboolean(*func)(const g
...
 
+/* update Geany's non-filetype working directories in the build menu to wdir */
+void set_non_ft_build_wdir(gchar* wdir) {
+	gint i;
+	for(i = 0; i < build_get_group_count(GEANY_GBG_NON_FT); ++i)
+		build_set_menu_item(GEANY_BCS_PREF, /* medium priority */
+							GEANY_GBG_NON_FT, /* non-filetype build entries */
+							i,
+							GEANY_BC_WORKING_DIR, /* working dir field */
+							wdir);
```
b4n: braces around the `for` would be nice for readability IMO as the body spans multiple lines (even though it's a single statement so doesn't change meaning)

donjan:
@b4n @frlan hope you guys (or someone additional) can help me with these.

1. What would be the best (from what we see in `geanyplugin.h`) way of reloading the build menu from the config files? I'd like to add this explicitly because right now switching between non-project tabs with geanyprj loaded doesn't update the working dir from e.g. filedefs. [0]

2. I can't call `build_menu_update(doc);` from a plugin because the linker doesn't see its implementation. Is this supposed to be in `geanyplugin.h` <- `build.h` and we should also build `build.o` from Geany core and link it (help welcome) - or is it not supposed to be used from plugins?

---

[0] Currently I just unset the path instead of keeping it on the last geanyprj value. Setting it to the lowest priority (GEANY_BCS_DEF) doesn't help because the last value still lingers in higher priority. Being able to have two files from separate projects open in tabs next to each other is the killer feature of geanyprj, and works nicely with working dirs now, but probably others would like it to reload filedef paths when tabbing to a non-project file.

b4n:
@elextr build API question for ya ;)

@donjan that function isn't part of the API (yet?), so it's expected you can't use it.

elextr:
@donjan you shouldn't need to call `build_update_menu()`.

`build_set_menu_item()` already calls it, and `build_remove_menu_item()` documents it does, but doesn't actually do so.  

That should be fixed rather than adding `build_update_menu()` to the API.

donjan:
@elextr I'm aware that it shouldn't be needed, it was just one attempt on the way of achieving point 1. So, understood: no usage due to not being in the documented API :)
Probably fixing `build_remove_menu_item()` should be handled in a separate bug report.

Side note: the changes presented here work nicely when switching between geanyprj projects and also when loading a Geany core project file.
My issue is that when tabbing from a geanyprj file to a non-project file, the build menu needs to be repopulated as given by the file type and `geany.conf`.

I can't play with the current `GeanyBuildSource` values because chosing below `GEANY_BCS_HOME_FT` is not enough to make geanyprj settings show up, while using `GEANY_BCS_HOME_FT` or `GEANY_BCS_PREF` lets the project paths linger (not get updated) at non-project files due to same priority in the respective cases.

One solution would be a plugin API that lets me reload stored settings from the config files (which I don't see currently).

Another one would be to introduce something like `GEANY_BCS_PLUGIN` either higher than or just below the `GEANY_BCS_PROJ` entries, so plugins get their own place for build menu changes.

Thoughts?

elextr:
@donjan remember I don't know anything about your plugin and what it does, so you need to explain what you are trying to do from first principles.  I'm afraid your post above is just confusing, sorry.

geanyprj/src/project.c
```diff
@@ -463,5 +478,6 @@ void geany_project_save(struct GeanyPrj *prj)
 		g_hash_table_foreach(prj->tags, geany_project_save_files, &data);
 	}
 	save_config(config, prj->path);
+	set_non_ft_build_wdir(g_current_project->working_dir);
```
frlan: When creating a new project based upon an open file, this is causing a segfault during my tests

```
Thread 1 "geany" received signal SIGSEGV, Segmentation fault.
0x00007fffe579e791 in geany_project_save (prj=prj@entry=0x555556805350) at project.c:481
481		set_non_ft_build_wdir(g_current_project->working_dir);
(gdb) bt
#0  0x00007fffe579e791 in geany_project_save (prj=prj@entry=0x555556805350) at project.c:481
#1  0x00007fffe579d4dc in on_new_project (menuitem=<optimized out>, user_data=<optimized out>) at menu.c:250
#2  0x00007ffff5995f75 in g_closure_invoke () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#3  0x00007ffff59a7f82 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#4  0x00007ffff59b0bcc in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#5  0x00007ffff59b0faf in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#6  0x00007ffff73a28de in gtk_widget_activate () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#7  0x00007ffff7270826 in gtk_menu_shell_activate_item () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#8  0x00007ffff7270b5b in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#9  0x00007ffff7253b11 in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#10 0x00007ffff59961a4 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#11 0x00007ffff59b0391 in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#12 0x00007ffff59b0faf in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#13 0x00007ffff73a03dc in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#14 0x00007ffff7250ace in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#15 0x00007ffff7252b5e in gtk_main_do_event () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#16 0x00007ffff6d63bf5 in ?? () from /usr/lib/x86_64-linux-gnu/libgdk-3.so.0
#17 0x00007ffff6d94be2 in ?? () from /usr/lib/x86_64-linux-gnu/libgdk-3.so.0
#18 0x00007ffff52b67f7 in g_main_context_dispatch () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#19 0x00007ffff52b6a60 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#20 0x00007ffff52b6d82 in g_main_loop_run () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#21 0x00007ffff7251cc5 in gtk_main () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#22 0x00007ffff79ad097 in main_lib (argc=<optimized out>, argv=<optimized out>) at libmain.c:1233
#23 0x00007ffff4cd12b1 in __libc_start_main () from /lib/x86_64-linux-gnu/libc.so.6
#24 0x000055555555478a in _start ()

```

geanyprj/src/project.c
```diff
@@ -311,6 +315,14 @@ void geany_project_set_run_cmd(struct GeanyPrj *prj, const gchar *run_cmd)
 }
 
 
+void geany_project_set_working_dir(struct GeanyPrj *prj, const gchar *working_dir)
+{
+	if (prj->working_dir)
+		g_free(prj->working_dir);
```
Thannoy: (minor) You can `g_free` unconditionally since freeing NULL is a valid no-op (cd manual [here](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-free)).

PR #489, opened 2016-10-15
add plugin geanypy-preview

xiota:
Plugin to display preview of documents written in various markup languages.
- Supports: Markdown, reStructuredText, Txt2tags, LaTeX, Docbook, Asciidoc, Textile, org-mode, [Fountain](http://fountain.io), and RFC-822-like messages
- May be used as a rudimentary web browser
- Can toggle visibility of menubar
- Requires geanypy, pandoc, and asciidoc/asciidoctor


elextr:
Suggest you use asciidoctor instead of asciidoc, its a lot faster so the preview will update faster.

xiota:
Thank you for the feedback.  Will look into using `asciidoctor`.

codebrainz:
We still need to get setup to have GeanyPy plugins distributed as first class plugins, especially now that Geany treats them as such.

Also, personally I don't like when there are multiple plugins that do the same thing. I haven't tested this plugin but if works well enough and we get it integrated into the build system, maybe we should remove the existing (performant but feature-limited) Markdown plugin in favour of this one.

PR #490, opened 2016-10-18
new maintainer for plugin geanyprj

donjan:
As discussed with b4n on IRC, we need a maintainer for the geanyprj plugin. Yurand hasn't responded in 9 days, so this is my bid to take over maintainership "for odd fixes".

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -127,8 +127,8 @@ W: http://plugins.geany.org/geanypg.html
 S: Odd Fixes
 
 geanyprj
-P: Yura Siamashka <yurand2@gmail.com>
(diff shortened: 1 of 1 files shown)


frlan:
@b4n @Yurand  I think its fine but just want to hear you "ok" or "I don't care" ;)

b4n:
@frlan I don't really have an opinion, apart that an interested maintainer is better than none :)

frlan:
Indeed. Glad someone is willing to take it over. As it is just kind of a formal act and no offensive against @Yurand I will merge it. In worst case we can start an edit war ;)

PR #501, opened 2016-11-11
Markdown: Fix relative paths

### LINKED ISSUE #500: [WhishList] geany-plugin-markdown
Hi,

I use geany-plugin-markdown to take notes in university. It makes really nice output, but i miss a little feature about image displaying.

I'm using fast relatives names. Using pandoc and latex i make pdf with no problems. But in preview it doesn't work :

Image preview working

I found out how to get a rendering using absolute path :

Image preview not working

It's very time costing to type absolute path every time.

Is there an easy way to get relative paths working in WebKit output ?

Thanks for your usefull software :) !

codebrainz:
Stop using a phony base URI and create one from the current filename, with various fallbacks in case the file isn't saved yet or URI encoding fails.

Closes #500

### DIFF
--- a/markdown/src/viewer.c
+++ b/markdown/src/viewer.c
@@ -23,6 +23,7 @@
 #include <string.h>
 #include <gtk/gtk.h>
 #include <webkit/webkitwebview.h>
+#include <geanyplugin.h>
 #ifndef FULL_PRICE
 # include <mkdio.h>
 #else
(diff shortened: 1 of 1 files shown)


markdown/src/viewer.c
```diff
@@ -356,7 +357,34 @@ markdown_viewer_update_view(MarkdownViewer *self)
...
+      gchar *base_dir = g_path_get_dirname(doc->real_path);
+      base_path = g_build_filename(base_dir, "index.html", NULL);
+      g_free(base_dir);
+    }
+    /* Otherwise assume use a file `index.html` in the current working directory. */
+    else {
+      gchar *cwd = g_get_current_dir();
+      base_path = g_build_filename(cwd, "index.html", NULL);
+      g_free(cwd);
+      g_debug("current document is not saved, using '%s' as base filename", base_path);
```
b4n: might be too spammy if this function is called often (e.g when typing?)

markdown/src/viewer.c
```diff
@@ -356,7 +357,34 @@ markdown_viewer_update_view(MarkdownViewer *self)
...
+    }
+
+    base_uri = g_filename_to_uri(base_path, NULL, &error);
+    if (base_uri == NULL) {
+      g_warning("failed to encode path '%s' as URI: %s", base_path, error->message);
+      g_error_free(error);
+      base_uri = g_strdup("file://./index.html");
+      g_debug("using phony base URI '%s', broken relative paths are likely", base_uri);
```
b4n: Same here if it's expected to happen in real situation (?)

markdown/src/viewer.c
codebrainz: Yeah, true, if each re-render.

markdown/src/viewer.c
codebrainz: I kind of want this message at least once, but yeah, each update is a bit much.

b4n: If you never expect this to actually happen and this test is mostly defensive, maybe don't worry too much about it.

codebrainz:
Force pushed removal of expected spammy debug message.

codebrainz:
Added 1.29 milestone in case this bug fix could be included, don't mind if it's not though.

PR #502, opened 2016-11-11
Markdown: Allow exporting Markdown as HTML

### LINKED ISSUE #497: geany-plugin-markdown: can save preview into an html?
The preview is perfect with multimarkdown and pandoc' markdown file. (some syntax of table different in multimarkdown and pandoc).

If we can use the WebKit preview' menu to save the result into an html file? Even convert into an pdf file with wkhtmltopdf. It seems quite easy to do it.

codebrainz:
Add a Tools menu item "Export Markdown as HTML" to allow saving of the rendered HTML to disk instead of just the live preview.

Closes #497

### DIFF
--- a/markdown/src/plugin.c
+++ b/markdown/src/plugin.c
@@ -251,12 +251,29 @@ on_view_pos_notify(GObject *obj, GParamSpec *pspec, MarkdownViewer *viewer)
   update_markdown_viewer(viewer);
 }
 
+static gchar *replace_extension(const gchar *utf8_fn, const gchar *new_ext)
+{
+  gchar *fn_noext, *new_fn, *dot;
+  fn_noext = g_filename_from_utf8(utf8_fn, -1, NULL, NULL, NULL);
+  dot = strrchr(fn_noext, '.');
+  if (dot != NULL) {
+    *dot = '\0';
+  }
+  new_fn = g_strconcat(fn_noext, new_ext, NULL);
+  g_free(fn_noext);
+  return new_fn;
+}
+
 static void on_export_as_html_activate(GtkMenuItem *item, MarkdownViewer *viewer)
 {
   GtkWidget *dialog;
   GtkFileFilter *filter;
+  gchar *fn;
+  GeanyDocument *doc;
+  gboolean saved = FALSE;
 
-  g_return_if_fail(DOC_VALID(document_get_current()));
+  doc = document_get_current();
+  g_return_if_fail(DOC_VALID(doc));
 
   dialog = gtk_file_chooser_dialog_new(_("Save HTML File As"),
     GTK_WINDOW(geany_data->main_widgets->window), GTK_FILE_CHOOSER_ACTION_SAVE,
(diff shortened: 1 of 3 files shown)


codebrainz:
I optimistically added 1.29 milestone though none of the new strings are translated yet, feel free to ignore :)

markdown/src/plugin.c
```diff
@@ -240,3 +250,47 @@ on_view_pos_notify(GObject *obj, GParamSpec *pspec, MarkdownViewer *viewer)
...
+
+  g_return_if_fail(DOC_VALID(document_get_current()));
+
+  dialog = gtk_file_chooser_dialog_new(_("Save HTML File As"),
+    GTK_WINDOW(geany_data->main_widgets->window), GTK_FILE_CHOOSER_ACTION_SAVE,
+    GTK_STOCK_CANCEL, GTK_RESPONSE_CANCEL,
+    GTK_STOCK_SAVE, GTK_RESPONSE_ACCEPT,
+    NULL);
+  gtk_file_chooser_set_do_overwrite_confirmation(GTK_FILE_CHOOSER(dialog), TRUE);
+  gtk_file_chooser_set_select_multiple(GTK_FILE_CHOOSER(dialog), FALSE);
```
b4n: I think multi-selection is disabled by default, and probably irrelevant for save too :)

markdown/src/plugin.c
```diff
@@ -240,3 +250,47 @@ on_view_pos_notify(GObject *obj, GParamSpec *pspec, MarkdownViewer *viewer)
...
+  gtk_file_chooser_set_filename(GTK_FILE_CHOOSER(dialog), "index.html");
```
b4n: Maybe rather use `basename(DOC_FILENAME(doc)) + ".html"`?

markdown/src/plugin.c
```diff
@@ -240,3 +250,47 @@ on_view_pos_notify(GObject *obj, GParamSpec *pspec, MarkdownViewer *viewer)
...
+
+  filter = gtk_file_filter_new();
+  gtk_file_filter_set_name(filter, _("HTML Files"));
+  gtk_file_filter_add_pattern(filter, "*.html");
+  gtk_file_filter_add_pattern(filter, "*.htm");
```
b4n: maybe rather use `gtk_file_filter_add_mime_type(filter, "text/html");`?

markdown/src/plugin.c
```diff
@@ -240,3 +250,47 @@ on_view_pos_notify(GObject *obj, GParamSpec *pspec, MarkdownViewer *viewer)
...
+
+  if (gtk_dialog_run(GTK_DIALOG(dialog)) == GTK_RESPONSE_ACCEPT) {
+    gchar *fn = gtk_file_chooser_get_filename(GTK_FILE_CHOOSER(dialog));
+    gchar *html = markdown_viewer_get_html(viewer);
+    GError *error = NULL;
+    if (! g_file_set_contents(fn, html, -1, &error)) {
+      dialogs_show_msgbox(GTK_MESSAGE_ERROR,
+        _("Failed to export Markdown HTML to file '%s': %s"),
+        fn, error->message);
+      g_error_free(error);
```
b4n: maybe re-show the save dialog in case of error?  I guess the most likely reason of failure is permission issue/invalid path, so I imagine the general action would be trying again.

markdown/src/plugin.c
codebrainz: Yeah, just being explicit.

markdown/src/plugin.c
codebrainz: Seems like a good idea if it does what it sounds like, is there similar for `*` (All files)?

markdown/src/plugin.c
codebrainz: Good idea.

markdown/src/plugin.c
b4n: Not sure, but I'm afraid not

PR #510, opened 2016-12-06
geanylua: Fix build with latest Scintilla

b4n:
When not defining `INCLUDE_DEPRECATED_FEATURES`, unprefixed structure names are not available anymore.  Thus, use prefixed names.

---
Since the last Scintilla update in Geany (3.7.1) unprefixed structure names are hidden.  This is good in theory, and will avoid thinks like 87d4bbcd16a41fce93b7da20c2135ff728b25791 in the future, but GeanyLua used unprefixed names.

So, in addition to the fix, do we want to restore the old names, or break API to reflect those are now removed, or do we consider it's not part of our API and thus don't care?  Of course as it's a structure name, it only affects building the code, the generated one is still binary compatible anyway.
@elextr @codebrainz @frlan

### DIFF
--- a/geanylua/glspi_sci.c
+++ b/geanylua/glspi_sci.c
@@ -739,7 +739,7 @@ static gint glspi_scintilla(lua_State* L)
 
 static gint glspi_find(lua_State* L)
 {
-	struct TextToFind ttf;
+	struct Sci_TextToFind ttf;
 
 	gint flags=0;
 	gint i,n;
@@ -820,7 +820,7 @@ struct CharacterRange {
 */
 
 /*
-struct TextToFind {
+struct Sci_TextToFind {
 		struct CharacterRange chrg;     // range to search
 		char *lpstrText;                // the search pattern (zero terminated)
 		struct CharacterRange chrgText; // returned as position of matching text


eht16:
I would say we should break the API, as Scintilla did and so we can stay close to Scintilla.

Just my 2cents.

b4n:
@eht16 by "breaking API", I mostly asked whether we should consider that the old names were part of the API or not.  I do think we should very likely not reintroduce those unprefixed name, we maybe we could just not care about them being gone.

elextr:
The unprefixed names should not be re-created.  The Geany API provides the Scintilla API as is, deprecations and breakages included.  So when the Scintilla API changes just bump the Geany API so plugins can adapt if they want to.

codebrainz:
+1 for break, especially if all Geany-Plugins are (now) working and 3rd party plugins can define `INCLUDE_DEPRECATED_FEATURES` before including `geanyplugin.h` to unbreak themselves if they don't want to use the nicer names.

frlan:
I wouldn't care about anything not in the API. Functions should be prefixed.

b4n:
@elextr @codebrainz @eht16 The thing is that we don't really have a way to reflect the change in the API version: we have one API number, and one ABI number.

* The change here doesn't break ABI (it's a build-only change).
* Changing the API number won't help anyone: the prefixed name exists since basically forever, so there's no real point in adding `#if`s in some code for that.

And changing neither of those will change the fact plugins using old unprefixed names will need to be changed to build against the latest Geany.

So, what could we do that has any kind of meaning?  We could indeed bump API to reflect new stuff in the Scintilla API (that could be a rule we follow in the future), but not necessarily this very change.

Opinions?  Or maybe I missed something we could actually do that would be sensible, dunno.

elextr:
Ok, if ALL the prefixed names already exist then just drop the unprefixed ones.  I suspect we will get very few questions, and the few we do get we can field in Github or ML.

b4n:
@elextr yes they all did exist before, unprefixed names were just `#define`s to the prefixed ones.

elextr:
> unprefixed names were just #defines to the prefixed ones.

ok, maybe #define them to `static_assert(true, "Use the prefixed name you idiot")`

b4n:
@elextr as tempting as it is, I don't really think its a great idea to reintroduce the unprefixed names to complain on people using them… is it? :)

b4n:
Maybe a note in Geany's NEWS is enough?  something like
```diff
@@ -3,6 +3,10 @@ Geany 1.30 (unreleased)
     Internationalization
     * Updated translations: lt
 
+    API
+    * Remove unprefixed Scintilla structure aliases.  Plugins must use
+      the `Sci_`-prefixed version from now on.
+
 
 Geany 1.29 (November 13, 2016)
 
```

frlan:
Please follow up elsewhere, going to merge this one now ;)

PR #511, opened 2016-12-17
Debugger: Move debug output accessing after check whether err != Null

frlan:
The debug-output is accessing err. If err is Null, it will fail most likely.

### DIFF
--- a/debugger/src/dbm_gdb.c
+++ b/debugger/src/dbm_gdb.c
@@ -1,18 +1,18 @@
 /*
  *      dbm_gdm.c
- *      
+ *
  *      Copyright 2010 Alexander Petukhov <devel(at)apetukhov.ru>
- *      
+ *
  *      This program is free software; you can redistribute it and/or modify
  *      it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


debugger/src/dbm_gdb.c
```diff
@@ -229,34 +229,38 @@ static void gdb_input_write_line(const gchar *line)
...
 
 	st = g_io_channel_flush(gdb_ch_in, &err);
 	if (err || (st == G_IO_STATUS_ERROR) || (st == G_IO_STATUS_EOF))
 	{
-#ifdef DEBUG_OUTPUT
-		dbg_cbs->send_message(err->message, "red");
-#endif
 		if (err)
+		{
+			#ifdef DEBUG_OUTPUT
```
b4n: style: the preproc directives used not to be indented

frlan: Actually I was thinking about whether or not but decided  for personal reasons ;)

frlan:
I just having Geany configured to purge trailing spaces etc so in the end its all cleaner.

PR #517, opened 2017-01-12
lineoperations: fixed -Werror=aggregate-return

smostertdev:
Oops, I think I broke Travis CI. 

I am returning a (small) structure when I should be using pointers.
This commit should fix this mess, my apologies.

### DIFF
--- a/lineoperations/src/lineoperations.c
+++ b/lineoperations/src/lineoperations.c
@@ -197,6 +197,7 @@ action_indir_manip_item(GtkMenuItem *menuitem, gpointer gdata)
 
 	lines_affected = func(lines, num_lines, new_file);
 
+
 	/* set new document */
 	sci_replace_sel(doc->editor->sci, new_file);
 
--- a/lineoperations/src/lineoperations.c
+++ b/lineoperations/src/lineoperations.c
@@ -41,24 +41,23 @@ struct lo_lines {
 
 /* selects lines in document (based off of lo_lines struct parameter) */
 static void
-select_lines(GeanyEditor *editor, struct lo_lines sel)
+select_lines(GeanyEditor *editor, struct lo_lines *sel)
 {
 	/* set the selection to beginning of first line */
 	sci_set_selection_start(editor->sci,
-			sci_get_position_from_line(editor->sci, sel.start_line));
+			sci_get_position_from_line(editor->sci, sel->start_line));
 
 	/* set the selection to end of last line */
 	sci_set_selection_end(editor->sci,
-			sci_get_line_end_position(editor->sci, sel.end_line) +
+			sci_get_line_end_position(editor->sci, sel->end_line) +
 									 editor_get_eol_char_len(editor));
 }
 
 
-/* get lo_lines struct based off of document */
-static struct lo_lines
-get_current_sel_lines(ScintillaObject *sci)
+/* get lo_lines struct 'sel' from document */
+static void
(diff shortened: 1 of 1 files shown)


codebrainz:
Weird that Travis breaks for perfectly valid code.

techee:
I think -Werror=aggregate-return is too aggressive and shouldn't be used. I run into the same problem in https://github.com/geany/geany/pull/1263 where some some functions in ctags return MIOPos which is a tiny structure. I'll prepare a patch to remove this flag from the build - this pull request will then be unnecessary.

techee:
Created https://github.com/geany/geany-plugins/pull/529. Depending whether it's OK the first patch from this pull request could be dropped. The second is still necessary as build fails without it.

lineoperations/src/lineoperations.c
```diff
@@ -160,27 +157,32 @@ action_indir_manip_item(GtkMenuItem *menuitem, gpointer gdata)
...
 	GeanyDocument *doc    = document_get_current();
 
-	struct lo_lines sel   = get_current_sel_lines(doc->editor->sci);
-	gint   num_lines      = (sel.end_line - sel.start_line) + 1;
-
 	gint   num_chars      = 0;
 	gint   i              = 0;
 	gint   lines_affected = 0;
 
+	struct lo_lines *sel = g_malloc(sizeof(struct lo_lines));
```
b4n: instead of allocating on the heap you could pass `&sel` to the `get_current_sel_lines()` call below

lineoperations/src/lineoperations.c
```diff
@@ -223,18 +227,26 @@ action_sci_manip_item(GtkMenuItem *menuitem, gpointer gdata)
 	/* function pointer to gdata -- function to be used */
 	gint (*func)(ScintillaObject *, gint, gint) = gdata;
 	GeanyDocument *doc  = document_get_current();
-	struct lo_lines sel = get_current_sel_lines(doc->editor->sci);
+
+	struct lo_lines *sel = g_malloc(sizeof(struct lo_lines));
```
b4n: ditto

smostertdev:
Thanks for all the input, guys.
Since the aggregate-return check was removed from Travis, should I still go ahead and make the small changes for this PR?
Do you have a preference on returning small structs/pointers @b4n? Either way is fine with me, I'm just lazy.

b4n:
@smostertdev For now I prefer not returning aggregates, but it's really your call as I can't really back my opinion up with facts right now, and @elextr and @codebrainz seem happy about it.

codebrainz:
@b4n I'm indifferent, I just mentioned that it's legal code.

elextr:
@b4n, the more its discussed the more convinced I become that returns of small values via pointer parameters instead of the function return value is ugly and inefficient.  It limits the benefits of inlining and defeats the compiler's ability to employ copy elision and requires the caller to explicitly declare a variable.  GCC's `aggregate-return` warning should be changed to `large-aggregate-return` instead.

frlan:
@codebrainz @b4n How going on here?

frlan:
@b4n Please review as soon as you have some spare time. Will postpone it to another release

frlan:
@b4n  Ping ;)

elextr:
@frlan, be patient, your previous ping isn't quite a year ago yet :)

LarsGit223:
LGBI

PR #521, opened 2017-02-15
Use online version of plugin help

techee:
This avoids platform-specific paths to the documentation and the html formating is also nicer.


b4n:
Just a note: BTW, one reason not to use the online version is to have a consistent version of the manual for the installer plugin without having to keep a dozen versions online.
But I get the point, and don't think it should be a problem.

frlan:
@b4n  I was also thinking about this and of users not having access to internet. We should find a general solution here but for now the patch did improve things a little IMHO

PR #526, opened 2017-02-16
GeanyPy: fix Document.text_changed

vfaronov:
@kugel- @codebrainz Mind having a look? I assume that `Py_RETURN_NONE` here was a typo or copy-paste error.

PR #527, opened 2017-02-17
Sync with latest upstream

codebrainz:
Closes #526

### DIFF
--- a/geanypy/src/Makefile.am
+++ b/geanypy/src/Makefile.am
@@ -5,11 +5,11 @@ geanyplugin_LTLIBRARIES		=	geanypy.la
 geanyplugindir				=	$(libdir)/geany
 
 geanypy_la_LDFLAGS			=	-module -avoid-version -Wl,--export-dynamic
-geanypy_la_CPPFLAGS			=	@GEANY_CFLAGS@ @PYGTK_CFLAGS@ @PYTHON_CPPFLAGS@ \
+geanypy_la_CPPFLAGS			=	@PYTHON_CPPFLAGS@ \
 								-DGEANYPY_PYTHON_DIR="\"$(libdir)/geany/geanypy\"" \
 								-DGEANYPY_PLUGIN_DIR="\"$(libdir)/geany\"" \
 								-DG_LOG_DOMAIN=\"GeanyPy\"
-geanypy_la_CFLAGS			=	@GEANYPY_CFLAGS@ @GMODULE_CFLAGS@
+geanypy_la_CFLAGS			=	@PYGTK_CFLAGS@ @GEANY_CFLAGS@ @GEANYPY_CFLAGS@ @GMODULE_CFLAGS@
 geanypy_la_LIBADD			=	@GEANY_LIBS@ @PYGTK_LIBS@ \
 								$(PYTHON_LDFLAGS) $(PYTHON_LIBS) \
 								@PYTHON_EXTRA_LIBS@ @PYTHON_EXTRA_LDFLAGS@ \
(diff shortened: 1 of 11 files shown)


geanypy/src/geanypy-editor.c
```diff
@@ -62,7 +62,7 @@ Editor_get_property(Editor *self, const gchar *prop_name)
 		PyObject *py_doc;
 		py_doc = (PyObject *) Document_create_new_from_geany_document(
 									self->editor->document);
-		if (py_doc && py_doc != Py_None)
+		if (!py_doc || py_doc == Py_None)
```
codebrainz: I didn't look into this difference, I just picked the kind of condition that made the branch execute when it seemed plausible to me.

geanypy/src/Makefile.am
```diff
@@ -5,11 +5,11 @@ geanyplugin_LTLIBRARIES		=	geanypy.la
 geanyplugindir				=	$(libdir)/geany
 
 geanypy_la_LDFLAGS			=	-module -avoid-version -Wl,--export-dynamic
-geanypy_la_CPPFLAGS			=	@GEANY_CFLAGS@ @PYGTK_CFLAGS@ @PYTHON_CPPFLAGS@ \
+geanypy_la_CPPFLAGS			=	@PYTHON_CPPFLAGS@ \
 								-DGEANYPY_PYTHON_DIR="\"$(libdir)/geany/geanypy\"" \
 								-DGEANYPY_PLUGIN_DIR="\"$(libdir)/geany\"" \
 								-DG_LOG_DOMAIN=\"GeanyPy\"
-geanypy_la_CFLAGS			=	@GEANYPY_CFLAGS@ @GMODULE_CFLAGS@
+geanypy_la_CFLAGS			=	@PYGTK_CFLAGS@ @GEANY_CFLAGS@ @GEANYPY_CFLAGS@ @GMODULE_CFLAGS@
```
kugel-: I think the previous code is better. *_CFLAGS derived from PKG_CONFIG_CHECK() contain include paths (and usually nothing else) so they should be added to automake's *_CPPFLAGS indeed.

kugel-:
One thing, the rest looks good to me.

geanypy/src/Makefile.am
codebrainz: I don't really mind since they all end up jammed in the same part of the command line, but it seemed logical to group the pre-processor flags and the compiler flags into their respective variables.

kugel-: *_CFLAGS derived from PKG_CONFIG_CHECK() contains mostly (usually exclusively) pre-processor flags, despite the name. And the order on the command line matters.

codebrainz: I guess as long as user environment `CPPFLAGS` override `CPPFLAGS` and likewise but after with `CFLAGS` it doesn't matter.

b4n: I don't think this changes matters much, but IMO it's not bad, rather good.  Yes, *pkg-config* will mostly fill the `*_CFLAGS` variable with preprocessor flags, but it might put other stuff here if it wants.  Also, it's fine putting preprocessor flags in `*_CFLAGS` AFAIK, only library order matters.

geanypy/src/geanypy-editor.c
b4n: Agreed, the previous one seems weird: returning `None` when `py_doc` is *not* `NULL` or `None`?  weird.

New condition seems more reasonable to me.  I wonder if `Py_RETURN_NONE` is different than `return Py_None`, but it doesn't change anything.

geanypy/src/geanypy-glog.c
```diff
@@ -0,0 +1,45 @@
+#if defined(HAVE_CONFIG_H) && !defined(GEANYPY_WINDOWS)
```
b4n: no license header?

geanypy/src/geanypy-editor.c
codebrainz: I think it's effectively like 

```c
#define Py_RETURN_NONE do { Py_INCREF(Py_None); return Py_None; } while (0)
``` 

or something similar.

geanypy/src/geanypy-glog.c
codebrainz: Not in each file.

vfaronov:
GeanyPy is listed in `MAINTAINERS` as “Currently diverged from upstream” — I figure that’s no longer correct?

PR #529, opened 2017-02-18
Remove -Werror=aggregate-return from travis flags

techee:
While in general one should avoid returning big structs for performance
reasons, it's no big problem for small structs and it can lead to more
readable code.

Having this check on for everything seems to be too strict - if someone
is returning big structs like crazy, someone will probably notice during
the review. Otherwise it's no problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -125,7 +125,7 @@ before_install:
 
 before_script:
   # prepare for GP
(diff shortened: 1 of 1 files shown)


elextr:
Agree and the same should happen with Geany itself, despite @b4n (IMHO old fashioned ugly and unneccessary) [request](https://github.com/geany/geany/pull/1382#discussion-diff-98349717R120).

PR #531, opened 2017-02-19
git-changebar: Add the possibility to undo hunk at cursor position

techee:
The patch adds a new keybinding action for undoing hunks. When cursor is
placed at a line with a diff and the keybinding is invoked, the
corresponding hunk is undone.

To get the previous contents at the given position, a temporary Scintilla
instance is used into which the previous state of the document is loaded
and the corresponding text from the previous state of the document is
extracted. This temporary Scintilla object is then destroyed.

### DIFF
--- a/git-changebar/src/gcb-plugin.c
+++ b/git-changebar/src/gcb-plugin.c
@@ -68,6 +68,10 @@ PLUGIN_SET_TRANSLATABLE_INFO (
 
 #define RESOURCES_ALLOCATED_QTAG \
   (g_quark_from_string (PLUGIN"/git-resources-allocated"))
+#define UNDO_LINE_QTAG \
+  (g_quark_from_string (PLUGIN"/git-undo-line"))
+#define DOC_ID_QTAG \
+  (g_quark_from_string (PLUGIN"/git-doc-id"))
 
 
 enum {
@@ -166,6 +170,7 @@ static GAsyncQueue     *G_queue               = NULL;
 static GThread         *G_thread              = NULL;
 static gulong           G_source_id           = 0;
 static gboolean         G_monitoring_enabled  = TRUE;
+static GtkWidget       *G_undo_menu_item      = NULL;
 static struct {
   gint    num;
   gint    style;
@@ -916,6 +921,8 @@ update_diff (const gchar *path,
       }
     }
     
+    gtk_widget_set_visible (G_undo_menu_item, contents != NULL);
+    
     if (contents && (allocated || allocate_resources (sci))) {
       diff_buf_to_doc (contents, doc, diff_hunk_cb, sci);
     } else if (! contents && allocated) {
@@ -977,6 +984,8 @@ update_diff_push (GeanyDocument  *doc,
                   gboolean        force)
 {
   g_return_if_fail (DOC_VALID (doc));
+  
+  gtk_widget_hide (G_undo_menu_item);
   
   if (G_source_id) {
     g_source_remove (G_source_id);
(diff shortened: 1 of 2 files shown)


techee:
This is something I wanted for a long time so here's an attempt to implement it.

Note that I based the patch mostly on what I have seen in git-changebar (basically combined the "goto next hunk" code and get_widget_for_buf_range()) without knowing libgit or the majority of the plugin code. So if anyone asks why I did something certain way, the answer is "because I didn't know what I was doing" :-).

techee:
I've added a fix for https://github.com/geany/geany-plugins/issues/532.

frlan:
I've got it inside local branch and it seems to work fine. But did not some kind of extended code review or something.

git-changebar/src/gcb-plugin.c
```diff
@@ -1044,13 +1053,13 @@ goto_next_hunk_diff_hunk_cb (const git_diff_delta *delta,
       if (data->next_line >= 0) {
         return 1;
       } else if (data->line < hunk->new_start - 1) {
-        data->next_line = hunk->new_start - 1;
+        data->next_line = (hunk->new_start == 0) ? 0 : hunk->new_start - 1;
```
b4n: maybe it'd be worth creating a function/macro for that, as it's repeated 3 times?  Not sure of a name for it yet though :)

git-changebar/src/gcb-plugin.c
```diff
@@ -1086,6 +1097,128 @@ on_kb_goto_next_hunk (guint kb)
...
+                                         "UTF-8", doc->encoding, NULL);
+  }
+  
+  scintilla_send_message (old_sci, SCI_ADDTEXT, old_buf_len, (glong) old_buf);
+
+  old_pos_start = sci_get_position_from_line (old_sci, old_start);
+  old_pos_end = sci_get_position_from_line (old_sci, old_start + old_lines);
+  old_range = sci_get_contents_range (old_sci, old_pos_start, old_pos_end);
+
+  sci_insert_text(doc->editor->sci, pos, old_range);
```
b4n: style: space before the open parenthesis

techee:
@b4n Thanks for having a look at this (I want this feature badly!!!). Please let me know if you have some other suggestions, I'll address them together with your above comments.

git-changebar/src/gcb-plugin.c
```diff
@@ -1086,6 +1106,131 @@ on_kb_goto_next_hunk (guint kb)
...
+      if (data->new_lines > 0) {
+        gint pos = sci_get_position_from_line(sci, data->new_start - 1);
+        sci_set_target_start(sci, pos);
+        pos = sci_get_position_from_line(sci, data->new_start + data->new_lines - 1);
+        sci_set_target_end(sci, pos);
+        sci_replace_target(sci, "", FALSE);
+      }
+
+      if (data->old_lines > 0) {
+        gint line = sci_get_current_line (sci);
```
b4n: Why current line?  couldn't it be `gint pos = sci_get_position_from_line (sci, data->new_start - 1);` instead?  I don't get exactly how you can be sure it's the current line, and not relying on that would allow for that to be used on non-current lines too (like adding an entry to the editor menu).

And I don't get why caring about position of the delete marker as we don't use them here, do we?

git-changebar/src/gcb-plugin.c
```diff
@@ -1086,6 +1106,131 @@ on_kb_goto_next_hunk (guint kb)
...
+          line++; /* marker for deleted hunk is on previous line except the 1st line */
+        }
+        pos = sci_get_position_from_line (sci, line);
+
+        insert_buf_range (doc, contents, pos,
+                          data->old_start - 1,
+                          data->old_lines);
+      }
+
+      sci_scroll_caret(sci);
```
b4n: Similarly, what about

```C
      scintilla_send_message (sci, SCI_SCROLLRANGE,
                              sci_get_position_from_line (sci, data->new_start + data->old_lines - 1),
                              sci_get_position_from_line (sci, data->new_start - 1));
```

git-changebar/src/gcb-plugin.c
b4n: hum, OK, the marker comment is a bit misleading, but indeed there's the same problem here, so we indeed need a similar fix, sorry.  But still not relying on the current line should be easy enough and better IMO.

techee: `gint pos = sci_get_position_from_line (sci, data->new_start - 1);` returns the pos on the previous line which is wrong when undoing deleted-only diff - I need to advance it to the next line if `new_lines == 0` (see a few lines below). I could probably get the length of the line and add it to pos in this case.

Just to understand - what does scintilla return from sci_get_current_line() when you right-click somewhere - I'd expect it doesn't change the current line, does it?

git-changebar/src/gcb-plugin.c
techee: Maybe something like

```
data->next_line = REMOVED_MARKER_POS (hunk->new_start);
```

?

techee:
> Looks mostly good, but I'd rather not rely on where the caret is. Not relying on it makes it fairly easy to add a right-click version of the feature

Maybe I misunderstand how you meant it but how does the right-click version know at which position to undo when it don't use the caret position? I'd expect the caret is somewhere where you want to undo, you right-click (at which point I expect Scintilla still preserves the caret position but I may be wrong) and select undo from the menu. But you still need the caret position, not?

git-changebar/src/gcb-plugin.c
b4n: > `gint pos = sci_get_position_from_line (sci, data->new_start - 1);` returns the pos on the previous line which is wrong when undoing deleted-only diff - I need to advance it to the next line if `new_lines == 0` (see a few lines below).

Indeed, I was confused at first.  I suggest 00e37457c20c94e23b1eef8ea98ea1977c446635

> Just to understand - what does scintilla return from sci_get_current_line() when you right-click somewhere - I'd expect it doesn't change the current line, does it?

No it doesn't, but Geany gives us the position of the right click, and other features in the menu work at the click position, not the caret position.  So for the feature I want to do the same: propose undoing the diff at the right click position, not at the caret one.  See 0c4415dbe61792de805e45e72433e4feaac988f5 for a proposal.

git-changebar/src/gcb-plugin.c
b4n: It's actually similarly wrong, but there's a fixed version in 00e37457c20c94e23b1eef8ea98ea1977c446635

b4n:
> Maybe I misunderstand how you meant it but how does the right-click version know at which position to undo when it don't use the caret position? I'd expect the caret is somewhere where you want to undo, you right-click (at which point I expect Scintilla still preserves the caret position but I may be wrong) and select undo from the menu. But you still need the caret position, not?

As said above, no: the right click undo would undo at the position of the click, which indeed doesn't move the caret, which means we need be able to undo arbitrary hunks regardless of the caret position.

b4n:
A series of proposed fixes and changes: https://github.com/b4n/geany-plugins/commits/git-changebar/techee/sidebar_undo/review
Feel free to reuse, squash, comment on, bash, nitpick, throw away or anything else you could thunk of :)

git-changebar/src/gcb-plugin.c
b4n: Yeah OK sounds good

b4n:
BTW, you maybe could add a line to the README about the new feature :)

techee:
Thanks @b4n - I've finally had time to have a look at it. Your patches look good (I reviewed only the patches dealing with the undo stuff and didn't spend much time looking at the menu feature but it seems to work fine). I've just squashed your whitespace fix to the original commit and left the rest as separate commits. I've also updated README and added the REMOVED_MARKER_POS() macro.

The last patch is one more minor modification - it updates the cursor position when undoing deleted hunk - originally the cursor was placed at the position where the hunk started but I think it should be placed at the end of the undone deletion which corresponds more closely to what normal undo does. The question is what to do with modified hunks (i.e. those with both old lines and new lines). I still find it more natural to place the cursor behind them so it behaves the same way as delete-only hunk.

frlan:
@b4n  @techee Merging it for 1.31?

techee:
> @b4n @techee Merging it for 1.31?

I'm fine with it, depends on Colomban.

frlan:
postponed to 1.32

vfaronov:
I have been running with this change for some time. Very useful, thank you @techee and @b4n. No problems so far (I have only a few small repos though).

frlan:
As I'm also running this patchset for some time now I'm going to merge it. Let us more beta test it on master.

PR #535, opened 2017-02-20
Add additional checks for doc != NULL and removing such ones later

frlan:
This shall ensure we don't access doc before checking it against NULL. also it stops cppcheck from failing.


b4n:
S/A #517

smostertdev:
Yes, this looks good. Go ahead a merge this, and I'll fix the other PR.

PR #536, opened 2017-02-20
treebrowser: Don't change directory on project save

techee:
This behavior might be unwanted when changing project settings which
is unrelated to project base path and also fixes filebrowser behavior
with the updated way project-save signal is emitted.

Merging should wait until how the discussion in https://github.com/geany/geany/pull/1400 evolves.


techee:
I wanted to get some feedback on the updated patches in https://github.com/geany/geany/pull/1400 before merging this. But hopefully the patch isn't too problematic and if something pops up in Geany, I can provide a patch for geany-plugins.

PR #540, opened 2017-03-04
Add note to plugins which are currently not actively maintained

frlan:
This is based on MAINTAINERS file. 

Idea is, to show enduser that the plugin might need its help.

### DIFF
--- a/geanydoc/src/geanydoc.c
+++ b/geanydoc/src/geanydoc.c
@@ -47,7 +47,7 @@ static GtkWidget *keyb2;
 PLUGIN_VERSION_CHECK(224)
 /* All plugins must set name, description, version and author. */
 PLUGIN_SET_TRANSLATABLE_INFO(LOCALEDIR, GETTEXT_PACKAGE,
-	_("Doc"), _("Call documentation viewer on current symbol."), VERSION,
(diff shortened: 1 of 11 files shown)


codebrainz:
I don't like the wording "no active development" since that doesn't really mean anything but it might turn users to a different plugin. For example the Scope plugin is probably the best debugging plugin, but this marks it with a big warning (which doesn't mean anything).

IMO it would be better to put a note at the end of the description like "This plugin currently has no maintainer. Would you like to help by contributing to this plugin?" or something like that.

If not that, then it should at least just say "No active maintainer" at the end of the description, not in stars at the very front.

frlan:
I did not at it at the end of description as I wanted to make it prominent to the user.

frlan:
@codebrainz Do you have a suggestion for a better wording? We got a bunch of orphaned and odd-fixes  plugins which might not work as long as nobody is willing to adopt them.

codebrainz:
I gave two in my previous message:

> This plugin currently has no maintainer. Would you like to help by contributing to this plugin?

And

> No active maintainer

frlan:
Ahdammid. Thanks for rereading your post to me. Good ideas. I will change my PR

frlan:
@codebrainz  I did rewording. I hope I improved it a little ;)

PR #550, opened 2017-03-27
Add prj_org != NULL check to fix project reload crash.

rwmaudsley:


kugel-:
This fix resolves the crash but the @techee has opened #555 as an alternative fix.

techee:
Sorry @ghost, I didn't notice your patch - I think it's fine to fix the crash alone. However, in #555 I added another sanity check now to avoid expanding incorrect paths if another project is opened immediately afterwards so I'd prefer if #555 could be applied instead of your patch.

frlan:
As by @techee I will not apply this one for the moment.

frlan:
But thanks for contributing. (I always forget to say it but even not merged PR are good support)

Issue #549, opened 2017-03-27
Project organizer crashes when closing project


interoceto:
After updating to geany 1.30.1 + geany-plugins 1.30, geany allways crashes (segmentation fault) when closing project with project organizer plugin enabled.
Tested on windows 10 & ubuntu 16.04 x64

rwmaudsley:
Submitted pull request https://github.com/geany/geany-plugins/pull/550 for this.

kugel-:
@techee FYI

```
Thread 1 "geany" received signal SIGSEGV, Segmentation fault.
0x00007fffe0016aa2 in expand_path (utf8_expanded_path=0x13cf090 "", select=select@entry=0) at prjorg-sidebar.c:1168
1168		foreach_slist (elem, prj_org->roots)
(gdb) bt
#0  0x00007fffe0016aa2 in expand_path (utf8_expanded_path=0x13cf090 "", select=select@entry=0) at prjorg-sidebar.c:1168
#1  0x00007fffe0016c5f in expand_on_idle (ptr=0x8659a0) at prjorg-sidebar.c:1224
#2  0x00007ffff529945a in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#3  0x00007ffff5299810 in  () at /usr/lib/libglib-2.0.so.0
#4  0x00007ffff5299b32 in g_main_loop_run () at /usr/lib/libglib-2.0.so.0
#5  0x00007ffff724ed05 in gtk_main () at /usr/lib/libgtk-3.so.0
#6  0x00007ffff79a7a40 in main_lib (argc=<optimized out>, argv=<optimized out>) at ../../src/libmain.c:1236
#7  0x00007ffff4cad511 in __libc_start_main () at /usr/lib/libc.so.6
#8  0x000000000040061a in _start ()
```

Reverting 8111515 helps

techee:
I can't reproduce it myself but I think the reason is that the idle function gets called after project is closed. Can someone confirm that #555 helps?

codebrainz:
I think I experienced this bug very infrequently. P/O is one of the plugins I always have activated and once in a blue moon I got a crasher. Hopefully this is the problem because it didn't happen enough for me to troubleshoot it :)

interoceto:
Hi!
[...]

PR #555, opened 2017-03-30
projectorganizer: Make sure project is open before trying to expand the tree

techee:
Since expand_on_idle() is called as an idle function, it can happen it's
executed after project is closed in which case it should just return
without accessing variables with invalid values.

Should fix crash reported in #549.

### DIFF
--- a/projectorganizer/src/prjorg-sidebar.c
+++ b/projectorganizer/src/prjorg-sidebar.c
@@ -48,6 +48,13 @@ typedef enum
 	MATCH_PATTERN
 } MatchType;
 
+typedef struct
+{
+	GeanyProject *project;
+	GPtrArray *expanded_paths;
+} ExpandData;
+
+
 static GdkColor s_external_color;
 static GtkWidget *s_toolbar = NULL;
 static gboolean s_pending_reload = FALSE;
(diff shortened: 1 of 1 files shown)


codebrainz:
@techee wouldn't it be better to keep track of the source IDs and connect/disconnect them as a project is opened/closed? Seems more future proof, IMO.

kugel-:
@techee The crash is gone. But you have to tell if it's the proper fix or not.

techee:
@codebrainz Yeah, it might theoretically happen that one project is closed, another opened afterwards after which the idle function is called still for the first project and expand incorrect paths as a result.

I've just submitted a new patch - instead of saving the source IDs which I'd have to store to some global variable, I just pass the Geany project pointer to the idle function and there check if it's still the same project. Should do the same job.

codebrainz:
> instead of saving the source IDs which I'd have to store to some global variable

Plugins can now have associated data, so you could just store the handler ID in there with other plugin-wide state. In any case, I'm sure it's fine, my OCD just doesn't like when code doesn't undo this kind of stuff, which is why I'm such a big fan of C++ RAII :)

frlan:
I vererted the merge as on further testing I'm getting

```
Thread 1 "geany" received signal SIGSEGV, Segmentation fault.
0x00007fff898889c2 in expand_path (utf8_expanded_path=0x555556835670 "/tmp/gis_QFVDYY.", select=select@entry=1)
    at prjorg-sidebar.c:1175
1175		foreach_slist (elem, prj_org->roots)
(gdb) bt
#0  0x00007fff898889c2 in expand_path (utf8_expanded_path=0x555556835670 "/tmp/gis_QFVDYY.", 
    select=select@entry=1) at prjorg-sidebar.c:1175
#1  0x00007fff89888b4e in expand_on_idle (expand_data=0x55555683cfb0) at prjorg-sidebar.c:1240
#2  0x00007ffff50c76aa in g_main_context_dispatch () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x00007ffff50c7a60 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#4  0x00007ffff50c7d82 in g_main_loop_run () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#5  0x00007ffff741d3b7 in gtk_main () from /usr/lib/x86_64-linux-gnu/libgtk-x11-2.0.so.0
#6  0x00007ffff79aa107 in main_lib (argc=<optimized out>, argv=<optimized out>) at libmain.c:1233
#7  0x00007ffff4ae22b1 in __libc_start_main () from /lib/x86_64-linux-gnu/libc.so.6
#8  0x00005555555547ba in _start ()
```
Please double check (no plugins active)

frlan:
It crash on an Geany without an open document only instant save from saveactions was activated

techee:
Should be fixed in #559. Sorry for not noticing this myself, I just usually have some project open.

PR #564, opened 2017-04-19
git-changebar: Fix spurious line wrapping

### LINKED ISSUE #425: git-changebar: Height of overlay not adjusted when line wrap is in effect.
When linewrap is in effect the height of the overlay is not adjusted if the old line is too long.

Overlay when line wrap is activated.
linewrap-on

Overlay when line wrap is off. 
linewrap-off

### LINKED ISSUE #567: git-changebar: Lines are wrapped when they don’t need to be
1. Enable Git Change Bar.
2. Enable line wrapping in Geany Preferences → Editor.
3. Open a file that is under Git control (can’t reproduce on files that had been opened before line wrapping was enabled).
4. Change a (short) line in that file.
5. Point the mouse on the margin indicating a changed hunk.

Result: the line is wrongly wrapped in the tooltip (and its ending cannot be seen because of #425):

git-changebar-wrapped

Expected result: same as with line wrapping disabled:

git-changebar-not-wrapped

Note: #425 is about the fact that the wrapped part of the line cannot be seen; this issue is about the fact that the line should not be wrapped in the first place.

vfaronov:
This fixes #567 for me, but I haven’t investigated exactly why this extra pixel is necessary, so I may be doing something wrong.

### DIFF
--- a/git-changebar/src/gcb-plugin.c
+++ b/git-changebar/src/gcb-plugin.c
@@ -802,8 +802,11 @@ get_widget_for_buf_range (GeanyDocument *doc,
       break;
     }
   }
+  /* We need 2 extra pixels of width:
+   * 1 to avoid cropping the rightmost vertical bar of letters like H and M,
+   * 1 to avoid spurious line wrapping (issue #425). */
   gtk_widget_set_size_request (GTK_WIDGET (sci),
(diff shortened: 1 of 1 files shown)


b4n:
And thanks for also thinking of adding a comprehensive comment on the magic `+2` :)

Issue #507, opened 2016-11-24
Project Organizer: Help-Button not working


frlan:
At least under MacOS 10.11.6 running 1.29 the Help-button inside Tools->Plugin Manager is getting activated, but seem not to work.

techee:
@frlan This isn't related to projectorganizer only but happens for every plugin. The problem is that right now the installation path of the documentation is defined statically at the compilation time. This works for Linux where the documentation is always at the same location but doesn't work for Windows or OS X where the documentation is relative to the installation path (or the location of the bundle).

This would have to be fixed in Geany which should specify the location of its documentation relatively to the binary.

eht16:
On Windows, it is not completely broken for every plugin.
The following plugins (from the whole list of plugins available on Windows and have any Help reference implemented) are broken:
- commander
- geanyctags
- scope

A working example is:
geanygendoc

@techee we have ifdef's for MINGW in the Makefiles of some plugins which take care to make the documentation relative.
This could be ported for MacOS I guess but I agree a more generic solution would be nicer.
Maybe we should add some convenience or helper function to the API which helps plugins to construct the path at runtime with the help of the installation directory.

techee:
I went through other plugins and most of them use the online html version of the documentation which is probably the easiest solution. I created a pull request for geanyctags and projectorganizer here: https://github.com/geany/geany-plugins/pull/521.

techee:
So I just ran into this problem at another place

https://github.com/geany/geany-osx/issues/6

where some data files aren't read because absolute paths are used in the code. There are at least 5 plugins affected by this - I didn't notice this myself because I have these files installed at the given paths but people who only get the bundle run into this problem.

So now the question is what to do - there are unfortunately various problems:
1. Geany installation prefix can be different from the plugins installation prefix. So Geany itself should probably not contain the code for determining plugins data directory.
2. On OS X Geany and plugins can be built in two ways - without bundle (installed similarly to linux) and with bundle (all in one package). The first one is actually always done in order to create the bundle and is also useful for testing (apart from it also e.g. MacPorts uses this method) while the bundle is what normal users get from the Geany downloads site. Both of them are important and should be supported. This means that the correct method for determining path has to be determined dynamically unlike Windows or Linux where either relative or absolute paths are used and this cannot be done during configure.
[...]

PR #569, opened 2017-04-26
Use paths relative to the bundle on OS X

techee:
These patches try to fix the problems described in https://github.com/geany/geany-plugins/issues/507 for various plugins. They address only "serious" issues which prevent correct operation of the plugin like not locating .glade or .ui files, not "minor" problems like not showing documentation when clicking the Help button of the plugin.

There's the same function get_data_dir_path() in each plugin which locates the file in the correct directory for each platform. For OS X laucher script I added an environment variable GEANY_PLUGINS_SHARE_PATH which is set before the binary is started so it should be defined and can be used by the plugin.

### DIFF
--- a/geanygendoc/src/ggd-utils.c
+++ b/geanygendoc/src/ggd-utils.c
@@ -135,6 +135,24 @@ ggd_copy_file (const gchar *input,
   return success;
 }
 
+static gchar *
+get_data_dir_path (const gchar *filename)
+{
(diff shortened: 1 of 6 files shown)


frlan:
I'm using this patchset already on MacOS as well as did a smoke test for my Debian. Needs further testing on other platforms that's why fast merging. But looks fine to me

PR #586, opened 2017-06-24
Pass a GdkKeymap to gdk_keymap_* functions

### LINKED ISSUE #585: Macros and Numbered Bookmarks plugins crash Geany
I've tried using on 2 different PC's Both Windows, 1 is 10 Enterprise and the other 8.1 Pro,  both 64bit.

When I select the "Macros" option in the plugin manager is when it crashes.

eht16:
This fixes crashes and critial warnings.
Fixes #585.

On Windows passing NULL as GdkKeymap crashed Geany, on other platforms there were "only" CRITICALs but possibly it could crash Geany there as well.
Using `gdk_keymap_get_default()` should suffice to use a valid GdkKeymap.


frlan:
Looks good to me. Merging for further testing.

PR #588, opened 2017-07-04
fixed exec syntax

daavve:
Haven't used exec myself, but geanypy will not compile for me without this change.

-Dave

### DIFF
--- a/geanypy/geany/console.py
+++ b/geanypy/geany/console.py
@@ -584,7 +584,7 @@ class _Console(_ReadLine, code.InteractiveInterpreter):
         except: pass
 
         try:
-            exec "import __builtin__" in self.locals
+            exec("import __builtin__" in self.locals)
             strings.extend(eval("dir(__builtin__)", self.locals))
         except:
             pass


vfaronov:
What exactly do you mean by “will not compile”? Sounds like a Python 2 vs. 3 problem. `exec code in scope` is Python 2-specific syntax. `exec(code in scope)` is meaningless, because `code in scope` is a boolean expression that will always evaluate to `False` in this context.

daavve:
Sorry, my bad.

I checked the Makefile and found I am using python3 instead of python2.

I am running ArchLinux and python3 is the linked to /usr/bin/python
I had to change the following in the Makefile:

350,351c350,351
< PYTHON = /usr/bin/python
< PYTHON_CPPFLAGS = -I/usr/include/python3.6m
---
> PYTHON = /usr/bin/python2
> PYTHON_CPPFLAGS = -I/usr/include/python2.7
354,356c354,356
< PYTHON_LIBRARY = /usr/lib/x86_64-linux-gnu/libpython3.6m.so.1.0
< PYTHON_LIBS = -L/usr/lib -lpython3.6m
< PYTHON_SITE_PKG = /usr/lib/python3.6/site-packages
---
> PYTHON_LIBRARY = /usr/lib/libpython2.7.so
> PYTHON_LIBS = -L/usr/lib -lpython2.7
> PYTHON_SITE_PKG = /usr/lib/python2.7/site-packages


Where do I need to go to fix the generated makefile?


Thanks,

-Dave

daavve:
Figured it out.  I need to set my environment variable $PYTHON to /usr/bin/python2 before running autogen.sh

Issue #596, opened 2017-08-12
Travis fails, needs updating to C++11


elextr:
Geany now needs a C++11 compiler due to Scintilla 3.7.5 and therefore so does Geany Plugins since it uses the Geany compilation options, but Travis uses an old compiler so it fails.  For info see https://github.com/geany/geany/pull/1503.

LarsGit223:
I think this can be closed, see https://github.com/geany/geany-plugins/pull/597#issuecomment-322063066.

PR #598, opened 2017-08-15
Workbench

LarsGit223:
This is a new plugin called workbench.

It has the following features:
- assign multiple projects to a workbench
- assign directories and file patterns to each project for easy file listing
- create "global" workbench or project bookmarks to have quick access to often used files
- the plugin uses relative pathes (a project file path is relative to the workbench file path, a file path is relative to the project file path)
- languages: english and german

For instructions please see the readme file.

The plugin is heavily based on the project organizer plugin, thanks to Jiří Techet. Icons are taken from the gnome desktop.

### DIFF
--- a/workbench/src/wb_project.c
+++ b/workbench/src/wb_project.c
@@ -234,7 +234,10 @@ static GSList *wb_project_dir_get_file_list(WB_PROJECT_DIR *root, const gchar *u
 	dir = g_dir_open(locale_path, 0, NULL);
 	if (!dir || !real_path || g_hash_table_lookup(visited_paths, real_path))
 	{
-		g_dir_close(dir);
+		if (dir != NULL)
+		{
+			g_dir_close(dir);
+		}
 		g_free(locale_path);
 		g_free(real_path);
 		return NULL;


workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+					file++;
+				}
+				g_strfreev(bookmarks_strings);
+			}
+
+			/* Load project dirs */
+			for (index = 1 ; index < 1025 ; index++)
+			{
+				g_snprintf(key, sizeof(key), "Dir%u-BaseDir", index);
+				if (!g_key_file_has_key (kf, "Workbench", key, NULL))
```
b4n: this isn't really useful as [`g_key_file_get_string()`](https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html#g-key-file-get-string) (below) returns `NULL` if the key is not found

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+
+		g_key_file_free(kf);
+		g_free (contents);
+		success = TRUE;
+	}
+	else if (error != NULL)
+	{
+		g_set_error (error, 0, 0,
+					 "Internal error: param missing (file: %s, line %d)",
+					 __FILE__, __LINE__);
```
b4n: Here again, it probably rather calls for a `g_return_val_if_fail(prf != NULL, FALSE)` at the start of the function.  `GErrors` are supposed to carry runtime errors (like IO error, file missing, etc.), not programming errors (invalid parameters, etc.).

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+ *
+ * @return Address of the new structure.
+ *
+ **/
+WB_PROJECT *wb_project_new(gchar *filename)
+{
+	WB_PROJECT *new_prj;
+
+	new_prj = g_new(WB_PROJECT, 1);
+	memset(new_prj, 0, sizeof(*new_prj));
```
b4n: You could use [`g_new0()`](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-new0) (or my favorite `new_prj = g_malloc0(sizeof *new_prj)` which doesn't repeat the type, but that's another story 😄)

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+ **/
+void wb_project_free(WB_PROJECT *prj)
+{
+	GSList *elem;
+
+	/* Free directories first */
+	foreach_slist(elem, prj->directories)
+	{
+		wb_project_dir_free(elem->data);
+	}
```
b4n: shouldn't you `g_slist_free(prj->directories)`?

workbench/src/workbench.c
```diff
@@ -0,0 +1,804 @@
...
+}
+
+
+/** Set the filename.
+ *
+ * @param wb       The workbench
+ * @param filename Name of the workbench file
+ *
+ **/
+void workbench_set_filename(WORKBENCH *wb, gchar *filename)
```
b4n: `filename` argument could be `const`

workbench/src/workbench.c
```diff
@@ -0,0 +1,804 @@
...
+
+
+/** Add a project to the workbench.
+ *
+ * @param wb       The workbench
+ * @param filename Project file
+ * @return TRUE on success, FALSE otherwise
+ *
+ **/
+gboolean workbench_add_project(WORKBENCH *wb, gchar *filename)
```
b4n: Same, `filename` could be `const`

workbench/src/workbench.c
```diff
@@ -0,0 +1,804 @@
...
+
+/* Add a workbench bookmark */
+static gboolean workbench_add_bookmark_int(WORKBENCH *wb, gchar *filename)
+{
+	if (wb != NULL)
+	{
+		gchar *new;
+
+		new = g_strdup(filename);
+		if (new != NULL)
```
b4n: one more place where, if `filename` is not `NULL`, the test is not relevant as `g_strdup()` aborts on allocation failure

workbench/src/workbench.c
```diff
@@ -0,0 +1,804 @@
...
+		/* Save projects data */
+		for (index = 0 ; index < wb->projects->len ; index++)
+		{
+			entry = g_ptr_array_index(wb->projects, index);
+			g_snprintf(group, sizeof(group), "Project-%u", (index+1));
+			g_key_file_set_string(kf, group, "AbsFilename", entry->abs_filename);
+			g_key_file_set_string(kf, group, "RelFilename", entry->rel_filename);
+			g_key_file_set_boolean(kf, group, "UseAbsFilename", entry->use_abs);
+		}
+		contents = g_key_file_to_data (kf, &length, error);
```
b4n: in the (unlikely) event `g_key_file_to_data()` fails, you should make sure to properly handle it, because it is invalid to set a `GError` on top of another, and you reuse `error` in the `g_file_set_contents()` call below.

workbench/src/workbench.c
```diff
@@ -0,0 +1,804 @@
...
+		}
+		else
+		{
+			valid = FALSE;
+		}
+
+		if (!valid)
+		{
+			g_set_error (error, 0, 0,
+						 "File: %s is not a valid workbench file!",
```
b4n: probably worth translating?

LarsGit223:
@b4n:
As a general comment first a big thanks for the detailed code review :thumbsup:
I will come back to the specific points as soon as I got time to address them.

workbench/icons/Makefile.am
```diff
@@ -0,0 +1,20 @@
+icondir = $(datadir)/icons/hicolor/16x16/apps
```
LarsGit223: How do I do that? Just append more icon files to the makefile. How are the icons selected by the application? (The variable ```icondir``` doesn't seem to be used by the makefile???)

workbench/src/dialogs.c
```diff
@@ -0,0 +1,411 @@
...
+ **/
+gchar *dialogs_create_new_workbench(void)
+{
+	gchar *utf8_filename = NULL;
+	GtkWidget *dialog;
+
+	dialog = gtk_file_chooser_dialog_new(_("Create new workbench"),
+		GTK_WINDOW(wb_globals.geany_plugin->geany_data->main_widgets->window), GTK_FILE_CHOOSER_ACTION_SAVE,
+		_("_Cancel"), GTK_RESPONSE_CANCEL,
+		_("Create"), GTK_RESPONSE_ACCEPT, NULL);
```
LarsGit223: Done.

workbench/src/dialogs.c
```diff
@@ -0,0 +1,411 @@
...
+	gtk_file_chooser_set_current_name(GTK_FILE_CHOOSER(dialog), "new.geanywb");
+	gtk_file_chooser_set_do_overwrite_confirmation (GTK_FILE_CHOOSER(dialog), TRUE);
+
+	if (gtk_dialog_run(GTK_DIALOG(dialog)) == GTK_RESPONSE_ACCEPT)
+	{
+		gchar *locale_filename = gtk_file_chooser_get_filename(GTK_FILE_CHOOSER(dialog));
+		utf8_filename = utils_get_utf8_from_locale(locale_filename);
```
LarsGit223: Done.

workbench/src/dialogs.c
```diff
@@ -0,0 +1,411 @@
...
+gchar *dialogs_open_workbench(void)
+{
+    gchar *utf8_filename = NULL;
+	GtkWidget *dialog;
+	GtkFileFilter *filter;
+
+	dialog = gtk_file_chooser_dialog_new(_("Open workbench"),
+		GTK_WINDOW(wb_globals.geany_plugin->geany_data->main_widgets->window), GTK_FILE_CHOOSER_ACTION_OPEN,
+		_("_Cancel"), GTK_RESPONSE_CANCEL,
+		_("Open"), GTK_RESPONSE_ACCEPT, NULL);
```
LarsGit223: Done.

workbench/src/dialogs.c
```diff
@@ -0,0 +1,411 @@
...
+/** Shows the dialog "Open workbench".
+ *
+ * The dialog lets the user choose an existing workbench file (filter *.geanywb).
+ *
+ * @return The filename
+ *
+ **/
+gchar *dialogs_open_workbench(void)
+{
+    gchar *utf8_filename = NULL;
```
LarsGit223: Fixed.

workbench/src/dialogs.c
```diff
@@ -0,0 +1,411 @@
...
+	GtkWidget *dialog;
+	GtkFileFilter *filter;
+
+	dialog = gtk_file_chooser_dialog_new(_("Open workbench"),
+		GTK_WINDOW(wb_globals.geany_plugin->geany_data->main_widgets->window), GTK_FILE_CHOOSER_ACTION_OPEN,
+		_("_Cancel"), GTK_RESPONSE_CANCEL,
+		_("Open"), GTK_RESPONSE_ACCEPT, NULL);
+
+	filter = gtk_file_filter_new ();
```
LarsGit223: Sorry...did oversee that ones. I have read the coding guidelines and I try to follow them. Fixed.

workbench/src/dialogs.c
```diff
@@ -0,0 +1,411 @@
...
+	gtk_file_chooser_add_filter (GTK_FILE_CHOOSER(dialog), filter);
+	filter = gtk_file_filter_new ();
+	gtk_file_filter_set_name (filter, _("All Files"));
+	gtk_file_filter_add_pattern (filter, "*");
+	gtk_file_chooser_add_filter (GTK_FILE_CHOOSER(dialog), filter);
+
+	if (gtk_dialog_run(GTK_DIALOG(dialog)) == GTK_RESPONSE_ACCEPT)
+	{
+		gchar *locale_filename = gtk_file_chooser_get_filename(GTK_FILE_CHOOSER(dialog));
+		utf8_filename = utils_get_utf8_from_locale(locale_filename);
```
LarsGit223: Done.

workbench/src/dialogs.c
```diff
@@ -0,0 +1,411 @@
...
+gchar *dialogs_add_project(void)
+{
+    gchar *utf8_filename = NULL;
+	GtkWidget *dialog;
+	GtkFileFilter *filter;
+
+	dialog = gtk_file_chooser_dialog_new(_("Add project"),
+		GTK_WINDOW(wb_globals.geany_plugin->geany_data->main_widgets->window), GTK_FILE_CHOOSER_ACTION_OPEN,
+		_("_Cancel"), GTK_RESPONSE_CANCEL,
+		_("Add"), GTK_RESPONSE_ACCEPT, NULL);
```
LarsGit223: Done.

workbench/src/dialogs.c
```diff
@@ -0,0 +1,411 @@
...
+ **/
+gchar *dialogs_add_directory(WB_PROJECT *project)
+{
+	gchar *utf8_filename = NULL;
+	GtkWidget *dialog;
+
+	dialog = gtk_file_chooser_dialog_new(_("Add directory"),
+		GTK_WINDOW(wb_globals.geany_plugin->geany_data->main_widgets->window), GTK_FILE_CHOOSER_ACTION_SELECT_FOLDER,
+		_("_Cancel"), GTK_RESPONSE_CANCEL,
+		_("Add"), GTK_RESPONSE_ACCEPT, NULL);
```
LarsGit223: Done.

workbench/src/dialogs.c
```diff
@@ -0,0 +1,411 @@
...
+ *
+ * The dialog lets the user choose an existing project file
+ * (filter *.geany or *).
+ *
+ * @return The filename
+ *
+ **/
+gchar *dialogs_add_project(void)
+{
+    gchar *utf8_filename = NULL;
```
LarsGit223: Fixed.

workbench/src/dialogs.c
```diff
@@ -0,0 +1,411 @@
...
+	if (project != NULL)
+	{
+		gchar *path;
+
+		/* Set the current folder to the location of the project file */
+		path = wb_project_get_filename(project);
+		if (path != NULL)
+		{
+			path = g_path_get_dirname(path);
+			gtk_file_chooser_set_current_folder(GTK_FILE_CHOOSER(dialog), path);
```
LarsGit223: As I have removed the encoding conversion in the dialogs now, yes. The filename is set with the  filename returned by the file chooser dialog. ```wb_project_get_filename()``` returns that filename (a copy).

workbench/src/menu.c
```diff
@@ -0,0 +1,310 @@
...
+ **/
+gboolean menu_init(void)
+{
+	/* Create menu and root item/label */
+	menu_data.menu = gtk_menu_new();
+	menu_data.root_item = gtk_menu_item_new_with_label(_("Workbench"));
+	gtk_widget_show(menu_data.root_item);
+
+	/* Create new menu item "New Workbench" */
+	menu_data.item_new = gtk_menu_item_new_with_mnemonic(_("New"));
```
LarsGit223: Added mnemonics.

> Geany follows this in its own UI, so it would also be better for integration.

Are there any more detailed rules about this? E.g. "Open" has an ellipsis, "Settings" not.

workbench/src/menu.c
```diff
@@ -0,0 +1,310 @@
...
+	gtk_widget_show(menu_data.item_new);
+	gtk_menu_append(GTK_MENU (menu_data.menu), menu_data.item_new);
+	g_signal_connect(menu_data.item_new, "activate",
+					 G_CALLBACK(item_new_workbench_activate_cb), NULL);
+	geany_plugin_set_data(wb_globals.geany_plugin, menu_data.item_new, NULL);
```
LarsGit223: Hmm...removed the calls. I do not remember exactly. Maybe copied from some place. I also thought I can use it to get the items destroyed on ending geany/the plugin.

workbench/src/menu.c
```diff
@@ -0,0 +1,310 @@
...
+	gtk_menu_append(GTK_MENU (menu_data.menu), menu_data.item_close);
+	g_signal_connect(menu_data.item_close, "activate",
+					 G_CALLBACK(item_close_workbench_activate_cb), NULL);
+	geany_plugin_set_data(wb_globals.geany_plugin, menu_data.item_close, NULL);
+
+	/* Add our menu to the main window (left of the help menu) */
+	gtk_menu_item_set_submenu(GTK_MENU_ITEM(menu_data.root_item), menu_data.menu);
+	gtk_menu_shell_insert
+		(GTK_MENU_SHELL (ui_lookup_widget(wb_globals.geany_plugin->geany_data->main_widgets->window, "menubar1")),
+		 menu_data.root_item, 8);
```
LarsGit223: Is it possible to query the position of the "Help" menu? Maybe I should try to get that position and decrement it.

workbench/src/plugin_main.c
```diff
@@ -0,0 +1,147 @@
...
+	/* Set functions */
+	plugin->funcs->init = plugin_workbench_init;
+	plugin->funcs->cleanup = plugin_workbench_cleanup;
+	plugin->funcs->help = plugin_workbench_help;
+	plugin->funcs->callbacks = plugin_workbench_callbacks;
+
+	/* Register! */
+	if (GEANY_PLUGIN_REGISTER(plugin, 225))
+	{
+		return;
```
LarsGit223: Ooops...embarrassing :blush: Fixed.

workbench/src/plugin_main.c
```diff
@@ -0,0 +1,147 @@
...
+
+
+/* Show help */
+static void plugin_workbench_help (GeanyPlugin *plugin, gpointer pdata)
+{
+	utils_open_browser("http://plugins.geany.org/workbench.html");
+}
+
+
+PluginCallback plugin_workbench_callbacks[] = {
```
LarsGit223: Sure, changed.

LarsGit223:
Enough for today. Will address the rest tomorrow.

workbench/icons/Makefile.am
codebrainz: Automake uses the variable name, so `icondir` tells where to put the files in variable `icon_DATA` (or in the case here `dist_icon_DATA` - not sure what the `dist_` adds here since the stuff in `_DATA` should be distributed anyway).

Probably the best way to install multiple sizes is to do [like Geany does](https://github.com/geany/geany/tree/master/icons) and put them into separate subdirs for each size, and in the Makefile.am in each subdir, do like you did here.

workbench/src/menu.c
codebrainz: > Are there any more detailed rules about this? E.g. "Open" has an ellipsis, "Settings" not.

https://developer.gnome.org/hig/stable/writing-style.html.en#ellipses

workbench/src/plugin_main.c
codebrainz: @LarsGit223 in your defense, it's [straight out of the plugin example](http://www.geany.org/manual/reference/howto.html) :)

workbench/src/plugin_main.c
codebrainz: Also [straight out of the manual](http://www.geany.org/manual/reference/pluginsignals_8c.html) :)

codebrainz:
TravisCI failure due to #596

workbench/src/menu.c
codebrainz: You could use `gtk_container_get_children()` and use the list's length to figure out how many items there are, and then stick yours at -2 from the end. You could also use `gtk_container_get_children()` and iterate through looking for the item with the name from the Glade file for the Help menu (not guaranteed stable, but unlikely to change), and do how you said. Geany has no really good way to do this.

workbench/src/dialogs.c
b4n: don't apologize, you can user whichever style you like, it's your code, it doesn't really have to follow Geany code style.  I merely mentioned that for consistency, it's absolutely not a blocker for me.

workbench/src/menu.c
elextr: These reasons are why plugins generally add to the "Tools" menu rather than adding to the top level.

workbench/src/plugin_main.c
b4n: Oops :)  We should probably improve the example suggesting why a conditional might be useful instead of a useless code path indeed

workbench/src/plugin_main.c
b4n: @codebrainz not exactly, the part you linked is the old plugin interface where the variable is looked up by Geany on plugin load, @LarsGit223's plugin uses the new interface which passes the callback through the plugin registration call.

workbench/icons/Makefile.am
b4n: Yeah, or use a single Makefile and the `nobase_` prefix, like:

```make
iconsdir = $(datadir)/icons/hicolor
nobase_dist_icons_DATA = \
	16x16/apps/workbench-project.png \
	32x32/apps/workbench-project.png \
	scalable/apps/workbench-project.svg \
	…
```

And of course have the different size in the corresponding folders.

> not sure what the `dist_` adds here since the stuff in `_DATA` should be distributed anyway

No, [`_DATA` is not distributed by default](https://www.gnu.org/software/automake/manual/automake.html#index-_005fDATA), I guess because it's often generated out of templates and the like.

> How are the icons selected by the application?

They are installed as part of the [icon theme](https://developer.gnome.org/gtk3/stable/GtkIconTheme.html#GtkIconTheme.description) so referencing them by name can automatically load them at the most appropriate size.

codebrainz: >> not sure what the dist_ adds here since the stuff in _DATA should be distributed anyway

> No, _DATA is not distributed by default, I guess because it's often generated out of templates and the like.

Makes sense, why would you want to distribute files that you've indicated need to be installed? &lt;/sarcasm&gt; 

In my experience it actually does distribute them, either that or it doesn't cause `make distcheck` to fail, or else I've always been very lucky and somehow got them distributed automatically.

elextr: > Makes sense, why would you want to distribute files that you've indicated need to be installed? </sarcasm>

Thats autofools for ya :stuck_out_tongue_winking_eye: , but it is documented behaviour for `_DATA`, installs but not dist unless its explicit.

workbench/src/plugin_main.c
LarsGit223: Anyway, the variable name is only referenced in my plugin so it's better static.

workbench/icons/Makefile.am
LarsGit223: Well, then I think I will go for the single makefile approach suggested by @b4n because it seems to be the simplest solution.

Another thing regarding icons:
I noticed that there is no 16x16 icon for javascript files. That means if I have a javascript file listed in a workbench project directory, then it is bigger than the other icons. I assumed there would be references to icon files for each extension in geany but I did not find it. Is this coming from the OS? So in my case Ubuntu/Gnome. Should I open an issue there about the missing 16x16 icon for javascript?

elextr: The icons are obtained from GTK, so they are not in Geany, [see](https://github.com/geany/geany/blob/adc22a453b546b05f507fd6522ea13f3a92943c2/src/ui_utils.c#L3017).  And IIUC GTK gets them from the theme these days.

My Geany shows a JS icon in the documents tab for js files, so maybe its your theme thats deficient.

workbench/src/dialogs.c
LarsGit223: Thanks, no big thing. I think if someone is committing code to an existing project it is just "good manners" to follow the coding guidelines. The core developers and maintainers of a project are the one's who most likely spent the most time on the project so I think at least the coding guidelines should be followed. If someone does not follow coding guidelines in a large way and by intention that feels a bit disrespectful to me. Also, you can read code from other people better if the style is the same among a project.

But I have to admit I would like to see projects which just say: "Use code prettifier 'xyz' with settings 'abc' and you are fine.".

workbench/src/menu.c
```diff
@@ -0,0 +1,310 @@
...
+					 G_CALLBACK(item_new_workbench_activate_cb), NULL);
+	geany_plugin_set_data(wb_globals.geany_plugin, menu_data.item_new, NULL);
+
+	/* Create new menu item "Open Workbench" */
+	menu_data.item_open = gtk_menu_item_new_with_mnemonic(_("Open"));
+	gtk_widget_show(menu_data.item_open);
+	gtk_menu_append(GTK_MENU (menu_data.menu), menu_data.item_open);
+	g_signal_connect(menu_data.item_open, "activate",
+					 G_CALLBACK(item_open_workbench_activate_cb), NULL);
+	geany_plugin_set_data(wb_globals.geany_plugin, menu_data.item_open, NULL);
```
LarsGit223: Also removed.

workbench/src/popup_menu.c
```diff
@@ -0,0 +1,475 @@
...
+
+
+/* Handle popup menu item "Remove directory" */
+static void popup_menu_on_remove_directory(G_GNUC_UNUSED GtkMenuItem * menuitem, G_GNUC_UNUSED gpointer user_data)
+{
+	SIDEBAR_CONTEXT context;
+
+	if (sidebar_file_view_get_selected_context(&context)
+		&&
+		context.project != NULL && context.directory != NULL)
```
LarsGit223: Adjusted. Logical operator and following conditions are now on one line.

workbench/src/popup_menu.c
```diff
@@ -0,0 +1,475 @@
...
+void popup_menu_init(void)
+{
+	GtkWidget *item;
+
+	s_popup_menu.widget = gtk_menu_new();
+
+	item = gtk_menu_item_new_with_mnemonic(_("Add project"));
+	gtk_widget_show(item);
+	gtk_container_add(GTK_CONTAINER(s_popup_menu.widget), item);
+	g_signal_connect((gpointer) item, "activate", G_CALLBACK(popup_menu_on_add_project), NULL);
```
LarsGit223: Is removed now (all occurrences).

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+		path_list = g_slist_prepend(path_list, path_split);
+	}
+
+	if (path_list != NULL)
+		sidebar_create_branch(0, abs_base_dir, path_list, parent);
+
+	g_slist_foreach(lst, (GFunc) g_free, NULL);
+	g_slist_free(lst);
+	g_slist_foreach(path_list, (GFunc) g_strfreev, NULL);
+	g_slist_free(path_list);
```
LarsGit223: Nice, thanks.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+
+		foreach_slist (elem, dirs)
+		{
+			gtk_tree_store_insert_with_values(sidebar.file_store, &iter, parent, *position,
+				FILEVIEW_COLUMN_ICON, icon,
+				FILEVIEW_COLUMN_NAME, wb_project_dir_get_name(elem->data),
+				FILEVIEW_COLUMN_DATA_ID, DATA_ID_DIRECTORY,
+				FILEVIEW_COLUMN_ASSIGNED_DATA_POINTER, elem->data,
+				-1);
+			position++;
```
LarsGit223: Fixed.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+	else
+	{
+		icon = g_icon_new_for_string("workbench-nodirs", NULL);
+
+		gtk_tree_store_insert_with_values(sidebar.file_store, &iter, parent, *position,
+			FILEVIEW_COLUMN_ICON, icon,
+			FILEVIEW_COLUMN_NAME, _("No directories"),
+			FILEVIEW_COLUMN_DATA_ID, DATA_ID_NO_DIRS,
+			-1);
+		position++;
```
LarsGit223: Fixed.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+	}
+	g_object_unref(icon);
```
LarsGit223: Done.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+		name = get_any_relative_path(wb_project_get_filename(project), file);
+		gtk_tree_store_insert_with_values(sidebar.file_store, &iter, parent, *position,
+			FILEVIEW_COLUMN_ICON, icon,
+			FILEVIEW_COLUMN_NAME, name,
+			FILEVIEW_COLUMN_DATA_ID, DATA_ID_PRJ_BOOKMARK,
+			FILEVIEW_COLUMN_ASSIGNED_DATA_POINTER, file,
+			-1);
+		(*position)++;
+	}
+	g_object_unref(icon);
```
LarsGit223: Done.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+static void sidebar_insert_all_projects(GtkTreeIter *iter, gint *position)
+{
+	GIcon *icon_ok, *icon_ko, *icon;
+	guint index, max;
+
+	if (wb_globals.opened_wb == NULL)
+		return;
+
+	icon_ok = g_icon_new_for_string("workbench-project", NULL);
+	icon_ko = g_icon_new_for_string("workbench-project-error", NULL);
```
LarsGit223: Adjusted: replaced them with ```icon_good``` and ```icon_bad```.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+		child_position = 0;
+		/* Not required here as we build a completely new tree
+		   sidebar_remove_children(&iter); */
+		sidebar_insert_project_bookmarks(project, iter, &child_position);
+		sidebar_insert_project_directories(project, iter, &child_position);
+	}
+	gtk_tree_view_expand_all(GTK_TREE_VIEW(sidebar.file_view));
+
+	g_object_unref(icon_ok);
+	g_object_unref(icon_ko);
```
LarsGit223: Already fixed.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+					dialogs_show_msgbox(GTK_MESSAGE_INFO, "%s", info);
+				}
+				else
+				{
+					dialogs_show_msgbox(GTK_MESSAGE_ERROR, _("%s\nProject file not found!"), info);
+				}
+				g_free(info);
+			break;
+			case DATA_ID_DIRECTORY:
+			    info = wb_project_dir_get_info((WB_PROJECT_DIR *)address);
```
LarsGit223: Fixed.

workbench/src/utils.c
```diff
@@ -0,0 +1,343 @@
...
+	gchar *basedir, *basedir_end;
+	const gchar *start;
+	gchar *result;
+	guint length;
+	gint goback;
+
+	if (relative[0] != '.')
+	{
+		/* Not a relative directory. Simply return it. */
+		return strdup(relative);
```
LarsGit223: Fixed.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+				popup_context = POPUP_CONTEXT_DIRECTORY;
+			}
+			if (context.folder != NULL)
+			{
+				popup_context = POPUP_CONTEXT_FOLDER;
+			}
+			if (context.file != NULL)
+			{
+				popup_context = POPUP_CONTEXT_FILE;
+			}
```
LarsGit223: You are right, "POPUP_CONTEXT_FILE" is the most specific situation and "POPUP_CONTEXT_WB_BOOKMARK" the least specific. Changed it.

workbench/src/wb_globals.c
```diff
@@ -0,0 +1,36 @@
...
+# include "config.h"
+#endif
+
+#include "wb_globals.h"
+
+struct S_WB_GLOBALS
+{
+    GeanyPlugin *geany_plugin;
+    WORKBENCH   *opened_wb;
+};
```
LarsGit223: Removed it.

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+ **/
+void wb_project_set_filename(WB_PROJECT *prj, gchar *filename)
+{
+	if (prj != NULL)
+	{
+		guint offset;
+		gchar *ext;
+
+		prj->filename = g_strdup(filename);
+		prj->name = g_path_get_basename (filename);
```
LarsGit223: Well, today the function is only called once. But I changed it to be on the safe side in case of future changes.

workbench/icons/Makefile.am
LarsGit223: I see an icon, just the wrong size. It seems to be missing in ```/usr/share/icons/elementary-xfce/mimes/16```. So it clearly looks like a theme problem.

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+}
+
+
+/** Get the filename of a project.
+ *
+ * @param prj The project
+ * @return The filename
+ *
+ **/
+gchar *wb_project_get_filename(WB_PROJECT *prj)
```
LarsGit223: Done.

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+}
+
+
+/** Get the name of a project.
+ *
+ * @param prj The project
+ * @return The project name
+ *
+ **/
+gchar *wb_project_get_name(WB_PROJECT *prj)
```
LarsGit223: Done.

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+	if (queue == NULL || *queue == NULL)
+	{
+		return;
+	}
+
+	foreach_slist(elem, *queue)
+	{
+		g_free(elem->data);
+	}
+	g_slist_free(*queue);
```
LarsGit223: Thanks, changed.

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+	GDir *dir;
+	gchar *locale_path = utils_get_locale_from_utf8(utf8_path);
+	gchar *real_path = tm_get_real_path(locale_path);
+
+	dir = g_dir_open(locale_path, 0, NULL);
+	if (!dir || !real_path || g_hash_table_lookup(visited_paths, real_path))
+	{
+		g_free(locale_path);
+		g_free(real_path);
+		return NULL;
```
LarsGit223: Fixed it.

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+/* Create a new project dir with base path "utf8_base_dir" */
+static WB_PROJECT_DIR *wb_project_dir_new(const gchar *utf8_base_dir)
+{
+	guint offset;
+
+	if (utf8_base_dir == NULL)
+	{
+	    return NULL;
+	}
+	WB_PROJECT_DIR *dir = (WB_PROJECT_DIR *) g_new0(WB_PROJECT_DIR, 1);
```
LarsGit223: Removed it.

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+	gchar *searchdir;
+
+	wb_project_dir_remove_from_tm_workspace(root);
+	g_hash_table_remove_all(root->file_table);
+
+	if (!root->file_patterns || !root->file_patterns[0])
+	{
+		gchar **all_pattern = g_strsplit ("*", " ", -1);
+		pattern_list = get_precompiled_patterns(all_pattern);
+		g_strfreev(all_pattern);
```
LarsGit223: None, except not having seen it. Simplified it as suggested.

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+		{
+			prj->modified = FALSE;
+		}
+		g_free (contents);
+	}
+	else if (error != NULL)
+	{
+		g_set_error (error, 0, 0,
+					 "Internal error: param missing (file: %s, line %d)",
+					 __FILE__, __LINE__);
```
LarsGit223: Done.

workbench/src/wb_project.c
LarsGit223: Done.

workbench/src/wb_project.c
LarsGit223: Done.

workbench/src/wb_project.c
LarsGit223: Yes! Replaced it with ```g_slist_free_full(prj->directories, (GDestroyNotify)wb_project_dir_free);```

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+}
+
+
+/** Get the name of a project dir.
+ *
+ * @param directory The project dir
+ * @return The name
+ *
+ **/
+gchar *wb_project_dir_get_name (WB_PROJECT_DIR *directory)
```
LarsGit223: Fixed.

workbench/src/workbench.c
LarsGit223: Done.

workbench/src/workbench.c
LarsGit223: Fixed. All filename parameters and return values in ```wb_project.c``` and ```workbench.c``` can be ```const```.

workbench/src/workbench.c
LarsGit223: Ok, fixed. Also, if ```filename``` is ```NULL``` the function now also returns with ```FALSE``` immediately.

workbench/src/dialogs.c
```diff
@@ -0,0 +1,411 @@
...
+	GtkDialogFlags flags;
+	gchar *file_patterns_old, *ignored_file_patterns_old, *ignored_dirs_patterns_old;
+	gboolean changed;
+
+	/* Create the widgets */
+	flags = GTK_DIALOG_MODAL | GTK_DIALOG_DESTROY_WITH_PARENT;
+	dialog = gtk_dialog_new_with_buttons (_("Directory settings"),
+										  GTK_WINDOW(wb_globals.geany_plugin->geany_data->main_widgets->window),
+										  flags,
+										  GTK_STOCK_CANCEL,
```
LarsGit223: Maybe I have found it in some other code and then started to use it without changing the previous code. But I removed it because it is deprecated. There doesn't seem to be a replacement.

LarsGit223: In which directions is Geany itself going about Gtk Stock Items? It's still using it. Are there plans to change it?

workbench/src/menu.c
```diff
@@ -0,0 +1,310 @@
...
+}
+
+
+/** Cleanup menu data/mem.
+ *
+ **/
+void menu_cleanup (void)
+{
+	gtk_container_remove (GTK_CONTAINER(ui_lookup_widget(wb_globals.geany_plugin->geany_data->main_widgets->window, "menubar1")),
+						  menu_data.root_item);
```
LarsGit223: I read the documentation. Yes should work. Changed it.

workbench/src/menu.c
```diff
@@ -0,0 +1,310 @@
...
+	}
+	wb_globals.opened_wb = workbench_new();
+	workbench_set_filename(wb_globals.opened_wb, filename);
+	if (workbench_save(wb_globals.opened_wb, &error))
+	{
+		menu_item_new_deactivate();
+		menu_item_open_deactivate();
+		menu_item_save_activate();
+		menu_item_settings_activate();
+		menu_item_close_activate();
```
LarsGit223: Yes. Just changed it to work in a similar way like the sidebar and the popup menu.

workbench/src/menu.c
```diff
@@ -0,0 +1,310 @@
...
+						  menu_data.root_item);
+}
+
+
+/** Activate menu item "New".
+ *
+ **/
+void menu_item_new_activate (void)
+{
+	gtk_widget_set_sensitive(menu_data.item_new, TRUE);
```
LarsGit223: Already solved, see re-factored version of ```menu.c```.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+	{
+		gint length;
+		gchar text[200];
+
+		length = g_snprintf(text, sizeof(text), "%s", wb_project_get_name(project));
+		if (length < (sizeof(text)-1) && wb_project_is_modified(project))
+		{
+			text [length] = '*';
+			text [length+1] = '\0';
+		}
```
LarsGit223: Done.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+		gtk_tree_store_set(sidebar.file_store, &iter,
+			FILEVIEW_COLUMN_NAME, text,
+			-1);
```
LarsGit223: True. Merged the functions together and added a parameter to choose to only update the project title or all parts of the project.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+		{
+			icon = icon_ko;
+		}
+
+		length = g_snprintf(text, sizeof(text), "%s", wb_project_get_name(project));
+		if (length < (sizeof(text)-1) && wb_project_is_modified(project))
+		{
+			text [length] = '*';
+			text [length+1] = '\0';
+		}
```
LarsGit223: See above.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+		gchar text[200];
+
+		count = workbench_get_project_count(wb_globals.opened_wb);
+		length = g_snprintf(text, sizeof(text), "%s: %u %s",
+							workbench_get_name(wb_globals.opened_wb), count, _("Projects"));
+		if (length < (sizeof(text)-1) && workbench_is_modified(wb_globals.opened_wb))
+		{
+			text [length] = '*';
+			text [length+1] = '\0';
+		}
```
LarsGit223: Done.

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+ *
+ * @param dir The project dir
+ * @return The info string
+ *
+ **/
+gchar *wb_project_dir_get_info (WB_PROJECT_DIR *dir)
+{
+	guint pos = 0;
+	gchar *str;
+	gchar text[2048];
```
LarsGit223: Switched to GString.

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1491 @@
...
+/** Get an info string for the project.
+ *
+ * @param prj The project
+ * @return The info string
+ *
+ **/
+gchar *wb_project_get_info (WB_PROJECT *prj)
+{
+	guint pos = 0;
+	gchar text[2048];
```
LarsGit223: Switched to GString.

workbench/src/wb_project.c
LarsGit223: Removed it.

workbench/src/workbench.c
LarsGit223: According to the documentation "g_key_file_to_data()" never returns an error. That's kind of weird. I check the result anyway now and so error will not be re-used.

LarsGit223:
Enough work for today. Sorry for all the messages/noise :wink:

workbench/src/dialogs.c
b4n: For now we continue using them (they also do provide icons for most things), but I guess we'll eventually stop when starting to support GTK4.  So for now the direction is conservative, not removing it before needed.

po/be.po
```diff
@@ -3405,7 +3405,7 @@ msgstr ""
 msgid "Toggle Insert/Overwrite mode"
```
b4n: There is no relevant changes in this file

po/ca.po
```diff
@@ -3344,7 +3344,7 @@ msgstr "Cancel·la la la Selecció"
 msgid "Toggle Insert/Overwrite mode"
```
b4n: Nor in this one (and probably quite a few translation files)

workbench/src/menu.c
```diff
@@ -0,0 +1,231 @@
...
+{
+	/* Create menu and root item/label */
+	menu_data.menu = gtk_menu_new();
+	menu_data.root_item = gtk_menu_item_new_with_label(_("Workbench"));
+	gtk_widget_show(menu_data.root_item);
+
+	/* Create new menu item "New Workbench" */
+	menu_data.item_new = gtk_menu_item_new_with_mnemonic(_("_New"));
+	gtk_widget_show(menu_data.item_new);
+	gtk_menu_append(GTK_MENU (menu_data.menu), menu_data.item_new);
```
b4n: [`gtk_menu_append()`](https://developer.gnome.org/gtk2/stable/GtkMenu.html#gtk-menu-append) is deprecated for a long time in favor to [`gtk_menu_shell_append()`](https://developer.gnome.org/gtk2/stable/GtkMenuShell.html#gtk-menu-shell-append), and its use [breaks compilation under GTK3](https://travis-ci.org/geany/geany-plugins/jobs/266322308#L4962) which removed the former.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1156 @@
...
+} ExpandData;
+
+typedef struct SIDEBAR
+{
+    GtkWidget *file_view_vbox;
+    GtkWidget *file_view;
+    GtkTreeStore *file_store;
+    GtkWidget *file_view_label;
+}SIDEBAR;
+static SIDEBAR sidebar = {NULL, NULL, NULL, NULL, NULL};
```
b4n: [Excess initializer](https://travis-ci.org/geany/geany-plugins/jobs/266322308#L4946) (there are 4 fields, and you initialize 5).

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1156 @@
...
+
+	gtk_tree_view_append_column(GTK_TREE_VIEW(sidebar.file_view), column);
+
+	gtk_tree_view_set_headers_visible(GTK_TREE_VIEW(sidebar.file_view), FALSE);
+	gtk_tree_view_set_enable_search(GTK_TREE_VIEW(sidebar.file_view), TRUE);
+	gtk_tree_view_set_search_column(GTK_TREE_VIEW(sidebar.file_view), FILEVIEW_COLUMN_NAME);
+
+	pfd = pango_font_description_from_string(wb_globals.geany_plugin->geany_data->interface_prefs->tagbar_font);
+	gtk_widget_modify_font(sidebar.file_view, pfd);
+	pango_font_description_free(pfd);
```
b4n: you could use `ui_widget_modify_font_from_string()` from Geany's utility function (which also would avoid the [GTK3 deprecation warning](https://travis-ci.org/geany/geany-plugins/jobs/266322308#L4954))

workbench/src/menu.c
LarsGit223: Ok, I will follow the convention. Now it is attached to the tools menu.

po/be.po
LarsGit223: Yes, I tried to restore the not changed translation files to the original state but that did not work. The only really changed file is ```po/de.po```.

Maybe I should place a new fresh PR is all points are sorted out to keep the other translation files unchanged?

workbench/src/menu.c
LarsGit223: Thanks, replaced it.

workbench/src/sidebar.c
LarsGit223: Fixed.

workbench/src/sidebar.c
LarsGit223: Thanks, done.

workbench/src/workbench.c
LarsGit223: Yes, done.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1164 @@
...
+		sidebar_deactivate();
+	}
+	else
+	{
+		gint length;
+		gchar text[200];
+
+		count = workbench_get_project_count(wb_globals.opened_wb);
+		length = g_snprintf(text, sizeof(text), "%s: %u %s",
+							workbench_get_name(wb_globals.opened_wb), count, _("Projects"));
```
LarsGit223: Is corrected.

workbench/icons/Makefile.am
LarsGit223: I had a look again and changed my opinion: the directories in geany look clean and simple so I tried starting to do it that way, @codebrainz: thanks for the link.

But I can't get it to work. In ```workbench/icons/Makefile.am``` I have put:
```
SUBDIRS = 16x16 
#24x24 32x32 48x48 scalable tango

gtk_update_icon_cache = gtk-update-icon-cache -f -t

install-data-hook:
	@-if test -z "$(DESTDIR)"; then					\
		echo "Updating Gtk icon cache.";			\
		$(gtk_update_icon_cache) "$(datadir)/icons/hicolor";	\
		$(gtk_update_icon_cache) "$(datadir)/icons/Tango"; \
	else								\
		echo "*** Icon cache not updated. Remember to run:";	\
		echo "***";						\
		echo "***   $(gtk_update_icon_cache) '$(datadir)/icons/hicolor'";\
		echo "***   $(gtk_update_icon_cache) '$(datadir)/icons/Tango'";\
		echo "***";						\
	fi

uninstall-local:
	rm -f $(DESTDIR)$(datadir)/icons/hicolor/icon-theme.cache
	rm -f $(DESTDIR)$(datadir)/icons/Tango/icon-theme.cache
```
Right now I started with the original icons only, so only subdir ```16x16``` is used. In ```workbench/icons/16x16/Makefile.am``` I have put:
```
iconsdir = $(datadir)/icons/hicolor/16x16
icons_appsdir = $(iconsdir)/apps

dist_icons_apps_DATA =							\
	workbench-bookmark.png						\
	workbench-dir.png							\
	workbench-nodirs.png						\
	workbench-project.png						\
	workbench-project-error.png
```
But when I call ```make``` I get the following error:
```
Making all in src
make[1]: Entering directory '/home/lars/DokuWiki/repos/gp-branch-workbench2/geany-plugins/workbench/src'
make[1]: Nothing to be done for 'all'.
make[1]: Leaving directory '/home/lars/DokuWiki/repos/gp-branch-workbench2/geany-plugins/workbench/src'
Making all in icons
make[1]: Entering directory '/home/lars/DokuWiki/repos/gp-branch-workbench2/geany-plugins/workbench/icons'
 cd ../.. && /bin/bash /home/lars/DokuWiki/repos/gp-branch-workbench2/geany-plugins/missing automake-1.15 --foreign workbench/icons/Makefile
 cd ../.. && /bin/bash ./config.status workbench/icons/Makefile 
config.status: creating workbench/icons/Makefile
Making all in 16x16
make[2]: Entering directory '/home/lars/DokuWiki/repos/gp-branch-workbench2/geany-plugins/workbench/icons/16x16'
make[2]: *** No rule to make target 'all'.  Stop.
make[2]: Leaving directory '/home/lars/DokuWiki/repos/gp-branch-workbench2/geany-plugins/workbench/icons/16x16'
Makefile:509: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/home/lars/DokuWiki/repos/gp-branch-workbench2/geany-plugins/workbench/icons'
Makefile:578: recipe for target 'all-recursive' failed
make: *** [all-recursive] Error 1
```
So everything looks fine until the point where directory ```16x16``` is entered. I know make but not automake and also not the details of the geany makefile. I do not see any target in the original geany 16x16/Makefile.am so I wonder why it's not working.

I need some help here please.

LarsGit223: (I will now be AFK for about an hour)

workbench/src/menu.c
LarsGit223: @codebrainz: thanks. I added ellipses to "New", "Open", "Add project" and "Add directory".

workbench/src/popup_menu.c
```diff
@@ -0,0 +1,475 @@
...
+/** Setup/Initialize the popup menu.
+ *
+ **/
+void popup_menu_init(void)
+{
+	GtkWidget *item;
+
+	s_popup_menu.widget = gtk_menu_new();
+
+	item = gtk_menu_item_new_with_mnemonic(_("Add project"));
```
LarsGit223: Done.

workbench/src/utils.c
```diff
@@ -0,0 +1,343 @@
...
+ **/
+gchar *get_combined_path(const gchar *base, const gchar *relative)
+{
+	gchar *basedir, *basedir_end;
+	const gchar *start;
+	gchar *result;
+	guint length;
+	gint goback;
+
+	if (relative[0] != '.')
```
LarsGit223: I re-factored the function. It is now handling all relative pathes and uses ```g_strconcat```.

workbench/src/utils.c
```diff
@@ -0,0 +1,343 @@
...
+		else
+		{
+			break;
+		}
+		goback--;
+	}
+
+	length = strlen(basedir)+strlen(start);
+	result = g_new(char, length+1);
+	if (result == NULL)
```
LarsGit223: See above.

workbench/src/utils.c
```diff
@@ -0,0 +1,343 @@
...
+
+
+/** Get the relative path.
+ *
+ * The function examines the relative path of @a target in relation to
+ * @a base. It is not required that target is a sub-directory of base.
+ *
+ * The function can only properly handle pathes which consist of a maximum of
+ * 29 sub-directories. If the pathes exceed this maximum the function aborts
+ * and NULL is returned.
```
LarsGit223: Re-factored the function, now uses GPtrArray. Also fixed the memory leaks.

workbench/src/utils.c
```diff
@@ -0,0 +1,343 @@
...
+		else
+		{
+			break;
+		}
+		index++;
+	}
+
+	pos = 0;
+	length = 0;
+	relative = g_new (char *, 30);
```
LarsGit223: Fixed, see above.

build/workbench.m4
```diff
@@ -0,0 +1,10 @@
+AC_DEFUN([GP_CHECK_WORKBENCH],
+[
+    GP_ARG_DISABLE([Workbench], [auto])
+    GP_COMMIT_PLUGIN_STATUS([Workbench])
+    AC_CONFIG_FILES([
+        workbench/Makefile
+        workbench/src/Makefile
+        workbench/icons/Makefile
```
codebrainz: Put the `16x16` Makefile here also, it gets the Makefile processed by Autoconf.

workbench/icons/Makefile.am
codebrainz: I think it's what I mentioned [here](https://github.com/geany/geany-plugins/pull/598/files#r134122552).

build/workbench.m4
LarsGit223: Yes, that was the missing line. THANKS.

workbench/icons/Makefile.am
LarsGit223: Yes, thanks. Works now. I added icons for 24x24, 32x32, 48x48 and scalable. The scalable icons are only symbolic which means they are painted in one color only (gray).

Now I got the problem the the icons for the first two levels (the project and the directories added to it) are much bigger than the icons for the folders and files contained below the directories. How can I control the size?

po/be.po
b4n: > Yes, I tried to restore the not changed translation files to the original state but that did not work.

It might be a little tricky because the files were updated in master and you merged that, so you can't simply revert.  I restored the version from master in https://github.com/b4n/geany-plugins/tree/LarsGit223/workbench%2Brestored_trans (commit add14f7117926b93ae02e390f1a079863203d370).

> Maybe I should place a new fresh PR is all points are sorted out to keep the other translation files unchanged?

As you like, I guess ideally you'd rebase your changes on top of master and force-push to this PR once it's all sorted out.

workbench/icons/Makefile.am
codebrainz: I'm not sure I understand. If you mean the icons when loaded in to Geany, I believe the functions that load icons allow to specify the size (ex. `GTK_ICON_SIZE_MENU`, etc).

LarsGit223: See this screenshot:

workbenchscreenshoticons

I would like to have all icons the same size. I do not really specify a size anywhere.

The icons that are small in that screenshot are created with ```g_content_type_get_icon()```. There is no size in that function call. Shortly below that call there is this call: ```gtk_icon_theme_lookup_by_gicon(gtk_icon_theme_get_default(), icon, 16, 0);```. But as I understood this should not change the ```icon```.

The icons that are (too) large in the screenshot have simply been created with ```g_icon_new_for_string()```.

All icons were of small size before I committed the change for supplying different icon size sets. So how do I choose the size? I tried ```g_object_set()``` on the ```GtkTreeView```'s cell renderer (```g_object_set(renderer, "stock-size", 16, NULL);```, the cell renderer is a pixbuf created with ```gtk_cell_renderer_pixbuf_new()```).

Well, enough for today. Maybe I write an mail to the Gtk-App-Devel mailing list tomorrow. Bye for now.

workbench/icons/24x24/Makefile.am
```diff
@@ -0,0 +1,9 @@
+iconsdir = $(datadir)/icons/hicolor/16x16
```
b4n: wrong target directory, should be `24x24`

workbench/icons/32x32/Makefile.am
```diff
@@ -0,0 +1,9 @@
+iconsdir = $(datadir)/icons/hicolor/16x16
```
b4n: should be `32x32`

workbench/icons/48x48/Makefile.am
```diff
@@ -0,0 +1,9 @@
+iconsdir = $(datadir)/icons/hicolor/16x16
```
b4n: Should be `48x48`

workbench/icons/scalable/Makefile.am
```diff
@@ -0,0 +1,9 @@
+iconsdir = $(datadir)/icons/hicolor/16x16
```
b4n: should be `scalable`

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1154 @@
...
+	gtk_tree_view_set_model(GTK_TREE_VIEW(sidebar.file_view), GTK_TREE_MODEL(sidebar.file_store));
+
+	renderer = gtk_cell_renderer_pixbuf_new();
+	column = gtk_tree_view_column_new();
+	gtk_tree_view_column_pack_start(column, renderer, FALSE);
+	gtk_tree_view_column_add_attribute(column, renderer, "gicon", FILEVIEW_COLUMN_ICON);
+
+	renderer = gtk_cell_renderer_text_new();
+	gtk_tree_view_column_pack_start(column, renderer, TRUE);
+	gtk_tree_view_column_add_attribute(column, renderer, "markup", FILEVIEW_COLUMN_NAME);
```
b4n: you should use `"text"`" instead of `"markup"` as you pass raw filenames (and don't seem to need the markup features), otherwise it'll break with filenames containing `&`s, `<`s and the like.

workbench/src/wb_project.c
```diff
@@ -0,0 +1,1465 @@
...
+{
+	GSList *list = NULL;
+	GDir *dir;
+	gchar *locale_path = utils_get_locale_from_utf8(utf8_path);
+	gchar *real_path = tm_get_real_path(locale_path);
+
+	dir = g_dir_open(locale_path, 0, NULL);
+	if (!dir || !real_path || g_hash_table_lookup(visited_paths, real_path))
+	{
+		g_dir_close(dir);
```
b4n: you should guard this call against `NULL` `dir`

workbench/icons/Makefile.am
b4n: This is because you install all your icons in the `16x16` dir ;)

b4n:
Also, some interesting messages from my compiler on mean settings:
```
  CC       workbench_la-wb_project.lo
../../../workbench/src/wb_project.c: In function 'wb_project_dir_rescan_int':
../../../workbench/src/wb_project.c:606:2: warning: conversion to 'gint' from 'guint' may change the sign of the result [-Wsign-conversion]
  return filenum;
  ^
../../../workbench/src/wb_project.c: In function 'wb_project_dir_rescan':
../../../workbench/src/wb_project.c:701:12: warning: conversion to 'guint' from 'gint' may change the sign of the result [-Wsign-conversion]
  filenum = wb_project_dir_rescan_int(prj, root);
            ^
../../../workbench/src/wb_project.c:707:2: warning: conversion to 'gint' from 'guint' may change the sign of the result [-Wsign-conversion]
  return filenum;
  ^
  CC       workbench_la-sidebar.lo
../../../workbench/src/sidebar.c: In function 'sidebar_insert_workbench_bookmarks':
../../../workbench/src/sidebar.c:506:59: warning: unused parameter 'workbench' [-Wunused-parameter]
 static void sidebar_insert_workbench_bookmarks(WORKBENCH *workbench, GtkTreeIter *iter, gint *position)
                                                           ^
../../../workbench/src/sidebar.c: In function 'sidebar_reset_tree_store':
../../../workbench/src/sidebar.c:549:51: warning: unused parameter 'iter' [-Wunused-parameter]
 static void sidebar_reset_tree_store(GtkTreeIter *iter)
                                                   ^
```

workbench/icons/Makefile.am
LarsGit223: Ahrgh...stupid copy&paste fault. Thanks.

LarsGit223:
@b4n: which settings did you use? Where is the best place to change the compiler flags in the makefile?

elextr:
@b4n has an insanely long list of compiler settings (referred to in the project as `--b4n-pedantic`), but sane people accept `-Wall -Wextra` as a compromise, see `HACKING`.

Don't change the makefile, set it permanently with configure.

```
./configure CFLAGS=whatever
```

or temporarily to make

```
CFLAGS=whatever make
```

workbench/icons/Makefile.am
LarsGit223: Fixed.

workbench/src/sidebar.c
LarsGit223: Changed it.

workbench/src/wb_project.c
LarsGit223: Done.

LarsGit223:
@b4n, @elextr: Thanks, fixed the warnings.

@b4n: could you give me your CFLAGS list anyway please? It should be fixed but I did not get the warning about the sign change on the return values with using ```-Wall -Wextra``` only.

elextr:
>  not get the warning about the sign change on the return values with using -Wall -Wextra only.

Well, its perfectly legal C to assign `int` to `uint`, and there are old time system interfaces that use `int` where they only return positive values and actually should have used `uint`, so the warning is spurious (and unfixable) in some cases.  Thats why it is not included in `-Wall` and `-Wextra`.

The less used warnings tend to be ones that can cause more unsuppressable spurious results, or just plain stupid results (see a recent discussion between @b4n and I about returning small structs from functions that was triggered by one of his warnings :wink: ).  Thats why @b4n doesn't advise his option set and use it on Travis, the results cannot be taken at face value, they need interpretation.

LarsGit223:
> Thats why @b4n doesn't advise his option set and use it on Travis, the results cannot be taken at face value, they need interpretation.

Well, true for every warning I guess. It's always good to think about what you are doing :smile: In this case it looked like a inconsistency that was easy to fix. If it might not be fixable in future cases, then also fine. But first I need the option to be able to check it and then act on it accordingly. I do not want to propagate it as a default value or something, just for me.

codebrainz:
> But first I need the option to be able to check it and then act on it accordingly. 

According to [an old mailing list post](http://lists.geany.org/pipermail/devel/2013-September/008087.html):

```
-Wall -W -O2 -Wunused -Wunreachable-code -Wformat=2 -Wundef -Wshadow
-Wpointer-arith -Wbad-function-cast -Wwrite-strings -Wconversion
-Waggregate-return -Wstrict-prototypes -Wmissing-prototypes
-Wmissing-declarations -Wmissing-noreturn -Wmissing-format-attribute
-Wredundant-decls -Wnested-externs -pedantic
-Werror-implicit-function-declaration -fshow-column
```

LarsGit223:
@codebrainz: thanks, just compiled with that list. The warnings that I get can be ignored. I saved the list to my wiki so I can't lose it.

LarsGit223:
I close this one now, see follow-up PR #601.

PR #601, opened 2017-09-02
New plugin 'workbench': initial commit. (Follow-up/clean PR)

LarsGit223:
This is a new plugin called workbench.

It has the following features:

- assign multiple projects to a workbench
- assign directories and file patterns to each project for easy file listing
- create "global" workbench or project bookmarks to have quick access to often used files
- the plugin uses relative pathes (a project file path is relative to the workbench file path, a file path is relative to the project file path)
- languages: english and german

For instructions please see the readme file.

The plugin is heavily based on the project organizer plugin, thanks to Jiří Techet. Icons are taken from the gnome desktop.

**This is a follow-up/clean PR for PR #598. All changes are included, this PR is based on todays master.**


LarsGit223:
Just as a follow-up: I consider this PR as being in a good condition. If there are any more change-requests please let me know.

workbench/src/sidebar.c
```diff
@@ -0,0 +1,1152 @@
...
+	GSList *dir_list = NULL;
+	GSList *file_list = NULL;
+	GSList *elem;
+	gchar **path_arr;
+
+	foreach_slist (elem, leaf_list)
+	{
+		if (elem->data == NULL)
+		{
+			continue;
```
frlan: Is this ever passing this point?

LarsGit223: If I remeber right, yes. There was an error sceanario where I had a crash if NULL was not catched. I think it happened if a branch is empty because of not matching file patterns.

frlan:
Did you check it with make check?

LarsGit223:
Yes, I just did again because I do not remeber when I last checked it. No messages, seems to be OK (Only saw messages for plugin **xmlsnippets** but that's not my building site).

One question:
What exactly is make check doing? Do I need to install additional tools to get the full usage/checks?

codebrainz:
`make check` is for running tests, which I don't think GP has any of. @frlan probably meant `make distcheck` which checks that build system is OK (ie. no missing files, etc).

frlan:
It needs cppcheck

frlan:
Please check for at least 
sidebar.c:101: error: Uninitialized variable: elem
sidebar.c:249: error: Uninitialized variable: elem
sidebar.c:282: error: Uninitialized variable: elem
utils.c:76: error: Uninitialized variable: elem
wb_project.c:541: error: Uninitialized variable: elem
wb_project.c:585: error: Uninitialized variable: elem
wb_project.c:731: error: Uninitialized variable: elem
wb_project.c:782: error: Uninitialized variable: elem
wb_project.c:804: error: Uninitialized variable: elem2
wb_project.c:809: error: Uninitialized variable: elem
wb_project.c:862: error: Uninitialized variable: elem2
wb_project.c:867: error: Uninitialized variable: elem

LarsGit223:
I checked that before, they are fine. That are all calls to ```foreach_slist```, e.g. 
```
static guint wb_project_get_file_count(WB_PROJECT *prj)
{
	GSList *elem;
	guint filenum = 0;

	foreach_slist(elem, prj->directories)
	{
		filenum += ((WB_PROJECT_DIR *)elem->data)->file_count;
	}
	return filenum;
}
```
But ```elem``` gets assigned the elements from ```prj->directories```, so everything is OK. Looks like ```ccpcheck``` is having a problem here with ```foreach_slist```.

codebrainz:
You could probably add suppressions to `AM_CPPCHECKFLAGS`, initialize those variables to `NULL`, or just use a proper `for` loop like:

```c
for (GSList *elem = prj->directories; elem; elem = elem->next)
```

Which is only 11 extra characters, is much more readable/less magic, and won't need to be changed if those weird `foreach` macros are ever deprecated.

Issue #604, opened 2017-09-27
projectorganizer: Too many open files


kugel-:
Since a recent ubuntu update I get this when starting Geany.

`(geany:3522): GLib-ERROR **: Creating pipes for GWakeup: Too many open files`

If I disable git-changebar, it does start just fine. 

@b4n any idea?

kugel-:
@b4n Sorry, git-changebar was an innocent victim.

@techee projectorganizer keeps file descriptors open to lots of directores (my project base directory has many subdiretories). One something tries to open one more file descriptor everything falls down. Why does projectorganizer keep so many fds?

The limit seems to be 1024 on my system, for processes started by my local non-root user. projectorganizer uses more than 1000 of them.

elextr:
@kugel- check that you don't have a loop in your filesystem, even a soft link will make it go in a loop.

elextr:
@techee I'm thinking [this line](https://github.com/geany/geany-plugins/blob/59118d86ef8853943dec74c9f6d6fad9a535931a/projectorganizer/src/prjorg-project.c#L103) should check for soft links, or is it intended to follow them?  If it is intended to follow links then it should check the realpath of the directory has not been seen before to avoid loops.

kugel-:
I wonder why projectorganizer keeps FDs to all that dirs/files open.

elextr:
As it descends the directory tree it opens each directory and as it reads the entries recursively opens each one thats a directory.  So as it opens each child directory, each parent directory is still open since it hasn't finished reading the entries in it yet.  So if you get a loop in the filesystem it will simply run out of FDs.

kugel-:
In the non-crashing case, after Geany finished starting, the file descriptors are still open.

I don't see loops when examining /proc/$(pidof geany)/fd

kugel-:
Perhaps the limit is hit during traversal, and the code doesn't clean up properly?

techee:
I have to check. This shouldn't happen, there's no need to keep the fd's open for the directories - probably something isn't closed properly.

techee:
@elextr The plugin detects symlink loops and avoids them. It stores all visited paths to a hash table and if the paths start repeating, it stops the recursion.

elextr:
@techee but the constructed path gets longer and longer so it won't compare equal, you need to compare realpath()s.  Look at what happens in bash if you have the wd in the prompt, as you cd via a symlink the prompt gets longer and longer :)

techee:
[...]

PR #605, opened 2017-09-27
projectorganizer: Close dir created with g_dir_open() in some special cases

techee:
At the moment the directory isn't closed when tm_get_real_path() returns
NULL or when the given path was already visited (happens e.g. with
symlink loops).

### DIFF
--- a/projectorganizer/src/prjorg-project.c
+++ b/projectorganizer/src/prjorg-project.c
@@ -74,6 +74,9 @@ static GSList *get_file_list(const gchar *utf8_path, GSList *patterns,
 {
 	GSList *list = NULL;
 	GDir *dir;
+	const gchar *child_name;
+	GSList *child;
+	GSList *children = NULL;
 	gchar *locale_path = utils_get_locale_from_utf8(utf8_path);
 	gchar *real_path = tm_get_real_path(locale_path);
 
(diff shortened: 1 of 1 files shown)


techee:
@kugel- Would you test this patch? I hope it could fix #604 - otherwise I don't know. Do you use some symlinks that could cause that the same path is re-visited many times? Or something that causes that tm_get_real_path() returns NULL for many files? These are the only two cases in which the directory wasn't properly closed (for normal projects I haven't been able to reproduce the issue - no dirs were open).

Note that if you had 1024 nested directories like

```
first/second/third/.../1024-th
```

the current code will still need to open all the intermediate directories for traversal and you'll get the issue during project load. This would be rather hard to fix (or more precisely, the traversal would have to be slow because the intermediate dirs would have to be re-opened closed all the time) but this is a rather crazy case so I'll leave it as it is.

techee:
Disregard my previous comment about the deeply nested directories - it's easy to fix and I've added a patch that does that so now there's at most 1 directory open during traversal.

Anyway, the second patch is just a bonus - it's not something that could cause that many fd's are open after the project is loaded and not the cause of the original problem.

b4n:
LGTM.  First commit makes a lot of sense and fixes and actual possible issue.  Second seems a little more theoretical and slightly slower in the regular case, but fair enough anyway.

elextr:
LGTM

@b4n, I would have said the second would be faster, reading everything while its all cached, but anyway it hardly matters in a function doing so much IO.

techee:
And to have all opinions covered, I think the real-world speed will be identical :-).

kugel-:
Your changes seem to work. Thank you. I didn't test the first fix individually, just the HEAD of your branch.

techee:
@kugel- Good, thanks for testing.

@frlan OK to merge?

PR #608, opened 2017-10-02
Workbench: fixed not closed file handles on directory scanning (thanks to techee). Fixes #607.

### LINKED ISSUE #607: Workbench: open file handles not closed on directory scanning
@LarsGit223 the Workbench has the same bug as fixed for ProjectOrganizer in #605 (it looks like some of the ProjectOrganizer code was copied).

Interestingly, the change in 99dcc2344ebf23d40d5426035d0ae950b197cf14 is already in the Workbench plugin but the main change needs to be done here as well.

LarsGit223:
Nothing more to say.

### DIFF
--- a/workbench/src/wb_project.c
+++ b/workbench/src/wb_project.c
@@ -228,6 +228,9 @@ static GSList *wb_project_dir_get_file_list(WB_PROJECT_DIR *root, const gchar *u
 {
 	GSList *list = NULL;
 	GDir *dir;
+	const gchar *child_name;
+	GSList *child;
+	GSList *children = NULL;
 	gchar *locale_path = utils_get_locale_from_utf8(utf8_path);
 	gchar *real_path = utils_get_real_path(locale_path);
 
@@ -245,14 +248,17 @@ static GSList *wb_project_dir_get_file_list(WB_PROJECT_DIR *root, const gchar *u
 
 	g_hash_table_insert(visited_paths, real_path, GINT_TO_POINTER(1));
 
-	while (TRUE)
+	while ((child_name = g_dir_read_name(dir)))
+	children = g_slist_prepend(children, g_strdup(child_name));
+
+	g_dir_close(dir);
+
+	foreach_slist(child, children)
 	{
 		const gchar *locale_name;
 		gchar *locale_filename, *utf8_filename, *utf8_name;
 
-		locale_name = g_dir_read_name(dir);
-		if (!locale_name)
-			break;
+		locale_name = child->data;
 
 		utf8_name = utils_get_utf8_from_locale(locale_name);
 		locale_filename = g_build_filename(locale_path, locale_name, NULL);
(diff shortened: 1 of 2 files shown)


workbench/src/wb_project.c
```diff
@@ -245,14 +248,17 @@ static GSList *wb_project_dir_get_file_list(WB_PROJECT_DIR *root, const gchar *u
 
 	g_hash_table_insert(visited_paths, real_path, GINT_TO_POINTER(1));
 
-	while (TRUE)
+	while ((child_name = g_dir_read_name(dir)))
```
frlan: I think you should make this more explicit here

workbench/src/wb_project.c
```diff
@@ -228,6 +228,9 @@ static GSList *wb_project_dir_get_file_list(WB_PROJECT_DIR *root, const gchar *u
 {
 	GSList *list = NULL;
 	GDir *dir;
+	const gchar *child_name;
+	GSList *child;
```
frlan: = NULL;

workbench/src/wb_project.c
LarsGit223: What do you mean? To write out the code like this:
``` c
	child_name = g_dir_read_name(dir);
	while (child_name)
	{
		children = g_slist_prepend(children, g_strdup(child_name));
		child_name = g_dir_read_name(dir);
	}
```
?

b4n: I don't think repeating the call is a good idea.  I don't mind the original code myself, but some people prefer things like this: `while ((child_name = g_dir_read_name(dir)) != NULL)` so it's more obvious the assignation is not a typo.

frlan: At least I needed to think twice what those lines do. I'm not a 100%-C-geek but I don't think we need to use every syntax candy available.

LarsGit223: Honestly I simply copied techee's code. I also don't like assignment's in while loops that much but I can live with it. I suggest the following code:
``` c
	while ((child_name = g_dir_read_name(dir)) != NULL)
	{
		children = g_slist_prepend(children, g_strdup(child_name));
	}
```
So we have the explicit ```NULL``` pointer check and the curly braces to make it more readable. If that fine for you two then I will apply that change.

codebrainz: The original code (with or without the explicit `NULL` check) is pretty much the way I've always seen `readdir` being used. The extra parenthesis aren't required but they make compilers shut up about mistaking `==` for `=`.

frlan: @codebrainz  So you are suggesting to merge it?

codebrainz: Just that the part of the code in question seems reasonable.

LarsGit223: I changed the code to the last suggested code piece. Please merge.

PR #609, opened 2017-10-02
projectorganizer: Replace deprecated tm_get_real_path().

LarsGit223:
Nothing more to say.


techee:
@LarsGit223 The function was introduced in API 235 so the required API has to be updated too in the plugin.

LarsGit223:
Ok, thanks. Will do it for geanylua, projectorganizer and workbench soon (not today).

PR #610, opened 2017-10-02
geanylua: Replace deprecated tm_get_real_path().

LarsGit223:
Nothing more to say.


techee:
@LarsGit223 Needs updating API version too.

PR #613, opened 2017-10-04
workbench: Increased required geany API version to 235 for usage of 'utils_get_real_path'.

LarsGit223:
Nothing more to say.


eht16:
I should have done this change in the first place in #606. Sorry for that and thanks for doing my work :).

LarsGit223:
Never mind, s... happens. At least I learned something :wink:

PR #614, opened 2017-10-04
addons/markword: Deselect when the selection changes

gkatev:
Not sure if desired or not, but the marked results currently do not clear when the selected text is modified/deleted. The modification detects the deletion and clears the markers. Only works when single-click deselect is enabled.

### DIFF
--- a/addons/src/ao_markword.c
+++ b/addons/src/ao_markword.c
@@ -182,6 +182,17 @@ void ao_mark_editor_notify(AoMarkWord *mw, GeanyEditor *editor, SCNotification *
 		
 		if(priv->enable_markword && priv->enable_single_click_deselect)
 			clear_marker();
+	} 
+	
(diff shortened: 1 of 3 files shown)


addons/src/ao_markword.c
```diff
@@ -190,6 +208,14 @@ void ao_mark_document_open(AoMarkWord *mw, GeanyDocument *document)
 		FALSE,
 		G_CALLBACK(on_editor_button_press_event),
 		mw);
+		
+	plugin_signal_connect(
```
eht16: This is the wrong place for the signal connect.
Here we connect to the signal on each opened document but never disconnect from it.

It would be better to re-use the already existing signal callback `ao_editor_notify_cb` in addons.c and add there a call to `on_editor_notify` after making it public.

PR #620, opened 2017-10-08
Added common utils lib

LarsGit223:
This PR is a "visibility" PR to let people know that I am working on a common utils lib for geany-plugins.
Right now this only includes a function for scanning a directory for files.

PR #622, opened 2017-10-11
Added common utils lib (2).

LarsGit223:
The lib is placed in sub-directory 'utils'.
So far it only includes utility functions for scanning a directory for files.

This is a fresh PR only including the changes for the common lib in folder utils. It does not include changes on other plugins. This replaces #620.


eht16:
I like the idea of this library and besides the mentioned additional use
for "relative bundle path code used on Windows and Mac", we probably can
move some of the plugin configuration config file handling in each plugin
here. There is a lot of duplicated code around.

Since we add a new dependency for other plugins here (currently just Workbench but probably
more in the future), we should emphasize this in the NEWS to help distro packagers when releasing. They probably need to add something like geany-plugins-common package and make all other plugin sub packages depend on this (as it is already on Debian).

Apropos, in order to be able to compile it on Windows I had to add:
`-no-undefined` to `libgeanypluginutils_la_LDFLAGS`.
Without, linking fails with:
`can't build shared library unless -no-undefined is specified`.
I'm not sure whether we should add it there or better in
`build/vars-build.mk` where it is already specified for `AM_LDFLAGS`.

codebrainz:
I also mentioned some weirdness about `AM_LDFLAGS` in [the PR](https://github.com/LarsGit223/geany-plugins/pull/1) I made for [the other/first](https://github.com/geany/geany-plugins/pull/620) PR for this, but it's mysteriously missing now. I think it was that it cannot be used by anything but plugins since it puts libtool `-module -avoid-version` flags in there. That is why [the Makefile.am explicitly uses `GP_LDFLAGS`](https://github.com/geany/geany-plugins/pull/622/commits/1f6923e012aa26fd95fc3f3cabd66e2b7d32ee20#diff-00b6bde048b921fd76923b9d7416829cR13) instead of implicitly inheriting the `AM_LDFLAGS`.

Also, I left [a TODO](1f6923e012aa26fd95fc3f3cabd66e2b7d32ee20#diff-4bf485625e498fde51df9a8ce1925c08R4) in the .m4 template that might be nice to fix.

PR #628, opened 2017-10-15
workbench: use commonlib.

LarsGit223:
This replaces the call to 'wb_project_dir_get_file_list' with a call to 'filelist_scan_directory' from the common utils lib. Therefore the functions 'wb_project_dir_get_file_list', 'get_precompiled_patterns' and 'patterns_match' could be deleted.


LarsGit223:
Attention:
This requires PR #622 to be merged first/also!

PR #631, opened 2017-10-16
Improve/fix some things related to the utils library

b4n:
Most importantly IMO, add a prefix (C "namespace") to functions from that lib.  Let's not do the same mistake as Geany there.

2 additional things that could be nice:
1. install header files so it can be used by 3rd parties as well
2. build some developer documentation (I kind of hate Doxygen for its total lack of automated main page, but we could use that nonetheless like Geany does)


elextr:
But your namespace starts with `g`, you know Gnome/GTK claims ALL `g` namespaces ;-P

codebrainz:
Overall looks good to me. Not tested.

codebrainz:
> 1. install header files so it can be used by 3rd parties as well

I was going to do that, but then I thought that no 3rd party plugin is going to depend on all of Geany-Plugins just to get some helper functions. In any case, it's super easy to install the header(s), as you know.

elextr:
> install header files so it can be used by 3rd parties as well

Don't plugin writers count as third parties?  Or is it required that plugin devs (at least C/C++ ones) must install a git version of Geany?

LarsGit223:
> build some developer documentation (I kind of hate Doxygen for its total lack of automated main page, but we could use that nonetheless like Geany does)

I assume the _"Geany Plugin API Documentation"_ is part of Geany and not of Geany Plugins, right? So we need to start a completely new main page/doxygen structure?

frlan:
Currently I don't think we should offer this outside the g-p project. There are a bunch of plugins not even known to me having their own little helper functions. We might could think about migrating some of the functions later to Geany's plugin interface

Issue #313, opened 2015-11-28
Rename GeanyLaTeX to get rid of Geany-prefix


frlan:
It's a geany-plugins. There don't need to be any Geany-prefix on name.

LarsGit223:
I think this can be closed.

codebrainz:
Closed by #632

PR #632, opened 2017-10-16
geanylatex: renamed to latex

LarsGit223:
I changed all occurrences of ```geanylatex``` to ```latex```.

In ```geany-plugins.nsi``` I kept ```geanylatex``` and added ```latex```. So both DLL's should be removed on un-install. I hope the un-installer does not throw an error if a file to-be-deleted does not exist.

ToDo's:
- change pdf documentation
- change screenshots
- eventually adjust config handling to read in config for ```geanylatex``` and ```latex```?

**One question:**
How can an old/renamed lib be un-installed under linux? Is it possible/does the build system support it?


eht16:
> In ```geany-plugins.nsi``` I kept ```geanylatex``` and added ```latex```. So both DLL's should be removed on un-install. I hope the un-installer does not throw an error if a file to-be-deleted does not exist.

Yep, it's fine this way. NSIS handle deleting non-existing files
gracefully which is what we want here.

codebrainz:
> How can an old/renamed lib be un-installed under linux? Is it possible/does the build system support it?

I don't think anyone bothered last time plugins were renamed but I think you could do it by using a custom `uninstall-local` target. @b4n or @sardemff7 or someone probably has a better idea how.

elextr:
[This](https://github.com/LarsGit223/geany-plugins/blob/299176db7ceb987c09ce4c63b4e7c7aea774a593/latex/src/latex.c#L197) changes the directory where the config is stored.  How do you get the users existing configuration copied over?

frlan:
I started to this some tim eago https://github.com/frlan/geany-plugins/tree/feature/rename_geanylatex

LarsGit223:
@elextr: It doesn't. That's why I mentioned it on the ToDo's. Can we boldly copy over the settings to a new file? Or rename the file if we see it under the old name?

@frlan: sorry, didn't know that. How far are you? If you want to send over a PR yourself than I can close this?

elextr:
> Can we boldly copy over the settings to a new file? Or rename the file if we see it under the old name?

I don't know, thats why I asked the question :)

The experts will have to say if the installation can be made to copy the settings file (including on windows) or if the plugin itself will have to do it and if rename or copy is needed (especially on widnows).

LarsGit223:
Well, I thought copying in the source code. I think there is a standard C remove command. But we would also need a command to remove the old dir and create a new one. Not sure if there is a standard C lib call for it across Linux, Windows and Mac.

LarsGit223:
Forget my question, found the GIO functions. That should do it.

LarsGit223:
Closes #313.

latex/doc/latex.tex
```diff
@@ -53,7 +53,7 @@
 
 \dedication{\normalsize \textbf{Note:} Please note that this document has been created on
 \today. If you are using a devel version from GIT, please compile and check
-\texttt{doc/geanylatex.tex} from sources. Please check Page \pageref
```
frlan: I woul'd rather keep the name of old version ins ChangeLog as they are.

LarsGit223: Maybe I am misunderstanding something but in that line it is only the name of the latex documentation. Not the name printed inside the log. Should I only change that line or a different one?

frlan: At this very point the plugin was called GeanyLATeX so I'm voting for not changing that part of documentation, which is pretty much the ChangeLog.

frlan: Oh. My fault. 
I pointed to the wrong line. Sorry.

LarsGit223: Ähemm..ok. Can you please point me to the right line? Or just send an PR.

frlan: Completely my failure. Sorry. Nevermind.

PR #634, opened 2017-10-18
workbench: import Geany project data

LarsGit223:
The workbench plugin now imports the "base_path" and "file_patterns" from Geany's project settings
as an workbench project directory. That directory is then treated as the project's base dir.

In the config file all settings related to the project's base dir are prefixed with "Prj-"
instead of e.g. "Dir1-". On loading of a workbench's project, the plugin checks if there
is a "base_path" in the "project" section but no "Prj-BaseDir" in the "Workbench" section.
In that case the data from the Geany project is imported. That means, the import happens only once.
On import, the project is marked as changed (a '*' is displayed after the project's name).

The project's base dir is always labeled "Base dir" and has a specific icon to show that it differ's
from the other directories. Also, it is always listed first/on top of the tree.


workbench/icons/48x48/Makefile.am
```diff
@@ -2,6 +2,7 @@ iconsdir = $(datadir)/icons/hicolor/48x48
 icons_appsdir = $(iconsdir)/apps
 
 dist_icons_apps_DATA =							\
+	workbench-basedir.png						\
```
frlan: Do we really need to ship these? Aren't there any good stock ones? Just wondering

LarsGit223: Well, I just searched through my installed gnome icons and picked the ones I liked. I did not do any research. Do you now any link to a comfortable gallery of stock icons? I only found C code/a list of defines/names. But no gallery???

LarsGit223: I found one at http://www.pygtk.org/pygtk2reference/gtk-stock-items.html. But I a, not sure if that is a complete reference.

frlan:
I was thinking about something like https://www.geany.org/manual/gtk/gtk/gtk-Stock-Items.html

frlan:
There is some upstream Gtk-docu also but this one was nearer ;)

LarsGit223:
Thanks. Some of the icons are actually icons I used but not all. Maybe I am going to change them later but not in this PR. Suggestions are always welcome of course.

workbench/icons/48x48/Makefile.am
codebrainz: You could also just use the user's icon theme instead of foisting your aesthetics on them :) This icon would be named `user-home`, although I'm not sure that's semantically correct either (it's meant for user's home directory just like this icon).

codebrainz: See: https://developer.gnome.org/icon-naming-spec/

frlan:
Please use stock items when possible instead of shipping own icons. Stock icons are part of the theme and fit to user's environment, shipped ones might not.

elextr:
> shipped ones might not

See for example the fallen over bookcase icon in the geany toolbar :)

codebrainz:
> Please use stock items when possible instead of shipping own icons. Stock icons are part of the theme and fit to user's environment, shipped ones might not.

`s/stock items/theme icons/`

Stock items are deprecated (sadly).

LarsGit223:
@all:
Could we merge this anyway? I would like to have this change separated from icon changes.

Later if the discussion is sorted out I can make a PR for the icon changes.

@codebrainz:
> Stock items are deprecated (sadly).

Well, isn't the concept of named icons quite the same? A set of different functions and common names instead of #defined names but gtk still ships the icons and makes them usuable by a common name.

codebrainz:
> Well, isn't the concept of named icons quite the same? A set of different functions and common names instead of #defined names but gtk still ships the icons and makes them usuable by a common name.

For icons it's similar, except I don't believe there are any actual fallback icons built-in to GTK+ anymore. Stock items also provided standard translated labels (and tooltips?) for each item and a cleaner/simpler API.

PR #639, opened 2017-10-29
Connect button-press signal handlers on plugin activation

### LINKED ISSUE #638: Addons: mark all not working
@eht16 When enabled "Mark all occurances of a word when double clicking it" no longer works.  (Fairly recent Git Geany and G-P GTK-3).  There have been various discussions and pull requests around the operation of finds mark all functionality IIRC, did something there break addons?

eht16:
If the plugin gets activated *after* Geany has been started already,
we need to iterate all open documents and attach the "button-press"
signal handler.

Fixes #638.

### DIFF
--- a/addons/src/ao_markword.c
+++ b/addons/src/ao_markword.c
@@ -82,7 +82,7 @@ static void ao_mark_word_set_property(GObject *object, guint prop_id,
 			/* if the plugin is loaded while Geany is already running, we need to connect the
 			 * button press signal for open documents, if Geany is just booting,
 			 * it happens automatically */
-			if (main_is_realized())
+			if (priv->enable_markword && main_is_realized())
 			{
 				connect_documents_button_press_signal_handler(AO_MARKWORD(object));
 			}
--- a/addons/src/ao_markword.c
+++ b/addons/src/ao_markword.c
@@ -65,6 +65,7 @@ enum
 
 
 static void ao_mark_word_finalize  			(GObject *object);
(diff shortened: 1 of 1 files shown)


frlan:
Not sure wether it's layer 8 issue but when deactivating the plugin again, it's still marking

addons/src/ao_markword.c
```diff
@@ -78,6 +79,13 @@ static void ao_mark_word_set_property(GObject *object, guint prop_id,
 	{
 		case PROP_ENABLE_MARKWORD:
 			priv->enable_markword = g_value_get_boolean(value);
+			/* if the plugin is loaded while Geany is already running, we need to connect the
+			 * button press signal for open documents, if Geany is just booting,
+			 * it happens automatically */
+			if (main_is_realized())
```
b4n: shouldn't this check for `priv->enable_markword` too, not to do that on deactivation as well?

eht16: Uhhh, of course, thanks. Stupid me. Fixed in 71e26d8f69092ddaa2da305bb3cee4f76378a8b5.

eht16:
@frlan details on reproducing? Even without 71e26d8f69092ddaa2da305bb3cee4f76378a8b5 it doesn't mark anything after deactivation.

frlan:
Hmmm. Weird. I'm pretty sure I tested it with a clean version. Will retest it with latest HEAD of that branch later and give feedback. Maybe just another layer 8 issue on my end :(

frlan:
Ah... Was an issue with automark-plugin.

PR #641, opened 2017-10-30
treebrowser: Use consistent capitalization in menu

medovina:
Before this change, the treebrowser used an inconsistent mix of
capitalized and uncapitalized words in the context menu.  This
changes them to all be capitalized for consistency with other
GTK applications.

### DIFF
--- a/treebrowser/src/treebrowser.c
+++ b/treebrowser/src/treebrowser.c
@@ -1206,15 +1206,15 @@ create_popup_menu(const gchar *name, const gchar *uri)
 	gboolean is_dir 		= is_exists ? g_file_test(uri, G_FILE_TEST_IS_DIR) : FALSE;
 	gboolean is_document 	= document_find_by_filename(uri) != NULL ? TRUE : FALSE;
 
-	item = ui_image_menu_item_new(GTK_STOCK_GO_UP, _("Go _up"));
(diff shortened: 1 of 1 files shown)


elextr:
Maybe you should have reviewed [this](https://developer.gnome.org/hig/stable/writing-style.html.en) first.

medovina:
@elextr, I actually did look at that page while making this change, and generally followed the GNOME guideliines for header capitalization.

I chose to make only one difference from those guidelines: I used the capitalization "Create New File" rather than "Create new File".  That's because "Create new File" looks strange to me since "new" is an important word in the menu item.

Is there some other text in this change that you are concerned about?

elextr:
The `New`  was the one I happened to notice :)  And for me the lower case looks better but YMMV, if thats the only case then it doesn't matter.

medovina:
Actually, looking at this again I think I'd rather change those menu items like this:

Create New Directory -> New Folder
Create New File -> New File

This is shorter and would follow the capitalization guidelines, and is also more consistent with other GTK programs (e.g. Nautilus, gedit).  OK with you?

elextr:
Yeah, `folder` seems to be more common than `directory`, and shorter is (mostly) better.

PR #644, opened 2017-11-05
Fix cppcheck warnings

eht16:
Is 26fd7328cd409ceb16557a0fe637d5c028c0598f enough for properly enabling the test program on `make check`? To me it seems, it was just missing or was it intentional that the test program isn't run on `make check`?

### DIFF
--- a/xmlsnippets/src/tests.c
+++ b/xmlsnippets/src/tests.c
@@ -80,7 +80,7 @@ static gboolean test(gint ordinal, const gchar *input, const gchar *completion_n
 	else if (!completion_needed && !success)
 		return TRUE;
 
-	if (strcmp(c.completion, completion_needed) != 0)
+	if (g_strcmp0(c.completion, completion_needed) != 0)
 	{
 		g_warning(
 			"Test %d: FAIL\n"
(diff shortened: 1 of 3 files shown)


xmlsnippets/src/tests.c
```diff
@@ -80,7 +80,9 @@ static gboolean test(gint ordinal, const gchar *input, const gchar *completion_n
 	else if (!completion_needed && !success)
 		return TRUE;
 
-	if (strcmp(c.completion, completion_needed) != 0)
+	if ((!completion_needed && c.completion) ||
+		(completion_needed && !c.completion) ||
+		strcmp(c.completion, completion_needed) != 0)
```
codebrainz: Could also use [`g_strcmp0`](https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strcmp0).

eht16: Of course, did it and re-pushed.

eht16:
If there is nothing else, I'd like to merge this soon as the nightly builds are broken for days due to cppcheck issues.

codebrainz:
LGBI

PR #670, opened 2018-01-09
GeanyLua: Fix right-trim.lua

Skif-off:
Now script works with files with CRLF too (and with CR only). ```right-trim.lua ``` is presented as example (_geanylua/**examples**/edit/right-trim.lua_), but very useful (IMHO :)).


Skif-off:
Maybe it's a trifle and to nothing?

PR #674, opened 2018-01-14
SpellCheck: Support Enchant 2.0 and later

eht16:
Enchant 2.0 dropped some deprecated API and especially introduced
setting custom config/dictionary paths via an environment variable.
So, handle the API differences between Enchant 1.3, 1.5 and 2.0 and
document the changed custom dictionary behavior with 2.0 and later.


eht16:
Tested against Enchant 2.1 and 1.6.

frlan:
I did only a short check with 1.6, but looks okish for me

PR #682, opened 2018-01-24
treebrowser: fallback to "text-x-generic" icon if info lookup fails.

### LINKED ISSUE #657: [treebrowser] no more icon for html files
Since a recent update, running geany 1.30.1, the panel doesn't show any icon for html files : 

treebrowser

Tested with different icons sets on ubuntu 16.04 on unity. Strangely the documents panel show these icons.

LarsGit223:
If 'gtk_icon_theme_lookup_by_gicon()' returns NULL then try again with icon "text-x-generic".
Fixes #657.


brunob:
@b4n i'm not using geany from source anymore but from http://ppa.launchpad.net/geany-dev/ppa/ubuntu ; how can i test this ?

PR #683, opened 2018-01-29
Prettyprint usage

sagarchalise:
Curretly PrettyprintXML can be used files only,

this pr changes that to

-  selection based prettty print if selection is valid XML
- Donot change doctype to XML if HTML
- use of plugin API functions when taking buffer text

P.S. Sorry about all the merge commits.


codebrainz:
> P.S. Sorry about all the merge commits.

Why not just squash them?

PR #686, opened 2018-01-31
Add github names to maintainers file

elextr:
Have the maintainers github user name recorded somewhere so they can be pinged by users rather than waiting for them to notice an issue for their plugin was posted.

I have filled in the ones I know for sure, to everybody else, no offense, I am working from memory.


frlan:
@elextr  I will add the missing ones later

PR #689, opened 2018-02-05
Translation: Fix a typo

Skif-off:
You can see diff.


frlan:
Well... You don't need to update all po files. However, will merge it anyway. Thanks.

Skif-off:
@frlan 
Sorry, I thought it would be easier. (I forgot that "POT-Creation-Date" fields are updated automatically.)

PR #697, opened 2018-02-23
scope: port to GTK3.

LarsGit223:
I ported the scope plugin to GTK3 and VTE 2.91.

### DIFF
--- a/build/scope.m4
+++ b/build/scope.m4
@@ -1,7 +1,7 @@
 AC_DEFUN([GP_CHECK_SCOPE],
 [
     GP_ARG_DISABLE([Scope], [auto])
-    GP_CHECK_PLUGIN_GTK2_ONLY([Scope])
+    GP_CHECK_PLUGIN_GTK3_ONLY([Scope])
 
     case "$host_os" in
         cygwin* | mingw* | win32*)
@@ -10,7 +10,7 @@ AC_DEFUN([GP_CHECK_SCOPE],
 
         *)
             GP_CHECK_PLUGIN_DEPS([scope], [VTE],
-                                 [vte >= 0.17])
+                                 [vte-2.91])
             AC_CHECK_HEADERS([util.h pty.h libutil.h])
             PTY_LIBS="-lutil"
             ;;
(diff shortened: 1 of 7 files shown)


Skif-off:
@LarsGit223 
Why ```GP_CHECK_PLUGIN_GTK3_ONLY```? Is plugin don't support build with GTK2 now?

LarsGit223:
@Skif-off: no. As the PR is right now, it would only support GTK3.

Skif-off:
@LarsGit223 
Sorry, I didn't understand :) Is Geany drop to support GTK2?

LarsGit223:
@Skif-off:
No. By default geany compiles/is built for GTK2. If you specify the option ```--enable-gtk3``` on build then it will be build for GTK3. Then only plugins would work with that version of geany which support GTK3.

Which OS/distro do you use? Many linux distros support both GTK versions.

LarsGit223:
To be explicit: Geany will still support GTK2. It will not be dropped.

Skif-off:
@LarsGit223 
Thanks, I was confused by the [GP_CHECK_PLUGIN_GTK**3_ONLY**](https://github.com/geany/geany-plugins/pull/697/files#diff-c6ec9ad09f6351d19420270032720e7eR4).

P.S. I use Xubuntu 17.10, but I compile on my own (I use [PR 1752](https://github.com/geany/geany/pull/1752)) and [I don't like GTK3](https://github.com/geany/geany/issues/1746).

elextr:
@Skif-off To be even more explicit, geany will still support gtk2, its just the changes in this PR make this plugin gtk3 only.  Which is why I keep repeating myself that it should be a separate plugin so it leaves the GTK2 version available for those lucky people where it does not hang Geany.

@LarsGit223 I don't suppose by any chance that you have removed the recursion that I understand is the trigger for the Glib bug?

LarsGit223:
@elextr: no, the changes are only regarding GTK3 and VTE compatibility.

LarsGit223:
@frlan: please don't merge this yet as there is still an open discussion how to proceed regarding GTK2 and GTK3 support.

frlan:
I will not merge it when disabling Gtk2 without suitable replacement for Gtk2 users

LarsGit223:
I am closing this ion favour of PR #750.

PR #699, opened 2018-02-24
addons: use new plugin API

LarsGit223:
Switched to new plugin API:
- moved all plugin API code to the end of file
- therefore also some function declarations became needless and could be removed
- added help function


eht16:
LGBI

PR #703, opened 2018-02-24
commander: use new plugin API.

LarsGit223:
Nothing more to say.

### DIFF
--- a/commander/src/commander-plugin.c
+++ b/commander/src/commander-plugin.c
@@ -32,18 +32,8 @@
 /*#define DISPLAY_SCORE 1*/
 
 
-GeanyPlugin      *geany_plugin;
-GeanyData        *geany_data;
-
-PLUGIN_VERSION_CHECK(226)
-
-PLUGIN_SET_TRANSLATABLE_INFO (
-  LOCALEDIR, GETTEXT_PACKAGE,
-  _("Commander"),
-  _("Provides a command panel for quick access to actions, files and more"),
-  VERSION,
-  "Colomban Wendling <ban@herbesfolles.org>"
-)
+static GeanyPlugin      *geany_plugin = NULL;
+static GeanyData        *geany_data = NULL;
 
 
 /* GTK compatibility functions/macros */
@@ -762,11 +752,15 @@ on_plugin_idle_init (gpointer dummy)
   return FALSE;
 }
 
-void
-plugin_init (GeanyData *data)
+
+static gboolean
+plugin_commander_init (GeanyPlugin *plugin, G_GNUC_UNUSED gpointer pdata)
 {
   GeanyKeyGroup *group;
-  
+
+  geany_plugin = plugin;
+  geany_data = plugin->geany_data;
+
   group = plugin_set_key_group (geany_plugin, "commander", KB_COUNT, NULL);
   keybindings_set_item_full (group, KB_SHOW_PANEL, 0, 0, "show_panel",
                              _("Show Command Panel"), NULL,
@@ -783,10 +777,12 @@ plugin_init (GeanyData *data)
   /* delay for other plugins to have a chance to load before, so we will
    * include their items */
   plugin_idle_add (geany_plugin, on_plugin_idle_init, NULL);
+  return TRUE;
 }
 
-void
-plugin_cleanup (void)
+
+static void
+plugin_commander_cleanup (G_GNUC_UNUSED GeanyPlugin *plugin, G_GNUC_UNUSED gpointer pdata)
 {
   if (plugin_data.panel) {
     gtk_widget_destroy (plugin_data.panel);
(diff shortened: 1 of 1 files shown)


b4n:
@LarsGit223 just one question: why?

commander/src/commander-plugin.c
```diff
@@ -32,18 +32,8 @@
...
-
-PLUGIN_SET_TRANSLATABLE_INFO (
-  LOCALEDIR, GETTEXT_PACKAGE,
-  _("Commander"),
-  _("Provides a command panel for quick access to actions, files and more"),
-  VERSION,
-  "Colomban Wendling <ban@herbesfolles.org>"
-)
+static GeanyPlugin      *geany_plugin = NULL;
+static GeanyData        *geany_data = NULL;
```
b4n: These globals should be removed then, unless it's *highly* impractical

commander/src/commander-plugin.c
```diff
@@ -762,11 +752,15 @@ on_plugin_idle_init (gpointer dummy)
   return FALSE;
 }
 
-void
-plugin_init (GeanyData *data)
+
+static gboolean
+plugin_commander_init (GeanyPlugin *plugin, G_GNUC_UNUSED gpointer pdata)
```
b4n: style: each argument on its own line

commander/src/commander-plugin.c
```diff
@@ -783,10 +777,12 @@ plugin_init (GeanyData *data)
   /* delay for other plugins to have a chance to load before, so we will
    * include their items */
   plugin_idle_add (geany_plugin, on_plugin_idle_init, NULL);
+  return TRUE;
```
b4n: style: blank line above return

commander/src/commander-plugin.c
```diff
@@ -783,10 +777,12 @@ plugin_init (GeanyData *data)
...
 }
 
-void
-plugin_cleanup (void)
+
+static void
+plugin_commander_cleanup (G_GNUC_UNUSED GeanyPlugin *plugin, G_GNUC_UNUSED gpointer pdata)
```
b4n: style: each argument on its own line

commander/src/commander-plugin.c
```diff
@@ -796,8 +792,32 @@ plugin_cleanup (void)
   }
 }
 
-void
-plugin_help (void)
+
+static void
+plugin_commander_help (G_GNUC_UNUSED GeanyPlugin *plugin, G_GNUC_UNUSED gpointer pdat)
```
b4n: style: each argument on its own line

commander/src/commander-plugin.c
```diff
@@ -796,8 +792,32 @@ plugin_cleanup (void)
...
+{
+  utils_open_browser("https://plugins.geany.org/commander.html");
```
b4n: style: space before the open parenthesis

b4n: Why changing the target URI?  Why open a website when the same documentation is available locally?

commander/src/commander-plugin.c
```diff
@@ -762,11 +752,15 @@ on_plugin_idle_init (gpointer dummy)
   return FALSE;
 }
 
-void
-plugin_init (GeanyData *data)
+
```
b4n: style: only one blank line

commander/src/commander-plugin.c
```diff
@@ -796,8 +792,32 @@ plugin_cleanup (void)
...
+  utils_open_browser("https://plugins.geany.org/commander.html");
+}
+
+
+/* Load module */
+G_MODULE_EXPORT
+void geany_load_module(GeanyPlugin *plugin)
 {
-  utils_open_browser (DOCDIR "/" PLUGIN "/README");
+	/* Setup translation */
```
b4n: style: indentation should be 2 spaces

commander/src/commander-plugin.c
```diff
@@ -796,8 +792,32 @@ plugin_cleanup (void)
...
+	main_locale_init(LOCALEDIR, GETTEXT_PACKAGE);
```
b4n: style: space before open parenthesis

commander/src/commander-plugin.c
```diff
@@ -796,8 +792,32 @@ plugin_cleanup (void)
...
+	plugin->info->description = _("Provides a command panel for quick access to actions, files and more");
+	plugin->info->version = VERSION;
+	plugin->info->author = "Colomban Wendling <ban@herbesfolles.org>";
+
+	/* Set functions */
+	plugin->funcs->init = plugin_commander_init;
+	plugin->funcs->cleanup = plugin_commander_cleanup;
+	plugin->funcs->help = plugin_commander_help;
+
+	/* Register! */
```
b4n: this comment should be removed, as it adds no value and is not very formal

commander/src/commander-plugin.c
```diff
@@ -796,8 +792,32 @@ plugin_cleanup (void)
...
+	GEANY_PLUGIN_REGISTER(plugin, 226);
```
b4n: style: space before open parenthesis (OK, it was wrong for `PLUGIN_VERSOIN_CHECK()` before, but well)

LarsGit223:
> And there's at least one unrelated change (help location), which should at least be a separate commit.

Ok, I did not realize that it is opening the local file. Will remove it from this PR.

> I fail to see what value these changes give, given that the "old" API is fully functional (we spent some effort to make sure it'd still work flawlessly), and the new one, while nicer, doesn't provide any actual new feature this plugin would need. This looks like a change for the sake of change, which I don't really fancy (one reason being a bug likeliness to improvement ration very low). Also, if this gets done, it should fully embrace the new API philosophy and try and get rid of the global variables (unless actually problematic).

Hmmm....well it is not a change for the user in terms of functionality of course. But IMHO it is not just a change for the sake of change but a change for consistency and cleanup. Also, there was some reason to design the new API or not? The new API seems to be more easily extendable. This does not matter right now but maybe in the future if a API change/extension is required. I thought it would be nice to have all plugins using the new API and throw away the code which is required to support the old one.

And maybe I should not say "new API" but current API as it was introduced in Geany 1.26 (November 15, 2015).

Anyway, I should maybe have opened a discussion before starting this one. So I say sorry for my "bold" PR flood. This is your plugin, if you rather like to keep it like it is then I accept it of course.

commander/src/commander-plugin.c
LarsGit223: How? Move them into ```struct plugin_data``` or try to pass them as user pointers, e.g. from ```plugin_commander_init()``` to ```on_plugin_idle_init (-->create_panel...)```.

PR #716, opened 2018-03-01
workbench: pass error=NULL to "g_key_file_get_string_list()" for key "Bookmarks"

### LINKED ISSUE #714: workbench: GLib WARNING: GError set over the top of a previous GError or uninitialized memory.
The workbench plugin causes a needless GLib warning on opening a workbench file:
```GLib WARNING: GError set over the top of a previous GError or uninitialized memory.```

LarsGit223:
The key "Bookmarks" can be missing and that is not an error. So pass "NULL" for
parameter "error" on calling ```g_key_file_get_string_list()```. This prevents multiple
error messages from being written to one ```GErrror``` and so causing a GLib warning in
the debug messages. Fixes #714.

### DIFF
--- a/workbench/src/wb_project.c
+++ b/workbench/src/wb_project.c
@@ -1355,7 +1355,7 @@ gboolean wb_project_load(WB_PROJECT *prj, gchar *filename, GError **error)
 		gchar **bookmarks_strings;
 
 		/* Load project bookmarks from string list */
-		bookmarks_strings = g_key_file_get_string_list (kf, "Workbench", "Bookmarks", NULL, error);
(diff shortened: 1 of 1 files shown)


codebrainz:
You don't have to create an issue for each PR if you don't want to. It's nice to let people know you're working on something or to discuss, but for small stuff like this which is uncontroversial and/or you have already fixed, you could just submit the PR by itself. Either way is fine.

LarsGit223:
I know, thanks. But I rate this and the other little PR as a bug. And I really like to have it as a labeled issue so I can find things easier in the future.

LarsGit223:
@frlan: could you please label this with ```workbench``` and ```bug```?

PR #718, opened 2018-03-04
workbench: make "live update"-feature depending on GIO and Glib version 2.46

LarsGit223:
This hopefully fixes some Travis CI problems.

If GIO is not available and the GLib version is older than 2.46 then the feature "Live Update" will be disabled. Everything else works normal. In the "Settings"-Dialog the option do enable/disable the feature is grayed out and the tooltip text for the option informs the user that the feature is not available on the system.

### DIFF
--- a/build/workbench.m4
+++ b/build/workbench.m4
@@ -2,6 +2,12 @@ AC_DEFUN([GP_CHECK_WORKBENCH],
 [
     GP_ARG_DISABLE([Workbench], [auto])
     GP_CHECK_UTILSLIB([Workbench])
+
+    PKG_CHECK_MODULES([GIO], [gio-2.0],
+        [AC_DEFINE([HAVE_GIO], 1, [Whether we have GIO])
+         have_gio=yes],
+        [have_gio=no])
+
     GP_COMMIT_PLUGIN_STATUS([Workbench])
     AC_CONFIG_FILES([
         workbench/Makefile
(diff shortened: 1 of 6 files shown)


codebrainz:
Why does it require such a high version of GIO? I think the file monitoring stuff was added in 2.18, wasn't it?

LarsGit223:
I am doing directory monitoring. For that I also use the events ```G_FILE_MONITOR_EVENT_RENAMED```, ```G_FILE_MONITOR_EVENT_MOVED_IN``` and ```G_FILE_MONITOR_EVENT_MOVED_OUT```. And for that the GNOME documentation says version 2.46 is required, see https://developer.gnome.org/gio/stable/GFileMonitor.html (under section ```enum GFileMonitorEvent```).

codebrainz:
I see. You could probably still use monitoring with older versions and just `#if`-guard out those flags.

elextr:
One problem here, this tests on the build machine, not the machine the plugin is installed on when its distributed.

elextr:
> One problem here, this tests on the build machine, not the machine the plugin is installed on when its distributed.

Follow up.

According to our Debian packager the plugin package is tied to a Glib version anyway, so its not going to cause problems for normal users who get it from the repositories.

And if a user is smart enough to build the plugins themselves they are smart enough to handle their Glib version.

So never mind a runtime check.

LarsGit223:
@codebrainz:
> I see. You could probably still use monitoring with older versions and just #if-guard out those flags.

I checked the GLib history and version 2.46 seems to exist since 2015. So if no one is really requiring it I will only support the new flags for a plugin which is as new/young as the workbench plugin.

@elextr:
> So never mind a runtime check.

I will also add a runtime check soon.

LarsGit223:
Ok, please forget about GLib version 2.46. It turned out I do not need that events. See PR #719.

PR #719, opened 2018-03-05
workbench: use simple/old file monitor events only

LarsGit223:
This removes the dependency/the requirement for GLib version 2.46.

### DIFF
--- a/workbench/src/wb_monitor.h
+++ b/workbench/src/wb_monitor.h
@@ -22,18 +22,12 @@
 #include <glib.h>
 #include "wb_project.h"
 
-#if defined(HAVE_GIO)
-#define __WB_LIVE_UPDATE 1
-#endif
-
 typedef struct S_WB_MONITOR WB_MONITOR;
 
-#ifdef __WB_LIVE_UPDATE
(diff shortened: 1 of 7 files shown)


workbench/src/wb_monitor.h
```diff
@@ -22,7 +22,7 @@
 #include <glib.h>
 #include "wb_project.h"
 
-#if defined(HAVE_GIO) && GLIB_CHECK_VERSION (2, 46, 0)
+#if defined(HAVE_GIO)
```
codebrainz: I don't think you need this `HAVE_GIO` stuff from the build system as GIO is part of GLib, Geany requires GIO already, and AFAIK some plugins too.

elextr: Yes, Geany itself stopped checking back in [2011](https://github.com/geany/geany/commit/591417697010200441af6b6ed76ef28ef4731d83)

LarsGit223: Yes, geany itself is not using it, I will remove it. Maybe it should be removed from other plugins also.

LarsGit223: Done.

PR #720, opened 2018-03-07
treebrowser: removed HAVE_GIO checks

LarsGit223:
Removed obsolete GIO check.

### DIFF
--- a/treebrowser/src/treebrowser.c
+++ b/treebrowser/src/treebrowser.c
@@ -19,9 +19,7 @@
 #include "geany.h"
 #include "geanyplugin.h"
 
-#ifdef HAVE_GIO
-# include <gio/gio.h>
-#endif
+#include <gio/gio.h>
 
 #ifdef G_OS_WIN32
 # include <windows.h>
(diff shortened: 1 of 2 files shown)


treebrowser/src/treebrowser.c
```diff
@@ -208,7 +206,7 @@ utils_pixbuf_from_stock(const gchar *stock_id)
 static GdkPixbuf *
 utils_pixbuf_from_path(gchar *path)
 {
-#if defined(HAVE_GIO) && GTK_CHECK_VERSION(2, 14, 0)
+#if GTK_CHECK_VERSION(2, 14, 0)
```
b4n: The GTK check is also mostly irrelevant given how old GTK 2.24 is, and that Geany is depending on 2.24 for a long time now.  But that's outside of the scope of this PR.

elextr:
> FWIW, this check was more or less always obsolete, GIO is in GLib, and if GTK exposes API using its types it ought to be present…

But `libgio` is a separate library, maybe that was really what should have been tested? Sure on Linux its in the glib package so it will allways be there, but maybe not on Windows or OSX (at least before msys).

codebrainz:
It's pointless for plugins to check for GIO since GTK+ and Geany both use and require the library.

PR #723, opened 2018-03-11
treebrowser: removed needless Gtk version checks

LarsGit223:
Geany depends on Gtk version 2.24 so all version checks for a lower version than 2.24 are needless.

### DIFF
--- a/treebrowser/src/treebrowser.c
+++ b/treebrowser/src/treebrowser.c
@@ -138,11 +138,7 @@ PLUGIN_SET_TRANSLATABLE_INFO(
 static GList*
 _gtk_cell_layout_get_cells(GtkTreeViewColumn *column)
 {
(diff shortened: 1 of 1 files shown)


treebrowser/src/treebrowser.c
```diff
@@ -138,11 +138,7 @@ PLUGIN_SET_TRANSLATABLE_INFO(
 static GList*
 _gtk_cell_layout_get_cells(GtkTreeViewColumn *column)
 {
-#if GTK_CHECK_VERSION(2, 12, 0)
```
codebrainz: Why not remove this whole function? It looks like it was only added for backwards compatibility pre-2.12, and has a confusing name.

LarsGit223: Done in #725.

PR #724, opened 2018-03-11
scope: new-maintainer: LarsGit223

LarsGit223:
As mentioned earlier I use the scope plugin from time to time and I would be willing to become the maintainer. I'll try my best to keep it alive.


frlan:
Can you please add update for MAINTAINERS-file?

PR #729, opened 2018-03-17
Spellcheck: support new "enchant-2" package name

### LINKED ISSUE #728: Spellcheck not available with latest Arch Linux release.
geany 1.33 (built on 2018-03-06 with GTK 3.22.28, GLib 2.54.3)
geany-plugins 1.33-1
aspell 0.60.6.1-5
enchant 2.2.3-1
hunspell 1.6.2-1
hunspell-en 2017.01.22-2
hunspell-en_US 2017.01.22-2
linux 4.15.8-1-ARCH

After a full system update, the Spellcheck plugin is no longer in the Plugin Manager dialog.

eht16:
Fixes #728.

### DIFF
--- a/build/spellcheck.m4
+++ b/build/spellcheck.m4
@@ -2,17 +2,34 @@ AC_DEFUN([GP_CHECK_SPELLCHECK],
 [
     GP_ARG_DISABLE([spellcheck], [auto])
 
+    ENCHANT_PACKAGE_NAME=enchant-2
     ENCHANT_VERSION=1.3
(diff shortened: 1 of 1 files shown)


eht16:
We check first for new `enchant-2` package and fallback to the "old" `enchant` package name if the newer one is not found.

Tested against Enchant 2.2 and 1.6.

eht16:
For now, it is not required so I didn't add it on purpose. I would add
it once we need it.

PR #735, opened 2018-04-16
vimode: A Vim Mode for Geany

techee:
I know what you think - the last thing Geany needed...

This plugin depends on adding the "key-press" signal to Geany and won't work without it (even though it will probably compile alright). It shouldn't be merged before this functionality is in Geany.

@frlan Now it's finally pull request ready so if you have some patches or suggestions, please let me know.


techee:
This is the corresponding pull request in Geany: https://github.com/geany/geany/pull/1829

elextr:
@techee since this plugin is the only way of testing the Geany PR, maybe you need some Vimist to test them both.

techee:
@elextr Nah, brave non-vim user is enough. To make sure the Geany PR works, one only needs to know 1 vi command - "i". When you press "i", you enter the insert mode without adding the "i" character to the editor (that means that the "key-press" event "return FALSE" works). Now you can type something which shows in the editor (which means that the "key-press" event "return TRUE" works).

That's it. Now you can quickly disable the plugin to avoid further brain damage.

elextr:
@techee `<esc><esc>:wq!` 1970s neurons still firing :)

pcworld:
I used an Ubuntu 14.04 VM to test this out.

- I took the sources of the [Geany 1.33 PPA](https://launchpad.net/~geany-dev/+archive/ubuntu/ppa), applied [your Geany patch](https://github.com/geany/geany/pull/1829) (I ignored the `GEANY_API_VERSION` hunk as that failed to apply), then built with `dpkg-buildpackage -us -uc -nc` and installed the resulting packages.  
I'm not sure if the Geany patch is still necessary though given the comments in the PR, and I see that you did a force push but I can't find the original commit hash to compare with.
- I cloned your branch and built vimode, copied `vimode.so` to `~/.config/geany/plugins` and enabled the plugin in Geany.

I get the block cursor so it seems the plugin itself is enabled. However it does not seem to capture keys correctly, i.e. typing text inserts text, Ctrl-W closes the current tab and pressing keys when text is selected replaces the selection. Pressing escape does not seem to have any effect. I did not change the vimode options. Note that `viw` behaves the same way.

Feel free to ignore if it's clear that it's my fault.

vimode/src/keypress.c
```diff
@@ -0,0 +1,196 @@
...
+
+#include <gdk/gdkkeysyms.h>
+
+KeyPress *kp_from_event_key(GdkEventKey *ev)
+{
+	guint mask = GDK_MODIFIER_MASK & ~(GDK_SHIFT_MASK | GDK_LOCK_MASK | GDK_CONTROL_MASK);
+	KeyPress *kp;
+
+	if (ev->state & mask)
+		return NULL;
```
pcworld: I found the cause of my issue with vimode not recognizing keypresses. I have num lock enabled, which maps to [`GDK_MOD2_MASK`](https://github.com/GNOME/gtk/blob/cfa04805a327300056b7cfb5b3c127e667b8f4c3/gdk/gdktypes.h#L230) in GDK in X11 (in my configuration at least). The following patch fixes it for me, however an even less restrictive mask might be appropriate in order to avoid further false positives:

```diff
--- a/vimode/src/keypress.c
+++ b/vimode/src/keypress.c
@@ -23,7 +23,7 @@
 
 KeyPress *kp_from_event_key(GdkEventKey *ev)
 {
-       guint mask = GDK_MODIFIER_MASK & ~(GDK_SHIFT_MASK | GDK_LOCK_MASK | GDK_CONTROL_MASK);
+       guint mask = GDK_MODIFIER_MASK & ~(GDK_SHIFT_MASK | GDK_LOCK_MASK | GDK_CONTROL_MASK | GDK_MOD2_MASK);
        KeyPress *kp;
 
        if (ev->state & mask)
```

techee: @pcworld Thanks, I've changed the code to return NULL only if ev->state contains Alt (GDK_MOD1_MASK). Should be enough and makes sure some other modifier isn't missed by accident.

Sorry for the force pushes, I didn't assume someone would make pull requests before it's merged. I'll do normal commits from now on so if you want to do more pull requests, there should be no problems now.

frlan:
@techee  Still depending on the PR on Geany core, right?

techee:
I've just tried to compile against Geany master and everything seems to work fine. So I guess the patch can be merged unless something else is missing.

frlan:
Let's test it in real life

techee:
@frlan Thanks! I'll write an announcement on the mailing list so more people learn about the plugin.

PR #746, opened 2018-05-21
Markdown: add support for webkit2gtk

codebrainz:
Markdown-specific changes from #677.

### DIFF
--- a/build/markdown.m4
+++ b/build/markdown.m4
@@ -49,13 +49,13 @@ AC_DEFUN([GP_CHECK_MARKDOWN],
     GTK_VERSION=2.16
     WEBKIT_VERSION=1.1.13
 
-    GP_CHECK_GTK3([webkit_package=webkitgtk-3.0],
+    GP_CHECK_GTK3([webkit_package=webkit2gtk-4.0],
                   [webkit_package=webkit-1.0])
-
     GP_CHECK_PLUGIN_DEPS([markdown], [MARKDOWN],
                          [$GP_GTK_PACKAGE >= ${GTK_VERSION}
                           $webkit_package >= ${WEBKIT_VERSION}
                           gthread-2.0])
+    AM_CONDITIONAL([MARKDOWN_WEBKIT2], [test "$webkit_package" = webkit2gtk-4.0])
 
     GP_COMMIT_PLUGIN_STATUS([Markdown])
 
--- a/markdown/src/Makefile.am
+++ b/markdown/src/Makefile.am
@@ -35,4 +35,8 @@ markdown_la_CFLAGS += $(LIBMARKDOWN_CFLAGS)
 markdown_la_LIBADD += $(LIBMARKDOWN_LIBS)
 endif
 
+if MARKDOWN_WEBKIT2
+markdown_la_CFLAGS += -DMARKDOWN_WEBKIT2
+endif
+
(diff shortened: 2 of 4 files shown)


build/markdown.m4
```diff
@@ -49,13 +49,13 @@ AC_DEFUN([GP_CHECK_MARKDOWN],
     GTK_VERSION=2.16
     WEBKIT_VERSION=1.1.13
 
-    GP_CHECK_GTK3([webkit_package=webkitgtk-3.0],
+    GP_CHECK_GTK3([webkit_package=webkit2gtk-4.0],
```
codebrainz: @hyperair I wonder if we can cascade from `webkit-1.0` to `webkit-3.0` to `webkitgtk-4.0` to try and support all versions?

codebrainz: Not tested, but I mean like this:

```sh
GP_CHECK_GTK3([webkit_package=webkit-3.0], [
  GP_CHECK_GTK3([webkit_package=webkit2gtk-4.0], [
    webkit_package=webkit-1.0
  ]
])
...
AM_CONDITIONAL([MARKDOWN_WEBKIT2], [test "x$webkit_package" = "xwebkit2gtk-4.0"])
...
```

@b4n @sardemff7 @dmaphy @evgeni @hyperair care to comment?

sardemff7:
If you don’t go with fallback checks, better drop the `#else`s. Other comments inside.

build/markdown.m4
codebrainz: I think the versions listed are "supported" in that they're known to build and work for the plugin, but not that they are "supported" upstream, obviously. IMO, if it's a simple matter of a few ugly `#ifdef` it's worthwhile even if just to allow people on slightly older distros to compile the latest plugin release tarball.

markdown/src/Makefile.am
```diff
@@ -35,4 +35,8 @@ markdown_la_CFLAGS += $(LIBMARKDOWN_CFLAGS)
 markdown_la_LIBADD += $(LIBMARKDOWN_LIBS)
 endif
 
+if MARKDOWN_WEBKIT2
+markdown_la_CFLAGS += -DMARKDOWN_WEBKIT2
```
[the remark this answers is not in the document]
codebrainz: Yeah, it does seem better.

codebrainz:
TODO: rebase ontop of #747.

markdown/src/Makefile.am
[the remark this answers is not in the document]
hyperair: `AC_DEFINE` affects everything, so plugins that don't care about this variable will get it too. I originally wanted to just have a `WEBKIT2` macro defined only for `markdown` based on the `MARKDOWN_WEBKIT2` conditional, but @codebrainz changed that.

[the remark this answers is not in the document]
codebrainz: It won't clash because I added the namespace prefix, so it doesn't seem like an issue.

build/markdown.m4
hyperair: Might be useful -- on Ubuntu's side, there's still one remaining release (trusty) that has webkitgtk-3.0 but not webkit2gtk-4.0.

frlan:
@codebrainz  Would you mind on doing thementioned rebase and I'd merge it

PR #748, opened 2018-05-31
vimode: Fix goto_down with line wrapping

pcworld:
SCI_LINEDOWN goes down only one display line, hence when line wrapping
was enabled and a line wrapped over more than two lines, it was
impossible to go to the next actual line when pressing j.

Note that this still doesn't emulate vim behavior correctly where
caret-x position is saved for actual lines and not display lines. But
this change at least makes j go the next line when wrapping is enabled.

Note that I'm not completely sure whether this change retains all workarounds [mentioned in `cmd_goto_up`](https://github.com/geany/geany-plugins/blob/341cab4e13c122b2b74e6f31a43be9adc653f8af/vimode/src/cmds/motion.c#L51-L54).

### DIFF
--- a/vimode/src/cmds/motion.c
+++ b/vimode/src/cmds/motion.c
@@ -78,7 +78,7 @@ static void goto_down(CmdParams *p, gint num)
 	/* see cmd_goto_up() for explanation */
 	one_above = p->line + num - 1;
 	one_above = one_above < last_line ? one_above : last_line - 1;
-	pos = SSM(p->sci, SCI_POSITIONFROMLINE, one_above, 0);
+	pos = SSM(p->sci, SCI_GETLINEENDPOSITION, one_above, 0);
 	SET_POS_NOX(p->sci, pos, FALSE);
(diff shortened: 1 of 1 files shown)


techee:
Yep, looks good to me. I don't use wrapping much so I haven't noticed this. I can imagine there will be more problems with wrapping with some other commands.

Your patch works correctly because SET_POS_NOX() doesn't update the remembered x position and subsequent SCI_LINEDOWN recovers it. It's real pity there's no getter for the remembered x position in Scintilla and that we have to do stuff like this.

@frlan Can be merged I think.

techee:
@pcworld Just a note - I don't look at the geany-plugins page that often and I'm not subscribed to receive all the pull requests so if you want to submit some other patches, it's better to @techee me so I'm notified something is happening.

frlan:
@techee  Good for merging anyway?

techee:
@frlan Yeah, it can be merged.

PR #749, opened 2018-06-03
Revert "Port both webhelper and markdown to webkit2gtk"

frlan:
Reverts geany/geany-plugins#677


frlan:
Puh. I really hope I didn't mess it up again

b4n:
I'm not 100% sure how easy it is to reapply commits that have been reverted, short of making sure they are different commits (as otherwise they are already merged), so it might be slightly annoying for that other Markdown PR.

Also, for WebHelper I don't really mind so long as the small problems gets fixed; I wanted to keep GTK2 support, but it's also true that in many recent distros (including next Debian) there is no more webkit1, so it's just not so useful anymore; and the delta between webkit1 and webki2 is bigger than a few GTK2/GTK3 conditionals.
The only thing is that IIRC there are a few real issues with the webkit2 version that got merged, and it'd be sad to release with them, if they really exist.  I might (or might not, no promises) have time to review and fix the issues before next release, but if people find and fix them already even better.  My point is that the only thing that I really don't fancy with the merge are the potential bugs in the new version, not *so* much the drop of GTK2 support.

But either way is good, and it'd be just fine for me to work on a new PR if this is reverted.

codebrainz:
> so it might be slightly annoying for that other Markdown PR

Not sure if I'm just confused, but wasn't it already merged? I honestly thought this revert had already happened as I checked the diff and #746 and #677 had no differences, so I didn't think it would be meaningful to merge until this revert happened.

> Also, for WebHelper I don't really mind so long as the small problems gets fixed; I wanted to keep GTK2 support

That was my issue with the Markdown changes, and it was somewhat fixed in follow ups, though I'd like to improve it to still [support all of the versions of webkit](https://github.com/geany/geany-plugins/pull/746#pullrequestreview-121957556) it's known to work with (the versions/package names are confusing but I guess `webkit-1.0`, `webkitgtk-3.0`, and `webkit2gtk-4.0`).

> it'd be just fine for me to work on a new PR if this is reverted

IMO this is the best way, #677 should have never been made to change two plugins at once (especially plugins with different maintainers), I think that's where a lot of the confusion/problems came from. Maybe in the future we should just decline PRs which affect multiple plugins unless it's related to translations or build system or other more global stuff like that.

elextr:
>  Not sure if I'm just confused, but wasn't it already merged? I honestly thought this revert had already happened as I checked the diff and #746 and #677 had no differences for Markdown, so I didn't think it would be meaningful to merge until this revert happened.

@frlan AFAICT the merge of #746 happened, but not the this revert, see markdown [history](https://github.com/geany/geany-plugins/commits/master/markdown).  This is still a pull request that is not merged.  Not sure what is going to happen when this revert is applied, but as @codebrainz says, since #746 is a subset of #677, reverting #677 now may also revert #746.

codebrainz:
@frlan that's what I thought, but then what is this PR about?

frlan:
@codebrainz  litter. Going to clean it up

frlan:
The PR was created by accident

elextr:
I think whats confusing is that fb82de4 doesn't show in the history of, for example, markdown/src/viewer.c.

PR #751, opened 2018-06-05
scope: removed invalid signal 'project-before-save' and callback function

LarsGit223:
This prevents a debug message like 

> "signal 'project-before-save' is invalid for instance '0xd7da10' of type 'GeanyObject'"

 being generated on loading the scope plugin.

### DIFF
--- a/scope/src/scope.c
+++ b/scope/src/scope.c
@@ -351,12 +351,6 @@ static void on_settings_save(G_GNUC_UNUSED GObject *obj, G_GNUC_UNUSED GKeyFile 
 	schedule_settings_saved(TRUE);
 }
 
-static void on_project_before_save(G_GNUC_UNUSED GObject *obj, G_GNUC_UNUSED GKeyFile *keyfile,
-	G_GNUC_UNUSED gpointer gdata)
-{
-	schedule_settings_saved(FALSE);
-}
-
 static gboolean on_editor_notify(G_GNUC_UNUSED GObject *obj, GeanyEditor *editor,
 	SCNotification *nt, G_GNUC_UNUSED gpointer gdata)
 {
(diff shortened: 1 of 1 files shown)


elextr:
Is it meant to be `project-before-close`?

b4n:
https://lists.geany.org/pipermail/devel/2012-February/006537.html and IIRC more of that but I can't find it right now.  AFAIR the author used the signal for the sake of people using a patched version of Geany with the signal, and leaving others with degraded features.  It'd be interesting to re-check the reasons and possibly add this to Geany, 6 years later… :confused:

LarsGit223:
Or maybe ```project-save```?

For the ```project-before-save``` event the following code would have been run (with ```conterm=FALSE```):
```
	guint i = 0;

	plugin_idle_add(geany_plugin, settings_saved, GINT_TO_POINTER(conterm));

	foreach_document(i)
	{
		if (utils_attrib(documents[i], SCOPE_LOCK))
			documents[i]->readonly = FALSE;
	}
```
And function ```settings_saved``` looks like this:
```
static gboolean settings_saved(gpointer gdata)
{
	guint i = 0;

	foreach_document(i)
	{
		documents[i]->readonly = scintilla_send_message(documents[i]->editor->sci,
			SCI_GETREADONLY, 0, 0);
	}

	if (gdata)
	{
		conterm_load_config();
		conterm_apply_config();
	}

	return FALSE;
}
```
The code in ```if (gdata)``` is not executed in this case as ```gdata``` would be ```FALSE```. The function is also called on the event ```save-settings``` and then ```conterm```/```gdata``` is set to ```TRUE``` and the config will be reloaded. This still works fine. If I e.g. change the terminal background color in the preferences the change is applied immediately.

I wonder what this READONLY check is good for. Can the document status change by changing/saving the project?

As I never have seen the six years old code in action I can't see any difference and it looks useless to me.

Maybe there were plans to also save some scope settings in the project then it would make sense to also reload the config at that point. But all settings which scope loads in ```conterm_load_config();``` come from ```geany.conf```.

b4n:
@LarsGit223 looking at the code I would think that the goal is not to write the project file with some files marked read-only just because the Scope plugin "locked" them.
I guess that without this, if you save the project settings while a debugging session is running, many of your files will be marked as read-only, so if you don't re-save it after the debugging session ends, next time you open it every file will be marked read-only.

elextr:
@b4n, why would scope mark files readonly? Isn't that for the user to do?

b4n:
IIUC it does this to prevent the user from modifying files during a debugging session and thus getting the state desynchronized from the debugging symbols (file/line reported by GDB)

elextr:
But it doesn't prevent it because the user can always remove the readonly can't they?  Seems to me its trying too hard to nappy change users and getting itself into trouble.

PR #754, opened 2018-06-14
vimode: When wrapping is on, "k" should go to the first line of wrapping

techee:
This is a sister implementation for "k" in order to behave like "j" when wrapping is on (see #748). The problem with the current implementation is that "k" moves caret to the last line of a wrapped line instead of the first one like "j".

@pcworld Can you check if it looks good to you? This one is a bit more complex because one has to either "iterate" over the wraps when going up (for the first line only) or go to the previous line and go down.

### DIFF
--- a/vimode/src/cmds/motion.c
+++ b/vimode/src/cmds/motion.c
@@ -43,20 +43,48 @@ void cmd_goto_right(CmdContext *c, CmdParams *p)
 
 void cmd_goto_up(CmdContext *c, CmdParams *p)
 {
-	gint one_below, pos;
+	gint one_above, pos;
 
 	if (p->line == 0)
 		return;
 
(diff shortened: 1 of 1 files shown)


pcworld:
Sorry I don't have time to test currently. Feel free to merge without my feedback, users can still complain afterwards. ;)

The behavior is kind of weird anyway, with or without this (always jumping to first or last line). In original vim j or k go to the same position in the line regardless of wrapping. I.e. it doesn't reset the column counter at line wraps.

Btw the original arrow up / arrow down behavior from Geany is `gk` and `gj` in vim, respectively. It goes up/down display lines.

techee:
@pcworld Yeah, it behaves differently because of how Scintilla works - not much we can do unfortunately.

@frlan Would you merge this?

PR #755, opened 2018-06-14
geanyctags: Use base path as relative to the project file path

### LINKED ISSUE #698: GeanyCtags does not respect relative project base path.
Geany 1.32.

Preferences->General->Miscellaneous: "Project file is stored inside the project base directory" is checked. A project is open.

If project base path is absolute, everything works as expected. However, if project base path is `.`, "Generate tags" hangs.

I found that Geany does not hang actually, but waits for `ctags`. However, it indexes *not* the project directory but *current* directory (which intentionally or accidentally is my home directory). The current directory contains many projects, indexing it requires a lot of time.

While Geany is waiting for `ctags`, interface freezes. I would expect Geany builds index in background (like Geany builds the project), so user interface does not freeze.

techee:
When prj->base_path is relative, construct path using prj->file_name
so base path is relative to the project file directory.

Fixes #698

### DIFF
--- a/geanyctags/src/geanyctags.c
+++ b/geanyctags/src/geanyctags.c
@@ -205,6 +205,25 @@ static gchar *generate_find_string(GeanyProject *prj)
 }
 
 
+static gchar *get_base_path(void)
+{
+	gchar *ret;
+	GeanyProject *prj = geany_data->app->project;
+	gchar *project_dir_utf8;
+
+	if (!prj)
+		return NULL;
+
+	if (g_path_is_absolute(prj->base_path))
+		return g_strdup(prj->base_path);
+
+	project_dir_utf8 = g_path_get_dirname(prj->file_name);
+	ret = g_build_filename(project_dir_utf8, prj->base_path, NULL);
+	g_free(project_dir_utf8);
+	return ret;
+}
+
+
 static void
 on_generate_tags(GtkMenuItem *menuitem, gpointer user_data)
 {
@@ -215,6 +234,7 @@ on_generate_tags(GtkMenuItem *menuitem, gpointer user_data)
 	{
 		gchar *cmd;
 		gchar *tag_filename;
+		gchar *base_path;
 
 		tag_filename = get_tags_filename();
 
@@ -236,10 +256,12 @@ on_generate_tags(GtkMenuItem *menuitem, gpointer user_data)
 			tag_filename, "\"", NULL);
 #endif
 
-		spawn_cmd(cmd, prj->base_path);
+		base_path = get_base_path();
+		spawn_cmd(cmd, base_path);
 
 		g_free(cmd);
 		g_free(tag_filename);
+		g_free(base_path);
 	}
 }
 
@@ -412,6 +434,7 @@ static void find_tags(const gchar *name, gboolean declaration, gboolean case_sen
 	tagFile *tf;
 	GeanyProject *prj;
 	gchar *tag_filename = NULL;
+	gchar *base_path;
 	tagEntry entry;
 	tagFileInfo info;
 	int last_line_number = 0;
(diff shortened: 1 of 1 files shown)


geanyctags/src/geanyctags.c
```diff
@@ -205,6 +205,26 @@ static gchar *generate_find_string(GeanyProject *prj)
...
+	GeanyProject *prj = geany_data->app->project;
+	gchar *project_dir_utf8;
+
+	if (!prj)
+		return NULL;
+
+	if (g_path_is_absolute(prj->base_path))
+		return prj->base_path;
+
+	g_free(ret);
```
b4n: not useful as it's always `NULL` at this point

geanyctags/src/geanyctags.c
```diff
@@ -205,6 +205,26 @@ static gchar *generate_find_string(GeanyProject *prj)
 }
 
 
+static const gchar *get_base_path(void)
```
b4n: this function has a leaky semantic: it returns *either* `prj->base_path` *or* an allocated buffer.  The caller must then free conditionally, which is tricky at best.

geanyctags/src/geanyctags.c
```diff
@@ -236,7 +256,7 @@ on_generate_tags(GtkMenuItem *menuitem, gpointer user_data)
 			tag_filename, "\"", NULL);
 #endif
 
-		spawn_cmd(cmd, prj->base_path);
+		spawn_cmd(cmd, get_base_path());
```
b4n: leaks when `prj->base_path` is relative

geanyctags/src/geanyctags.c
```diff
@@ -421,7 +441,7 @@ static void find_tags(const gchar *name, gboolean declaration, gboolean case_sen
 		return;
 
 	msgwin_clear_tab(MSG_MESSAGE);
-	msgwin_set_messages_dir(prj->base_path);
+	msgwin_set_messages_dir(get_base_path());
```
b4n: same

geanyctags/src/geanyctags.c
```diff
@@ -445,7 +465,7 @@ static void find_tags(const gchar *name, gboolean declaration, gboolean case_sen
 
 			if (!filter_tag(&entry, name_pat, declaration, case_sensitive))
 			{
-				path = g_build_filename(prj->base_path, entry.file, NULL);
+				path = g_build_filename(get_base_path(), entry.file, NULL);
```
b4n: more

geanyctags/src/geanyctags.c
```diff
@@ -456,7 +476,7 @@ static void find_tags(const gchar *name, gboolean declaration, gboolean case_sen
 				if (!filter_tag(&entry, name_pat, declaration, case_sensitive))
 				{
 					if (!path)
-						path = g_build_filename(prj->base_path, entry.file, NULL);
+						path = g_build_filename(get_base_path(), entry.file, NULL);
```
b4n: and… it's me again :grin:

geanyctags/src/geanyctags.c
```diff
@@ -205,6 +205,26 @@ static gchar *generate_find_string(GeanyProject *prj)
 }
 
 
+static const gchar *get_base_path(void)
+{
+	static gchar *ret = NULL;
```
b4n: I'm not particularly fond of the non-reentrant semantic, but it indeed has the advantage of making other code changes simple.

techee: Things start making much more sense once you notice the magic word "static", don't they ;-). And yes, it probably won't win any competition in code beautifulness but despite that it makes other code simpler and easier to maintain.

techee: @codebrainz Thanks. While I wouldn't care much for a standalone application (such code is used in Geany's tag manager for instance), your argument about dangers of `static` in dynamically loaded libraries is something worth considering so I have rewritten the patch to avoid this problem.

PR #765, opened 2018-08-10
projorg: add file operations to the side bar.

dgmz:
Add operations to quickly add, delete and rename files and directories from the project. These operations appear as operations in the side bar context menu.


dgmz:
Will post again after fixes.

PR #767, opened 2018-08-10
Add file operations to Project Organizer.

dgmz:
Add file and directory creation, renaming and removal operations to the Project Organizer sidebar.

### DIFF
--- a/projectorganizer/src/prjorg-project.c
+++ b/projectorganizer/src/prjorg-project.c
@@ -114,6 +114,7 @@ static GSList *get_file_list(const gchar *utf8_path, GSList *patterns,
 
 			if (!patterns_match(ignored_dirs_patterns, utf8_name))
 			{
+				list = g_slist_prepend(list, g_build_path(G_DIR_SEPARATOR_S, utf8_filename, PROJORG_SENTINEL_FILENAME, NULL));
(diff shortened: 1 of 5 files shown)


techee:
@dgmz Shame on me - took me ages to look at your patches :-(.

In general, I really like the added features and the implementation looks good too. I've just made some extra patches on top of yours with some minor fixes here:

https://github.com/techee/geany-plugins/tree/projorg_file_ops

Please let me know what you think about them.

The only thing I dislike is that empty directories are shown now - but at the same time this is necessary when you want to have the "create directory" functionality and allow users to add files to the empty directories. So it will probably have to stay this way.

techee:
> All look very good. Most of those are things I should have considered myself, sorry for that.

Great, good to hear!

> Regarding empty directories, yes, this makes the tree a bit more crowded. It would be possible to haved them hidden and show them only when their parent is selected, maybe with a dimmed icon, but I thought it may be a little bit confusing and, more important, it was too complex for the time I had available.

Yeah, I think it's not worth it. I was thinking about some config option to disable showing empty directories (which would kind of kill the "create directory" feature for users who would enable it) but I think I'll wait with this and see if users are complaining about it or just don't care.

I've created #783 with the extra patches and will close this pull request. Thanks a lot for contributing the code!

Issue #761, opened 2018-07-28
Unable to perform version check.


Beyarz:
Unable to perform version check.
Error code: 6
Error message: »Unacceptable TLS certificate«

Here's the debug:
19:57:29: Updatechecker MESSAGE	: Checking for updates (querying URL "http://geany.org/service/version.php")
19:57:30: Updatechecker WARNING	: Connection error: Code: 6; Message: Unacceptable TLS certificate

eht16:
I assume you are using Windows?
If so, try to create a new file in `C:\Program Files (x86)\Geany\ssl\certs\ca-bundle.crt` with the contents in https://pastebin.geany.org/eR622/, restart Geany and try again (obviously adjust the Geany installation path if it differs).

It might help to restart Geany after creating the file.
The pasted CA bundle only contains one of the Let's Encrypt intermediate certificates which is necessary for a proper TLS connection.

If it works then (it did on my Windows test system), then we will ship this certificate with future releases).

Beyarz:
I'm on macOS :)

eht16:
Ok, not sure where to place the file then.
Is there some global or system-wide way to add CA bundles?

If not, we probably need to ship the bundle for macOS as well after we found out where to place it :).

Beyarz:
I tried to import the certificate to the keychain access app but without any success.

techee:
No idea really (see #752).

Beyarz:
So I assume the issue is not with Geany?

techee:
No, it's not Geany itself (works under Linux), the problem is with the dependencies of Geany and their packaging on macOS.

Beyarz:
I see, so what can we do in order to solve this?

techee:
Well, one can play with gtk-osx packages, try updating to latest releases of libraries, fix compilation problems, check if all the files are bundled correctly... which I'm not going to do myself (no fun). But it's also possible someone already did it in the meantime and that the next time I make a release these things will be fixed.

Beyarz:
I'll give it a try!

techee:
@Beyarz You are a hero then! You can find everything related to Geany packaging here:

https://github.com/geany/geany-osx

PR #768, opened 2018-08-19
Windows installer: Install ca-certificates for UpdateChecker plugin

eht16:
For the UpdateChecker plugin we need the Let's Encrypt intermediate
certificate to properly validate TLS certificates when querying
geany.org for version check.

Slightly related to #761.


eht16:
- Tested on Windows
- without the certificate bundle ("bundle" here is only one of the intermediate certificates of Let's Encrypt) there is an error message like the one reported in #761 
- with the certificate bundle the version check performs successfully

Still I'm not completely sure if it is a good idea to distribute the certificate or whether this should rather be done by the user and/or OS and we just rely on that.

b4n:
I don't really like the idea that we're installing certificates that aren't even ours, but I guess that if we really have to installing the ones form MSYS is probably simpler and safer indeed.

eht16:
Installing "our" certificates is not an option with Let's Encrypt as those certificates expire after three months or we switch to a very short release cycle :).
Or we just disable certificate verification in the UpdateChecker plugin :).

More seriously, I got your point.
Ideally, Windows would provide the necessary CA certificates or `libsoup` would be able to access them (I'm not completely sure what's the real culprit here). But since isn't the case, I think re-distributing the available set of CA certificates bundled by MSYS and updating them with our releases, should be a good compromise.

PR #769, opened 2018-08-31
FIX: Spaces after dots in translatable strings

Skif-off:
See diff.

### DIFF
--- a/geanyvc/src/geanyvc.c
+++ b/geanyvc/src/geanyvc.c
@@ -1948,7 +1948,7 @@ plugin_configure(GtkDialog * dialog)
 	widgets.cb_attach_to_menubar = gtk_check_button_new_with_label(_("Attach menu to menubar"));
 	gtk_widget_set_tooltip_text(widgets.cb_editor_menu_entries,
 			     _("Whether menu for this plugin are getting placed either "
-			       "inside tools menu or directly inside Geany's menubar."
+			       "inside tools menu or directly inside Geany's menubar. "
 			       "Will take in account after next start of GeanyVC"));
 	gtk_button_set_focus_on_click(GTK_BUTTON(widgets.cb_attach_to_menubar), FALSE);
 	gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(widgets.cb_attach_to_menubar),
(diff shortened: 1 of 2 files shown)


workbench/src/dialogs.c
```diff
@@ -443,7 +443,7 @@ gboolean dialogs_workbench_settings(WORKBENCH *workbench)
 	ui_table_add_row(GTK_TABLE(table), 1, w_enable_live_update, NULL);
 	gtk_widget_set_tooltip_text(w_enable_live_update,
 		_("If the option is activated (default), then the list of files and the sidebar"
-		  " will be updated automatically if a file or directory is created, removed or renamed."
+		  " will be updated automatically if a file or directory is created, removed or renamed. "
```
b4n: Ideally the leading space here would also be on the previous line, but that's OK with this.

geanyvc/src/geanyvc.c
```diff
@@ -1949,7 +1949,7 @@ plugin_configure(GtkDialog * dialog)
 	gtk_widget_set_tooltip_text(widgets.cb_editor_menu_entries,
 			     _("Whether menu for this plugin are getting placed either "
 			       "inside tools menu or directly inside Geany's menubar."
-			       "Will take in account after next start of GeanyVC"));
+			       " Will take in account after next start of GeanyVC"));
```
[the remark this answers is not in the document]
b4n: Forget my comment, it was absurd, I read the diff the wrong way around.

Skif-off:
Done (fix + squash).

Sorry, I saw strings with space in beginning below/above and maked it so ([1](https://github.com/geany/geany-plugins/blob/master/workbench/src/dialogs.c#L455), [2](https://github.com/geany/geany-plugins/blob/master/workbench/src/dialogs.c#L455), [3](https://github.com/geany/geany-plugins/blob/master/workbench/src/dialogs.c#L455)).

workbench/src/dialogs.c
Skif-off: Will do it?

Skif-off: Ops, merged.

frlan: Did see the leading space to late. Will change it directly

b4n:
I also updated the strings in the translations to include the space, as it doesn't have to be reviewed by translators (I made sure that the translations didn't have the same dot issue, and they don't): 65475dac74800f8e07d0102beeb64ed1a2612212

LarsGit223:
I added a small section "Translation" to geany's HACKING file to prevent others from making the same fault as me.

PR #776, opened 2018-09-21
GeanyLua: update for new Scintilla 3.10

b4n:
@Skif-off could you take a look, as you seem to be interested in GeanyLua?

This change is required for GeanyLua to build against current Geany master.

I'm not sure of the implication of the changes in messages arguments, but this is a GeanyLua API break as well so it might be alright anyway.


Skif-off:
@b4n 
Thank you for PR! I also did it two-three of weeks ago and got the same changes.

b4n:
@Skif-off thanks for the review, and sorry not to have given you the time to submit yours ^^

PR #789, opened 2018-11-14
open with glade feature for .ui and .glade files in tree browser

aferust:
Treebrowser has "open with glade" feature for .ui and .glade files in this fork. Tested on Ubuntu 16.04 LTS 64 bit and Windows 10 64bit (msys - mingw64).

### DIFF
--- a/treebrowser/src/treebrowser.c
+++ b/treebrowser/src/treebrowser.c
@@ -942,6 +942,34 @@ on_menu_current_path(GtkMenuItem *menuitem, gpointer *user_data)
 }
 
 static void
+on_menu_open_with_glade(GtkMenuItem *menuitem, gchar *uri)
+{
+	gchar 				*cmd, *locale_cmd, *dir, *c;
+	GString 			*cmd_str 	= g_string_new("glade '%f'");
+	GError 				*error 		= NULL;
+
+	dir = g_file_test(uri, G_FILE_TEST_IS_DIR) ? g_strdup(uri) : g_path_get_dirname(uri);
+    
+	utils_string_replace_all(cmd_str, "%f", uri);
+    
+	cmd = g_string_free(cmd_str, FALSE);
+	locale_cmd = utils_get_locale_from_utf8(cmd);
+	if (! g_spawn_command_line_async(locale_cmd, &error))
+	{
+		c = strchr(cmd, ' ');
+		if (c != NULL)
+			*c = '\0';
+		ui_set_statusbar(TRUE,
+			_("Could not execute configured external command '%s' (%s)."),
+			cmd, error->message);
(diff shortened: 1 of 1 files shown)


eht16:
Do we want to add "Open with XY" for many different tools here?
Why is opening a file Glade in any way special? Next someone wants to add "Open with Firefox", then "Open with Gimp", then "Open with Foo", ...

Maybe a more generic approach like using "xdg-open" with a generic "Open" action is the better way?

b4n:
> Maybe a more generic approach like using "xdg-open" with a generic "Open" action is the better way?

"Open Externally" or such then, because I would expect the first action is to open it in Geany :smile: -- well, for text files at least.

aferust:
I thought glade is in gnome ecosystem and it deserves a special action for opening glade files. Such as, intellij has "open in scenebuilder" feature for *.fxml files of javafx.

elextr:
Glade may be in the Gnome ecosystem, but Geany is not. Geany is a somewhat portable application that uses GTK.

aferust:
Ok guys. I agrree with you more generic approach would fit a wide range of requirements. Please forgive my ignorance daring a pull request for this. I am new to all gtk stuff, and I could not find a customizable behavior for opening those files using glade within geany.  I know treebrowser has this: executable "%f", however It sets opening action for all extension types. May be it is my fault to unable to discover a pre-existing feature. I am open to get informed.

aferust:
Now, users can associate files with any executable like:
geany:.glade;geany:.ui;notepad:.txt:winamp:mp3
or whatever else
I hope everbody is happy now :)

treebrowser/src/treebrowser.c
```diff
@@ -62,8 +62,11 @@ static gboolean 			flag_on_expand_refresh 		= FALSE;
 # define CONFIG_OPEN_TERMINAL_DEFAULT "cmd"
 #endif
 
+# define CONFIG_OPEN_EXT_USER_DEF_DEFAULT "notepad++:.txt;glade:.ui;glade:.glade"
```
codebrainz: The first default here seems unusual...

aferust: fixed!

LarsGit223:
@aferust: I can resolve the conflicts if you like because they come from my PR #864. Doing that will change your PR and branch ```aferust:master```.

aferust:
@LarsGit223 : Yes, please go for it. I almost forgot that I did this PR a while ago. I am not sure what to do, I am not an expert on git system. Please tell me if you need me to do something.

LarsGit223:
@aferust: done, you don't need to do anything.

@medovina: your the maintainer of treebrowser - have a look to maybe.

PR #790, opened 2018-11-20
'Debugger' plugin port to GTK3 with GTK2 compatibility

bxgaillard:
This pull request is heavily based on #645 from which it borrows the majority of the changes.
The main improvement is that this one doesn't break compatibility with GTK2 (there are macros which select the correct code path accordingly).
I also removed some cosmetic-only changes to maintain the differences as small as possible and provided a corrected code for console font selection (which was just commented out in the original pull request).

This was successfully tested on Centos 7.5 with stock development packages.


bxgaillard:
Yes I'm aware, buy I somehow managed to completely break my forked repository, and it was then a mess to sort out, with tons of changes from the master branch... Sorry for the inconvenience.

codebrainz:
No worries, I mostly wanted to link the two issues and point out (because a lot of people don't realize) that you can just force push your changes to the existing PR (even if you messed it up badly).

PR #791, opened 2018-11-20
'Debugger' plugin port to GTK3 with GTK2 compatibility

bxgaillard:
This pull request is heavily based on #645 from which it borrows the majority of the changes.
The main improvement is that this one doesn't break compatibility with GTK2 (there are macros which select the correct code path accordingly).
I also removed some cosmetic-only changes to maintain the differences as small as possible and provided a corrected code for console font selection (which was just commented out in the original pull request).

This was successfully tested on Centos 7.5 with stock development packages.

### DIFF
--- a/debugger/src/btnpanel.c
+++ b/debugger/src/btnpanel.c
@@ -82,9 +82,13 @@ static void on_execute_until(GtkButton *button, gpointer user_data)
  */
 GtkWidget* btnpanel_create(on_toggle cb)
 {
+#if GTK_CHECK_VERSION(3, 0, 0)
+	GtkWidget *vbox = gtk_box_new(GTK_ORIENTATION_VERTICAL, CP_BUTTONS_PAD);
+	GtkWidget *hbutton_box = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, CP_BUTTONS_PAD);
+#else
 	GtkWidget *vbox = gtk_vbox_new(FALSE, CP_BUTTONS_PAD);
-
 	GtkWidget *hbutton_box = gtk_hbox_new(FALSE, CP_BUTTONS_PAD);
+#endif
 
 	runbtn = create_button("run.gif", _("Run"));
 	g_signal_connect(G_OBJECT(runbtn), "clicked", G_CALLBACK (debug_run), (gpointer)TRUE);
@@ -92,7 +96,12 @@ GtkWidget* btnpanel_create(on_toggle cb)
 	gtk_box_pack_start(GTK_BOX(hbutton_box), runbtn, TRUE, TRUE, 0);
 	gtk_box_pack_start(GTK_BOX(vbox), hbutton_box, FALSE, TRUE, 0);
 	
+#if GTK_CHECK_VERSION(3, 0, 0)
+	hbutton_box = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, CP_BUTTONS_PAD);
+	gtk_box_set_homogeneous(GTK_BOX(hbutton_box), TRUE);
+#else
 	hbutton_box = gtk_hbox_new(TRUE, CP_BUTTONS_PAD);
+#endif
 
 	restartbtn = create_button("restart.gif", _("Restart"));
 	g_signal_connect(G_OBJECT(restartbtn), "clicked", G_CALLBACK (debug_restart), (gpointer)TRUE);
@@ -104,7 +113,12 @@ GtkWidget* btnpanel_create(on_toggle cb)
 	gtk_box_pack_start(GTK_BOX(hbutton_box), stopbtn, FALSE, TRUE, 0);
 	gtk_box_pack_start(GTK_BOX(vbox), hbutton_box, FALSE, TRUE, 0);
 
+#if GTK_CHECK_VERSION(3, 0, 0)
+	hbutton_box = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, CP_BUTTONS_PAD);
+	gtk_box_set_homogeneous(GTK_BOX(hbutton_box), TRUE);
+#else
 	hbutton_box = gtk_hbox_new(TRUE, CP_BUTTONS_PAD);
+#endif
 
 	stepoverbtn = create_button("step_over.gif", _("Step over"));
 	g_signal_connect(G_OBJECT(stepoverbtn), "clicked", G_CALLBACK (debug_step_over), (gpointer)TRUE);
@@ -116,7 +130,12 @@ GtkWidget* btnpanel_create(on_toggle cb)
 	gtk_box_pack_start(GTK_BOX(hbutton_box), stepinbtn, FALSE, TRUE, 0);
 	gtk_box_pack_start(GTK_BOX(vbox), hbutton_box, FALSE, TRUE, 0);
 
+#if GTK_CHECK_VERSION(3, 0, 0)
+	hbutton_box = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, CP_BUTTONS_PAD);
+	gtk_box_set_homogeneous(GTK_BOX(hbutton_box), TRUE);
+#else
 	hbutton_box = gtk_hbox_new(TRUE, CP_BUTTONS_PAD);
+#endif
 
 	stepoutbtn = create_button("step_out.gif", _("Step out"));
 	g_signal_connect(G_OBJECT(stepoutbtn), "clicked", G_CALLBACK (debug_step_out), (gpointer)TRUE);
@@ -128,7 +147,11 @@ GtkWidget* btnpanel_create(on_toggle cb)
 	gtk_box_pack_start(GTK_BOX(hbutton_box), runcursorbtn, FALSE, TRUE, 0);
 	gtk_box_pack_start(GTK_BOX(vbox), hbutton_box, FALSE, TRUE, 0);
 
+#if GTK_CHECK_VERSION(3, 0, 0)
+	optbtn = create_stock_button("preferences-system", _("Settings"));
+#else
 	optbtn = create_stock_button(GTK_STOCK_PREFERENCES, _("Settings"));
+#endif
 	g_signal_connect(G_OBJECT(optbtn), "clicked", G_CALLBACK (on_settings), NULL);
 	gtk_box_pack_end(GTK_BOX(vbox), optbtn, FALSE, FALSE, 0);
 
--- a/debugger/src/cell_renderers/cellrendererbreakicon.c
+++ b/debugger/src/cell_renderers/cellrendererbreakicon.c
@@ -40,6 +40,27 @@ enum {
   PROP_HITSCOUNT
 };
 
+struct _CellRendererBreakIcon
+{
+  GtkCellRenderer parent;
+
+  guint enabled;
+  const gchar* condition;
+  guint hitscount;
+
+  GdkPixbuf *pixbuf_enabled;
+  GdkPixbuf *pixbuf_disabled;
+  GdkPixbuf *pixbuf_conditional;
+  GdkPixbuf *pixbuf_file;
+};
+
+struct _CellRendererBreakIconClass
+{
+  GtkCellRendererClass parent_class;
+
+  void (*clicked)(CellRendererBreakIcon *cell_renderer_toggle, const gchar *path);
+};
+
 static gpointer parent_class;
 static guint clicked_signal;
 
@@ -136,14 +157,23 @@ static void cell_renderer_break_icon_set_property (GObject *object, guint param_
 /*
  * get size of a cell
  */
+#if GTK_CHECK_VERSION(3, 0, 0)
+static void cell_renderer_break_icon_get_size(GtkCellRenderer *cell, GtkWidget *widget, const GdkRectangle *cell_area, 
+	gint *x_offset, gint *y_offset, gint *width, gint *height)
+#else
 static void cell_renderer_break_icon_get_size(GtkCellRenderer *cell, GtkWidget *widget, GdkRectangle *cell_area, 
 	gint *x_offset, gint *y_offset, gint *width, gint *height)
+#endif
 {
 	CellRendererBreakIcon *cellbreakpoint = (CellRendererBreakIcon *) cell;
 	gint pixbuf_width  = 0;
 	gint pixbuf_height = 0;
 	gint calc_width;
 	gint calc_height;
+	gint xpad;
+	gint ypad;
+	gfloat xalign;
+	gfloat yalign;
 	
 	if (cellbreakpoint->pixbuf_enabled)
 	{
@@ -166,21 +196,24 @@ static void cell_renderer_break_icon_get_size(GtkCellRenderer *cell, GtkWidget *
 		pixbuf_height = MAX (pixbuf_height, gdk_pixbuf_get_height (cellbreakpoint->pixbuf_file));
 	}
 	
-	calc_width  = (gint) cell->xpad * 2 + pixbuf_width;
-	calc_height = (gint) cell->ypad * 2 + pixbuf_height;
+	gtk_cell_renderer_get_padding(cell, &xpad, &ypad);
+	calc_width  = xpad * 2 + pixbuf_width;
+	calc_height = ypad * 2 + pixbuf_height;
+
+	gtk_cell_renderer_get_alignment(cell, &xalign, &yalign);
 	
 	if (cell_area && pixbuf_width > 0 && pixbuf_height > 0)
 	{
 		if (x_offset)
 		{
 			*x_offset = (((gtk_widget_get_direction (widget) == GTK_TEXT_DIR_RTL) ?
-				(1.0 - cell->xalign) : cell->xalign) * 
+				(1.0 - xalign) : xalign) * 
 				(cell_area->width - calc_width));
 			*x_offset = MAX (*x_offset, 0);
 		}
 		if (y_offset)
 		{
-			*y_offset = (cell->yalign * (cell_area->height - calc_height));
+			*y_offset = (yalign * (cell_area->height - calc_height));
 			*y_offset = MAX (*y_offset, 0);
 		}
 	}
@@ -200,8 +233,13 @@ static void cell_renderer_break_icon_get_size(GtkCellRenderer *cell, GtkWidget *
 /*
  * render a cell
  */
+#if GTK_CHECK_VERSION(3, 0, 0)
+static void cell_renderer_break_icon_render(GtkCellRenderer *cell, cairo_t *cr, GtkWidget *widget,
+	const GdkRectangle *background_area, const GdkRectangle *cell_area, GtkCellRendererState flags)
+#else
 static void cell_renderer_break_icon_render(GtkCellRenderer *cell, GdkDrawable *window, GtkWidget *widget,
 	GdkRectangle *background_area, GdkRectangle *cell_area, GdkRectangle *expose_area, GtkCellRendererState flags)
+#endif
 {
 	CellRendererBreakIcon *cellbreakpoint = (CellRendererBreakIcon*) cell;
 	
@@ -209,7 +247,15 @@ static void cell_renderer_break_icon_render(GtkCellRenderer *cell, GdkDrawable *
 	
 	GdkRectangle pix_rect;
 	GdkRectangle draw_rect;
+
+	gint xpad;
+	gint ypad;
+
+	gboolean is_expander;
+
+#if !GTK_CHECK_VERSION(3, 0, 0)
 	cairo_t *cr;
+#endif
 	
 	cell_renderer_break_icon_get_size (cell, widget, cell_area,
 		&pix_rect.x,
@@ -217,16 +263,23 @@ static void cell_renderer_break_icon_render(GtkCellRenderer *cell, GdkDrawable *
 		&pix_rect.width,
 		&pix_rect.height);
 	
-	pix_rect.x += cell_area->x + cell->xpad;
-	pix_rect.y += cell_area->y + cell->ypad;
-	pix_rect.width  -= cell->xpad * 2;
-	pix_rect.height -= cell->ypad * 2;
-	
+	gtk_cell_renderer_get_padding(cell, &xpad, &ypad);
+	pix_rect.x += cell_area->x + xpad;
+	pix_rect.y += cell_area->y + ypad;
+	pix_rect.width  -= xpad * 2;
+	pix_rect.height -= ypad * 2;
+
+#if GTK_CHECK_VERSION(3, 0, 0)
+	if (!gdk_rectangle_intersect (cell_area, &pix_rect, &draw_rect))
+#else
 	if (!gdk_rectangle_intersect (cell_area, &pix_rect, &draw_rect) ||
 		!gdk_rectangle_intersect (expose_area, &draw_rect, &draw_rect))
+#endif
 		return;
+
+	g_object_get(cell, "is-expander", &is_expander, NULL);
 	
-	if (cell->is_expander)
+	if (is_expander)
 	{
 		pixbuf = cellbreakpoint->pixbuf_file;
 	}
@@ -246,20 +299,29 @@ static void cell_renderer_break_icon_render(GtkCellRenderer *cell, GdkDrawable *
 	if (!pixbuf)
 		return;
 	
+#if !GTK_CHECK_VERSION(3, 0, 0)
 	cr = gdk_cairo_create (window);
+#endif
 	
 	gdk_cairo_set_source_pixbuf (cr, pixbuf, pix_rect.x, pix_rect.y);
 	gdk_cairo_rectangle (cr, &draw_rect);
 	cairo_fill (cr);
 	
+#if !GTK_CHECK_VERSION(3, 0, 0)
 	cairo_destroy (cr);
+#endif
 }
 
 /*
  * activate callback
  */
+#if GTK_CHECK_VERSION(3, 0, 0)
+static gint cell_renderer_break_icon_activate(GtkCellRenderer *cell, GdkEvent *event, GtkWidget *widget, const gchar *path,
+	const GdkRectangle *background_area, const GdkRectangle *cell_area, GtkCellRendererState flags)
+#else
 static gint cell_renderer_break_icon_activate(GtkCellRenderer *cell, GdkEvent *event, GtkWidget *widget, const gchar *path,
 	GdkRectangle *background_area, GdkRectangle *cell_area, GtkCellRendererState  flags)
+#endif
 {
 	if (!event ||
 		(
@@ -284,7 +346,7 @@ static void cell_renderer_break_icon_init (CellRendererBreakIcon *cell)
 	cell->condition = NULL;
 	cell->hitscount = 0;
 	
-	cell_renderer->mode = GTK_CELL_RENDERER_MODE_ACTIVATABLE;
+	g_object_set(cell_renderer, "mode", GTK_CELL_RENDERER_MODE_ACTIVATABLE, NULL);
 
 	cell->pixbuf_enabled = cell->pixbuf_disabled = cell->pixbuf_conditional = cell->pixbuf_file = 0;
 }
--- a/debugger/src/cell_renderers/cellrendererbreakicon.h
+++ b/debugger/src/cell_renderers/cellrendererbreakicon.h
@@ -36,27 +36,6 @@ G_BEGIN_DECLS
 typedef struct _CellRendererBreakIcon CellRendererBreakIcon;
 typedef struct _CellRendererBreakIconClass CellRendererBreakIconClass;
 
-struct _CellRendererBreakIcon
-{
-  GtkCellRenderer parent;
-
-  guint GSEAL (enabled);
-  const gchar* GSEAL(condition);
-  guint GSEAL (hitscount);
-
-  GdkPixbuf *GSEAL (pixbuf_enabled);
-  GdkPixbuf *GSEAL (pixbuf_disabled);
-  GdkPixbuf *GSEAL (pixbuf_conditional);
-  GdkPixbuf *GSEAL (pixbuf_file);
-};
-
-struct _CellRendererBreakIconClass
-{
-  GtkCellRendererClass parent_class;
-
-  void (*clicked)(CellRendererBreakIcon *cell_renderer_toggle, const gchar *path);
-};
-
 GType					cell_renderer_break_icon_get_type(void);
 GtkCellRenderer*		cell_renderer_break_icon_new (void);
 
--- a/debugger/src/cell_renderers/cellrendererframeicon.c
+++ b/debugger/src/cell_renderers/cellrendererframeicon.c
@@ -36,14 +36,36 @@ enum {
   PROP_ACTIVE_FRAME,
 };
 
+struct _CellRendererFrameIcon
+{
+  GtkCellRenderer parent;
+
+  guint active_frame;
+
+  GdkPixbuf *pixbuf_active;
+  GdkPixbuf *pixbuf_highlighted;
+};
+
+struct _CellRendererFrameIconClass
+{
+  GtkCellRendererClass parent_class;
+
+  void (*clicked)(CellRendererFrameIcon *cell_renderer_toggle, const gchar *path);
+};
+
 static gpointer parent_class;
 static guint clicked_signal;
 
 /*
  * activate callback
  */
+#if GTK_CHECK_VERSION(3, 0, 0)
+static gint cell_renderer_frame_icon_activate(GtkCellRenderer *cell, GdkEvent *event, GtkWidget *widget, const gchar *path,
+	const GdkRectangle *background_area, const GdkRectangle *cell_area, GtkCellRendererState  flags)
+#else
 static gint cell_renderer_frame_icon_activate(GtkCellRenderer *cell, GdkEvent *event, GtkWidget *widget, const gchar *path,
 	GdkRectangle *background_area, GdkRectangle *cell_area, GtkCellRendererState  flags)
+#endif
 {
 	if (!event ||
 		(
@@ -116,14 +138,23 @@ static void cell_renderer_frame_icon_set_property (GObject *object, guint param_
 /*
  * get size of a cell
  */
+#if GTK_CHECK_VERSION(3, 0, 0)
+static void cell_renderer_frame_icon_get_size(GtkCellRenderer *cell, GtkWidget *widget, const GdkRectangle *cell_area, 
+	gint *x_offset, gint *y_offset, gint *width, gint *height)
+#else
 static void cell_renderer_frame_icon_get_size(GtkCellRenderer *cell, GtkWidget *widget, GdkRectangle *cell_area, 
 	gint *x_offset, gint *y_offset, gint *width, gint *height)
+#endif
 {
 	CellRendererFrameIcon *cellframe = (CellRendererFrameIcon *) cell;
 	gint pixbuf_width  = 0;
 	gint pixbuf_height = 0;
 	gint calc_width;
 	gint calc_height;
+	gint xpad;
+	gint ypad;
+	gfloat xalign;
+	gfloat yalign;
 	
 	if (cellframe->pixbuf_active)
 	{
@@ -136,21 +167,24 @@ static void cell_renderer_frame_icon_get_size(GtkCellRenderer *cell, GtkWidget *
 		pixbuf_height = MAX (pixbuf_height, gdk_pixbuf_get_height (cellframe->pixbuf_highlighted));
 	}
 	
-	calc_width  = (gint) cell->xpad * 2 + pixbuf_width;
-	calc_height = (gint) cell->ypad * 2 + pixbuf_height;
+	gtk_cell_renderer_get_padding(cell, &xpad, &ypad);
+	calc_width  = xpad * 2 + pixbuf_width;
+	calc_height = ypad * 2 + pixbuf_height;
+
+	gtk_cell_renderer_get_alignment(cell, &xalign, &yalign);
 	
 	if (cell_area && pixbuf_width > 0 && pixbuf_height > 0)
 	{
 		if (x_offset)
 		{
 			*x_offset = (((gtk_widget_get_direction (widget) == GTK_TEXT_DIR_RTL) ?
-				(1.0 - cell->xalign) : cell->xalign) * 
+				(1.0 - xalign) : xalign) * 
 				(cell_area->width - calc_width));
 			*x_offset = MAX (*x_offset, 0);
 		}
 		if (y_offset)
 		{
-			*y_offset = (cell->yalign * (cell_area->height - calc_height));
+			*y_offset = (yalign * (cell_area->height - calc_height));
 			*y_offset = MAX (*y_offset, 0);
 		}
 	}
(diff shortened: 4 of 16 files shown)


elextr:
The maintainer of debugger is not very active lately, so anybody should feel free to review/test/comment otherwise this may be fairly slow to be added.

LarsGit223:
I'll try to have a look on it at the weekend.

debugger/src/dconfig.c
```diff
@@ -67,8 +67,13 @@ static GKeyFile *keyfile_project = NULL;
 static gboolean debug_config_loading = FALSE;
 
 /* saving thread staff */
+#if GTK_CHECK_VERSION(3, 0, 0)
+static GMutex change_config_mutex;
+static GCond cond;
+#else
```
LarsGit223: This is glib specific code and does not depend on gtk. So a check for the gtk version seems to be wrong here.

debugger/src/dconfig.c
```diff
@@ -269,8 +274,13 @@ static void save_to_keyfile(GKeyFile *keyfile)
  */
 static gpointer saving_thread_func(gpointer data)
 {
+#if GTK_CHECK_VERSION(3, 0, 0)
+	gint64 interval;
+	g_mutex_lock(&change_config_mutex);
+#else
```
LarsGit223: I assume this code could be used for all gtk versions, again, not gtk but glib dependent.

debugger/src/dconfig.c
```diff
@@ -308,11 +318,20 @@ static gpointer saving_thread_func(gpointer data)
 			debug_config_changed = FALSE;
 		}
 
+#if GTK_CHECK_VERSION(3, 0, 0)
+		interval = g_get_monotonic_time() + SAVING_INTERVAL;
+#else
```
LarsGit223: No gtk dependant code. It depends on the glib version. What is the reason for this change? If it is an improvement then the code could be used without any check as the function ```g_get_monotonic_time()``` exists since glib 2.28 and geany requires 2.32.

debugger/src/dconfig.c
```diff
@@ -308,11 +318,20 @@ static gpointer saving_thread_func(gpointer data)
...
 		g_get_current_time(&interval);
 		g_time_val_add(&interval, SAVING_INTERVAL);
+#endif
 	}
+#if GTK_CHECK_VERSION(3, 0, 0)
+	while (!g_cond_wait_until(&cond, &change_config_mutex, interval));
+	g_mutex_unlock(&change_config_mutex);
```
LarsGit223: This requires glib 2.32 at least, not gtk3.

LarsGit223: Sorry, I mean ```g_cond_wait_until()``` replacing ```g_cond_timed_wait```.

debugger/src/dconfig.c
```diff
@@ -458,21 +493,36 @@ void config_init(void)
 		g_free(data);
 	}
 
+#if GTK_CHECK_VERSION(3, 0, 0)
+	g_mutex_init(&change_config_mutex);
+	g_cond_init(&cond);
+	saving_thread = g_thread_new(NULL, saving_thread_func, NULL);
+#else
```
LarsGit223: Again, depends on glib 2.32 not gtk3.

debugger/src/debug.c
```diff
@@ -320,7 +324,11 @@ static void on_watch_dragged_callback(GtkWidget *wgt, GdkDragContext *context, i
     gpointer userdata)
 {
 	/* string that is dragged */
+#if GTK_CHECK_VERSION(3, 0, 0)
+	gchar *expression = (gchar*)gtk_selection_data_get_data(seldata);
```
LarsGit223: The function ```gtk_selection_data_get_data()``` exists since gtk 2.14 and so could be used for both verions.

debugger/src/debug.c
```diff
@@ -1066,7 +1105,13 @@ void debug_init(void)
 	configfile = g_strconcat(geany_data->app->configdir, G_DIR_SEPARATOR_S, "geany.conf", NULL);
 	g_key_file_load_from_file(config, configfile, G_KEY_FILE_NONE, NULL);
 	font = utils_get_setting_string(config, "VTE", "font", "Monospace 10");
+#if GTK_CHECK_VERSION(3, 0, 0)
+	fontdesc = pango_font_description_from_string(font);
+	vte_terminal_set_font(VTE_TERMINAL(terminal), fontdesc);
+	pango_font_description_free(fontdesc);
+#else
```
LarsGit223: If you like you can replace this lines with a call to function ```gp_vtecompat_set_font_from_string()``` from the geany-plugins util lib. Requires ```#include <gp_vtecompat.h>```.

debugger/src/dpaned.c
```diff
@@ -287,8 +288,13 @@ void dpaned_init(void)
 	/* setup notebooks */
 	gtk_notebook_set_scrollable(GTK_NOTEBOOK(debug_notebook_left), TRUE);
 	gtk_notebook_set_scrollable(GTK_NOTEBOOK(debug_notebook_right), TRUE);
+#if GTK_CHECK_VERSION(3, 0, 0)
+	gtk_notebook_set_group_name(GTK_NOTEBOOK(debug_notebook_left), NOTEBOOK_GROUP_NAME(NOTEBOOK_GROUP));
+	gtk_notebook_set_group_name(GTK_NOTEBOOK(debug_notebook_right), NOTEBOOK_GROUP_NAME(NOTEBOOK_GROUP));
+#else
```
LarsGit223: It seems like the code above should be used for both versions gtk2 and 3 as ```gtk_notebook_set_group_id()``` is deprecated since version 2.10.

debugger/src/stree.c
```diff
@@ -197,7 +197,11 @@ static void on_render_line(GtkTreeViewColumn *tree_column, GtkCellRenderer *cell
 		g_object_set(cell, "text", "", NULL);
 	else
 	{
+#if GTK_CHECK_VERSION(3, 0, 0)
+		GValue value = G_VALUE_INIT;
```
LarsGit223: Depends on glib 2.30, not on gtk3.

debugger/src/tpage.c
```diff
@@ -161,23 +166,41 @@ void tpage_pack_widgets(gboolean tabbed)
...
+#endif
 		gtk_box_pack_start(GTK_BOX(root), hbox, FALSE, FALSE, 0);
 		gtk_box_pack_start(GTK_BOX(hbox), target_label, FALSE, FALSE, 0);
 		gtk_box_pack_start(GTK_BOX(hbox), target_name, TRUE, TRUE, 0);
 		gtk_box_pack_start(GTK_BOX(hbox), target_button_browse, FALSE, FALSE, 0);
 		
 		/* lower hbox */
+#if GTK_CHECK_VERSION(3, 0, 0)
+		hbox = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, SPACING);
+		gtk_box_set_homogeneous(GTK_BOX(hbox), TRUE);
```
LarsGit223: Looks like ```gtk_box_set_homogeneous()``` could be used in both gtk versions?

debugger/src/tpage.c
```diff
@@ -196,27 +223,45 @@ void tpage_pack_widgets(gboolean tabbed)
 	{
 		GtkWidget *lbox, *rbox, *hbox;
 
+#if GTK_CHECK_VERSION(3, 0, 0)
+		root = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, SPACING);
+		gtk_box_set_homogeneous(GTK_BOX(root), TRUE);
```
LarsGit223: Here also the line could be used for both versions maybe.

debugger/src/tpage.c
```diff
@@ -250,26 +299,46 @@ static void tpage_create_widgets(void)
...
 	gtk_entry_set_editable(GTK_ENTRY(target_name), FALSE);
 	target_button_browse = create_stock_button(GTK_STOCK_OPEN, _("Browse"));
+#endif
 	gtk_widget_set_size_request(target_button_browse, BROWSE_BUTTON_WIDTH, 0);
 	g_signal_connect(G_OBJECT(target_button_browse), "clicked", G_CALLBACK (on_target_browse_clicked), NULL);
 
 	/* debugger */
 	debugger_label = gtk_label_new(_("Debugger:")); 
+#if GTK_CHECK_VERSION(3, 0, 0)
+	debugger_cmb = gtk_combo_box_text_new();
```
LarsGit223: The function ```gtk_combo_box_text_new()``` exists since gtk version 2.24 so we could use that code for both verions (```gtk_combo_box_new_text()``` is deprecated since 2.4).

debugger/src/tpage.c
```diff
@@ -250,26 +299,46 @@ static void tpage_create_widgets(void)
...
+#else
 	debugger_cmb = gtk_combo_box_new_text();
+#endif
 	modules = debug_get_modules();
 	for (iter = modules; iter; iter = iter->next)
 	{
+#if GTK_CHECK_VERSION(3, 0, 0)
+		gtk_combo_box_text_append_text(GTK_COMBO_BOX_TEXT(debugger_cmb), (gchar*)iter->data);
```
LarsGit223: The function ```gtk_combo_box_text_append_text()``` exists since gtk version 2.24 so we could use that code for both verions (```gtk_combo_box_append_text()``` is deprecated since 2.4).

debugger/src/tpage.c
```diff
@@ -366,7 +442,11 @@ int tpage_get_debug_module_index(void)
  */
 gchar* tpage_get_debugger(void)
 {
+#if GTK_CHECK_VERSION(3, 0, 0)
+	return gtk_combo_box_text_get_active_text(GTK_COMBO_BOX_TEXT(debugger_cmb));
```
LarsGit223: The function ```gtk_combo_box_text_get_active_text()``` exists since gtk version 2.24 so we could use that code for both versions (```gtk_combo_box_get_active_text()``` is deprecated since 2.6).

debugger/src/btnpanel.c
```diff
@@ -82,17 +82,28 @@ static void on_execute_until(GtkButton *button, gpointer user_data)
  */
 GtkWidget* btnpanel_create(on_toggle cb)
 {
+#if GTK_CHECK_VERSION(3, 0, 0)
+	GtkWidget *vbox = gtk_box_new(GTK_ORIENTATION_VERTICAL, CP_BUTTONS_PAD);
+
```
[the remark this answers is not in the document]
bxgaillard: I agree. Fixed.

debugger/src/cell_renderers/cellrendererbreakicon.c
```diff
@@ -166,21 +177,37 @@ static void cell_renderer_break_icon_get_size(GtkCellRenderer *cell, GtkWidget *
...
+#if GTK_CHECK_VERSION(3, 0, 0)
+	gtk_cell_renderer_get_padding(cell, &xpad, &ypad);
+	calc_width  = xpad * 2 + pixbuf_width;
+	calc_height = ypad * 2 + pixbuf_height;
+
+	gtk_cell_renderer_get_alignment(cell, &xalign, &yalign);
+#else
 	calc_width  = (gint) cell->xpad * 2 + pixbuf_width;
 	calc_height = (gint) cell->ypad * 2 + pixbuf_height;
+#endif
```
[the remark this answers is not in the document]
bxgaillard: I agree. Fixed.

debugger/src/cell_renderers/cellrendererbreakicon.c
```diff
@@ -200,23 +227,51 @@ static void cell_renderer_break_icon_get_size(GtkCellRenderer *cell, GtkWidget *
...
 	cell_renderer_break_icon_get_size (cell, widget, cell_area,
 		&pix_rect.x,
 		&pix_rect.y,
 		&pix_rect.width,
 		&pix_rect.height);
 	
+#if GTK_CHECK_VERSION(3, 0, 0)
+	gtk_cell_renderer_get_padding(cell, &xpad, &ypad);
+	pix_rect.x += cell_area->x + xpad;
+	pix_rect.y += cell_area->y + ypad;
```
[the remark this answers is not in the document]
bxgaillard: I agree. Fixed.

debugger/src/cell_renderers/cellrendererframeicon.c
```diff
@@ -136,21 +152,37 @@ static void cell_renderer_frame_icon_get_size(GtkCellRenderer *cell, GtkWidget *
 		pixbuf_height = MAX (pixbuf_height, gdk_pixbuf_get_height (cellframe->pixbuf_highlighted));
 	}
 	
+#if GTK_CHECK_VERSION(3, 0, 0)
+	gtk_cell_renderer_get_padding(cell, &xpad, &ypad);
+	calc_width  = xpad * 2 + pixbuf_width;
+	calc_height = ypad * 2 + pixbuf_height;
+
+	gtk_cell_renderer_get_alignment(cell, &xalign, &yalign);
+#else
```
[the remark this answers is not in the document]
bxgaillard: I agree. Fixed.

debugger/src/cell_renderers/cellrendererframeicon.c
```diff
@@ -170,23 +202,44 @@ static void cell_renderer_frame_icon_get_size(GtkCellRenderer *cell, GtkWidget *
...
 		&pix_rect.x,
 		&pix_rect.y,
 		&pix_rect.width,
 		&pix_rect.height);
 	
+#if GTK_CHECK_VERSION(3, 0, 0)
+	gtk_cell_renderer_get_padding(cell, &xpad, &ypad);
+	pix_rect.x += cell_area->x + xpad;
+	pix_rect.y += cell_area->y + ypad;
+	pix_rect.width  -= xpad * 2;
```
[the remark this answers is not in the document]
bxgaillard: I agree. Fixed.

debugger/src/dconfig.c
bxgaillard: I agree. I replaced the deprecated glib code with the new one.

debugger/src/dconfig.c
bxgaillard: I agree. I replaced the deprecated glib code with the new one.

debugger/src/dconfig.c
bxgaillard: This change has been made to use the new `g_cond_wait_until()` function which replaces the deprecated `g_cond_timed_wait()` one. The old code has been replaced with the new one.

debugger/src/dconfig.c
bxgaillard: Fixed: the old code has been replaced with the new one.

debugger/src/dconfig.c
bxgaillard: I agree. Fixed.

debugger/src/debug.c
bxgaillard: I agree. Fixed.

debugger/src/debug.c
bxgaillard: I used the common utils library, thanks for the suggestion.

debugger/src/dpaned.c
bxgaillard: I agree. Fixed.

debugger/src/stree.c
bxgaillard: Fixed: old code replaced with new one.

debugger/src/tpage.c
bxgaillard: Maybe it could, but `gtk_box_set_homogeneous()` is used because of the missing parameter in `gtk_box_new()`. Using it in the GTK2 version would be redundant.

debugger/src/tpage.c
bxgaillard: Same as above : it would be redundant in the GTK2 code.

debugger/src/tpage.c
bxgaillard: I agree. Fixed.

debugger/src/tpage.c
bxgaillard: I agree. Fixed.

debugger/src/tpage.c
bxgaillard: I agree. Fixed.

bxgaillard:
Thank you @LarsGit223 for the thorough PR review! I'm not a GTK+ expert at all, that's why I didn'y dive in the documentation and just applied the changes from #645 with preprocessing conditionals. As indicated in my replies I've applied your suggestions and pushed a new commit which should be more adequate.

debugger/src/tpage.c
LarsGit223: Ok, then let's keep it as-is.

debugger/src/tpage.c
LarsGit223: Ok, let's keep it to.

LarsGit223:
> I'm not a GTK+ expert at all, that's why I didn'y dive in the documentation and just applied the changes from #645 with preprocessing conditionals.

Therefore thanks for putting time in it even though it wasn't your PR originally :+1: Once the others agree to merge it, the only thing left to do would be to squash the commits into one.

bxgaillard:
> Therefore thanks for putting time in it even though it wasn't your PR originally :+1:

My pleasure.

> Once the others agree to merge it, the only thing left to do would be to squash the commits into one.

Done.

elextr:
@b4n @frlan are you going to wait for the maintainer or will you merge for 1.34?

b4n:
I did work on the debugger plugin and am willing to look at this, but I won't have time for 1.34

elextr:
Oh well, no debugger for Debian and descendants with 1.34.  Note that @LarsGit223 reviewed and presumably tested and he is scope maintainer.

bxgaillard:
An IDE without a debugger, blocked in this state for the lifetime of the next stable Debian release, that would be very sad.

LarsGit223:
@elextr, @b4n: Yes, I reviewed this. I have to admit that I never wrote a debugger myself. But the things changed here only apply to the GUI, not to the debugger related internals.

I tested the following:
- building under Ubuntu with GTK2 and 3
- simple test under Ubuntu with GTK2 and 3: run a program in the debugger, have some breakpoints. Inspect some variables. I clicked some icons and dialogs, nothing seems to be broken.
- I did not test the GUI extensively (that means, I did **not** make a checkpoint list of all debugger related icons and menus and tested each of it)
- I did not test or build under Windows or Mac
- I had a look at the GTK API calls and the GTK versions for which they are supported and suggested changes (see the review for details)
- Changes were applied and I re-tested and re-viewed again

debugger/src/cell_renderers/cellrendererbreakicon.c
```diff
@@ -200,33 +212,56 @@ static void cell_renderer_break_icon_get_size(GtkCellRenderer *cell, GtkWidget *
...
+	if (!gdk_rectangle_intersect (cell_area, &pix_rect, &draw_rect))
+#else
 	if (!gdk_rectangle_intersect (cell_area, &pix_rect, &draw_rect) ||
 		!gdk_rectangle_intersect (expose_area, &draw_rect, &draw_rect))
+#endif
 		return;
+
+	g_object_get_property(G_OBJECT(cell), "is-expander", &is_expander_value);
+	is_expander = g_value_get_boolean(&is_expander_value);
+	g_value_unset(&is_expander_value);
```
b4n: You should probably just use the simpler version: `g_object_get(cell, "is-expander", &is_expander, NULL)`

debugger/src/cell_renderers/cellrendererbreakicon.c
```diff
@@ -284,7 +328,11 @@ static void cell_renderer_break_icon_init (CellRendererBreakIcon *cell)
 	cell->condition = NULL;
 	cell->hitscount = 0;
 	
+#if GTK_CHECK_VERSION(3, 0, 0)
+	g_object_set(cell_renderer, "mode", GTK_CELL_RENDERER_MODE_ACTIVATABLE, NULL);
+#else
 	cell_renderer->mode = GTK_CELL_RENDERER_MODE_ACTIVATABLE;
+#endif
```
b4n: Here the GTK3 version would work on GTK2 too

debugger/src/cell_renderers/cellrendererbreakicon.h
```diff
@@ -40,14 +40,14 @@ struct _CellRendererBreakIcon
...
+  guint hitscount;
 
-  GdkPixbuf *GSEAL (pixbuf_enabled);
-  GdkPixbuf *GSEAL (pixbuf_disabled);
-  GdkPixbuf *GSEAL (pixbuf_conditional);
-  GdkPixbuf *GSEAL (pixbuf_file);
+  GdkPixbuf *pixbuf_enabled;
+  GdkPixbuf *pixbuf_disabled;
+  GdkPixbuf *pixbuf_conditional;
+  GdkPixbuf *pixbuf_file;
```
b4n: I guess this would ideally use a private member: the goal of `GSEAL()` during GTK3 transition preparation was to "hide" members to callers by mangling their name.  The "correct" solution is to use a private, opaque, structure.

debugger/src/cell_renderers/cellrendererframeicon.h
```diff
@@ -40,10 +40,10 @@ struct _CellRendererFrameIcon
 {
   GtkCellRenderer parent;
 
-  guint GSEAL (active_frame);
+  guint active_frame;
 
-  GdkPixbuf *GSEAL (pixbuf_active);
-  GdkPixbuf *GSEAL (pixbuf_highlighted);
+  GdkPixbuf *pixbuf_active;
+  GdkPixbuf *pixbuf_highlighted;
```
b4n: Same for private

debugger/src/debug.c
```diff
@@ -1048,12 +1068,24 @@ void debug_init(void)
...
+#if GTK_CHECK_VERSION(3, 0, 0)
+	pty = vte_pty_new_foreign_sync(pty_master, NULL, NULL);
+	vte_terminal_set_pty(VTE_TERMINAL(terminal), pty);
+	g_object_unref(pty);
+	scrollbar = gtk_scrollbar_new(GTK_ORIENTATION_VERTICAL, gtk_scrollable_get_vadjustment(GTK_SCROLLABLE(terminal)));
+	gtk_widget_set_can_focus(GTK_WIDGET(scrollbar), FALSE);
+#else
 	vte_terminal_set_pty(VTE_TERMINAL(terminal), pty_master);
 	scrollbar = gtk_vscrollbar_new(GTK_ADJUSTMENT(VTE_TERMINAL(terminal)->adjustment));
 	GTK_WIDGET_UNSET_FLAGS(scrollbar, GTK_CAN_FOCUS);
```
b4n: This line could be using the GTK3 version

debugger/src/tpage.c
```diff
@@ -250,26 +295,38 @@ static void tpage_create_widgets(void)
...
 	g_list_free(modules);
 	gtk_combo_box_set_active(GTK_COMBO_BOX(debugger_cmb), 0);
 
 	/* arguments */
 	args_frame = gtk_frame_new(_("Command Line Arguments"));
+#if GTK_CHECK_VERSION(3, 0, 0)
+	hbox = gtk_scrolled_window_new(
+		gtk_scrollable_get_hadjustment(GTK_SCROLLABLE(args_textview)),
+		gtk_scrollable_get_vadjustment(GTK_SCROLLABLE(args_textview)));
+	gtk_scrolled_window_set_policy(GTK_SCROLLED_WINDOW(hbox), GTK_POLICY_AUTOMATIC, GTK_POLICY_AUTOMATIC);
```
b4n: This is weird to be so different between versions, if the view should be in a scrolled window, it probably should be in any case, even if it works OK with GTK2 but shows a GTK3 bug.

Also, `args_textview` seems ti be used before assignment a few lines down.

debugger/src/tpage.c
```diff
@@ -279,8 +336,15 @@ static void tpage_create_widgets(void)
 
 	/* environment */
 	env_frame = gtk_frame_new(_("Environment Variables"));
+#if GTK_CHECK_VERSION(3, 0, 0)
+	hbox = gtk_scrolled_window_new(
+		gtk_scrollable_get_hadjustment(GTK_SCROLLABLE(args_textview)),
+		gtk_scrollable_get_vadjustment(GTK_SCROLLABLE(args_textview)));
+	gtk_scrolled_window_set_policy(GTK_SCROLLED_WINDOW(hbox), GTK_POLICY_AUTOMATIC, GTK_POLICY_AUTOMATIC);
```
b4n: same here, and using the adjustment from `args_textview` seem wrong here, shouldn't it be `tree`?

debugger/src/cell_renderers/cellrendererbreakicon.h
elextr: Whats the point of complicating things by making them "private"? This is a debugger plugin, not a library that needs to be protected from being used by millions.

b4n: Yeah it probably doesn't matter much here indeed.  I just mentioned that because the code previously used `GSEAL()` that's all.

debugger/src/cell_renderers/cellrendererbreakicon.c
bxgaillard: You're right, changed.

debugger/src/cell_renderers/cellrendererbreakicon.c
bxgaillard: Fixed.

debugger/src/cell_renderers/cellrendererbreakicon.h
bxgaillard: Yes I didn't understand the reason to use GSEAL() in the first place in the GTK+2 code. It just broke the GTK+3 version.
As a more elegant solution to the private nature of these members, I moved the structure definitions out of the .h into the .c.

debugger/src/cell_renderers/cellrendererframeicon.h
bxgaillard: Same answer as above.

debugger/src/tpage.c
bxgaillard: The code can't be the same between GTK+2 and GTK+3 because the GtkTextView (and the GtkTreeView) don't implement the GtkScrollable interface in GTK+2 whereas they do in GTK+3 and shall seemingly be handled with a GtkScrolledWindow to add the scrolling functionality from what I can understand.

However you're absolutely right about the use before assignment, this is fixed.

debugger/src/tpage.c
bxgaillard: Same as above. And the correct variable is effectively tree, this is fixed.

debugger/src/debug.c
bxgaillard: If you're talking about the use of `gtk_widget_set_can_focus()`, that's right and it's fixed.
The remaining code about PTY initialization seemingly has to be different though.

bxgaillard:
Thank you @b4n for your review. I applied your proposals, squashed the commits and pushed it.

debugger/src/tpage.c
b4n: Be it in GTK2 or GTK3, to actually be scrollable you need to put most widgets in a GtkScrolledWindow.  The widgets "supporting scrolling" like GtkTreeView and GtkTextView know how to communicate with the ScrolledWindow.  On GTK3 it's a proper interface; on GTK2 it was signals only.
Basically, what you usually would do is simply create a ScrolledWindow with passing `NULL` as the adjustments, and let GTK do it's thing when you add the child widget.

So, basically you could just use `gtk_scrolled_window_new(NULL, NULL)` followed by `gtk_container_add()` in bot GTK2 and GTK3.

debugger/src/debug.c
b4n: Yes I'm talking about `gtk_widget_set_can_focus()` I don't really know about the PTY stuff, but it seem that it has indeed to be different with VTE2.91.

debugger/src/tpage.c
bxgaillard: Thank you for taking the time to explain this to me. I changed the corresponding code accordingly, as well as some other code which used the same construction for fewer differences between GTK+2 and GTK+3 in debug.c.

andy5995:
I can test this on Debian buster and OpenBSD 6.4. Let me know if that's needed.

elextr:
@andy5995 testing is always appreciated

bxgaillard:
Hi @andy5995,
The problem you mention is with utilslib, which is part of geany-plugins and now used by the debugger plugin (it contains some compatibility functions). Did you activate it at configure time? It may need some other dependencies besides libvte.

LarsGit223:
I noticed the line ```...PTY_LIBS="-lutil"]``` in ```scope.m4```. Do we need this in ```debugger.m4``` also?

andy5995:
> Hi @andy5995,
> The problem you mention is with utilslib, which is part of geany-plugins and now used by the debugger plugin (it contains some compatibility functions). Did you activate it at configure time?

[config.log](https://github.com/geany/geany-plugins/files/2788940/config.log)

I only tried activating it after you told me about it ;) But I got the same result. But now I wonder if the test failing is related to the error mentioned on L15275.

```
checking whether the linker understands -Wl,-z,defs... yes
./configure[15275]: test: 3: unexpected operator/operand
checking for UTILSLIB... no
configure: error: Plugin Debugger depends on utilslib VTE support
```

I looked but couldn't find the problem.

andy5995:
I installed the vte-0.28.2p19 package and that problem was solved. New problem is that the built fails with an undefined reference error:

https://gist.github.com/andy5995/0b0fab05c70336bfff86ac4a3354d5a6#file-gistfile1-txt-L59

[config.log](https://github.com/geany/geany-plugins/files/2789099/config.log)

Could be related to @LarsGit223 's comment, as I see that function seems to be defined in the utility library.

I tried preceding the configure command with PTY_LIBS="-lutil" and LDADD"-lutil" but no effect.

b4n:
> `./configure[15275]: test: 3: unexpected operator/operand`

Could you paste your generated `configure`?  The error probably comes from there and a `test` call that doesn't please your shell

andy5995:
> > `./configure[15275]: test: 3: unexpected operator/operand`
> 
> Could you paste your generated `configure`? The error probably comes from there and a `test` call that doesn't please your shell

Sure...

https://gist.github.com/andy5995/fb2b16eb9dab45d9ab8160cdd6f2a48e

Also noticed it happening with other tests as well

```
checking libutil.h presence... no
checking for libutil.h... no
checking for glib-genmarshal... /usr/local/bin/glib-genmarshal
checking for glib-mkenums... /usr/local/bin/glib-mkenums
checking for rst2html... /usr/local/bin/rst2html
./configure[18255]: test: 3: unexpected operator/operand
checking for GTKSPELL... no
checking for gpgme-config... no
checking for special C compiler options needed for large files... no
checking for _FILE_OFFSET_BITS value needed for large files... no
./configure[19681]: test: 3: unexpected operator/operand
./configure[21201]: test: 3: unexpected operator/operand
checking for util.h... (cached) yes
checking for pty.h... (cached) no
checking for libutil.h... (cached) no
checking for ENCHANT_2_2... no
checking for ENCHANT_1_5... no
checking for ENCHANT_2_0... no
checking for glib-mkenums... (cached) /usr/local/bin/glib-mkenums
./configure[22250]: test: 3: unexpected operator/operand
```

b4n:
Apparently we've got a portability issue with the GTK version check.  @andy5995 could you check out #815 and see if it helps?  It should get rid of the *"unexpected operator/operand"* issues, and hopefully maybe more.

andy5995:
#815 fixed the build issue and configure warnings on OpenBSD.

andy5995:
One problem I noticed on Debian Buster and Slackware-current: when the Debugger plug-in is enabled, the "down arrow" disappears (screenshot below).

image

andy5995:
My debug info from Buster:

```
15:36:41: Geany INFO		: Using alternate configuration directory
15:36:41: Geany INFO		: Geany 1.35 (git >= bf5c9edd), en_US.utf8
15:36:41: Geany INFO		: GTK 3.24.3, GLib 2.58.2
15:36:41: Geany INFO		: System data dir: /home/andy/test-geany/share/geany
15:36:41: Geany INFO		: User config dir: /home/andy/test-geany/config
15:36:41: Geany INFO		: Loaded GTK+ CSS theme '/home/andy/test-geany/share/geany/geany.css'
15:36:41: Geany INFO		: Loaded GTK+ CSS theme '/home/andy/test-geany/share/geany/geany-3.20.css'
15:36:41: Geany INFO		: System plugin path: /home/andy/test-geany/lib/geany
15:36:41: Geany INFO		: Added filetype Clojure (61).
15:36:41: Geany INFO		: Added filetype Graphviz (62).
15:36:41: Geany INFO		: Added filetype CUDA (63).
15:36:41: Geany INFO		: Added filetype JSON (64).
15:36:41: Geany INFO		: Added filetype Arduino (65).
15:36:41: Geany INFO		: Added filetype Scala (66).
15:36:41: Geany INFO		: Added filetype Genie (67).
15:36:41: Geany INFO		: Added filetype Cython (68).
15:36:41: Geany INFO		: Loaded libvte from libvte-2.91.so
15:36:41: Geany INFO		: Loaded:   /home/andy/test-geany/lib/geany/debugger.so (Debugger)
15:36:41: Geany INFO		: unknown : None (UTF-8)
15:36:49: GdkPixbuf DEBUG	: gdk_pixbuf_from_pixdata() called on:
15:36:49: GdkPixbuf DEBUG	: 	Encoding raw
15:36:49: GdkPixbuf DEBUG	: 	Dimensions: 14 x 14
15:36:49: GdkPixbuf DEBUG	: 	Rowstride: 56, Length: 808
15:36:49: GdkPixbuf DEBUG	: 	Copy pixels == false
15:36:49: GdkPixbuf DEBUG	: gdk_pixbuf_from_pixdata() called on:
15:36:49: GdkPixbuf DEBUG	: 	Encoding raw
15:36:49: GdkPixbuf DEBUG	: 	Dimensions: 14 x 14
15:36:49: GdkPixbuf DEBUG	: 	Rowstride: 56, Length: 808
15:36:49: GdkPixbuf DEBUG	: 	Copy pixels == false
15:36:49: GdkPixbuf DEBUG	: gdk_pixbuf_from_pixdata() called on:
15:36:49: GdkPixbuf DEBUG	: 	Encoding raw
15:36:49: GdkPixbuf DEBUG	: 	Dimensions: 14 x 14
15:36:49: GdkPixbuf DEBUG	: 	Rowstride: 56, Length: 808
15:36:49: GdkPixbuf DEBUG	: 	Copy pixels == false
```

andy5995:
> One problem I noticed on Debian Buster and Slackware-current: when the Debugger plug-in is enabled, the "down arrow" disappears 

Does that reproduce on Ubuntu as well? @LarsGit223 ?

LarsGit223:
@andy5995: Yes it does.

Tested with your repository branch ```debugger-gtk3```:
- Ubuntu 18.04.02, gtk 3.22.30, glib 2.56.3: the arrow is missing
- Ubuntu 18.04.02, gtk 2.24.32, glib 2.56.3: the arrow is showing

Just in case you did not notice because "the down arrow is missing" does not describe the problem I noticed very good. I would say enabling the ```debugger``` plugin on gtk3 causes the menu bar in the message window to have a fixed size. It is not shrinkable/growable any more. So the arrows are gone. Also the menu items are not reduced any more if the height of the message window is too small (I mean that usually e.g. only two menu items are shown if there is only space for two and then you can use the arrows to reach the other buttons). Disabling the ```debugger``` plugin immediately brings back the usual behavior.

LarsGit223:
I had another look at the changes made by the PR but I cannot find the reason why the re-sizing of the message window does not work as expected under gtk3 if the debugger plugin is enabled.

bxgaillard:
I've looked into it and from what I can tell this is normal behaviour.

The main Geany GUI uses GtkPaneds for side panels (document browser, symbols, messages...), including the bottom one. When such a panel is too small to display its contents, the latter is located in hidden overflow.

This also happens for the left panel if you resize it: try to drag the splitter to the left, you'll see that the left arrow will slide under the left border of the main window. The exact same thing happens with the debugger plugin if the botton panel becomes too small. Or put more exactly: too small for the right button panel to fit un the bottom space.

So IMHO this "bug" is unrelated to the debugger plugin but is a general Geany GUI issue.

I can see two possible resolutions:
1. Fix the main Geany GUI so that the side panels don't overflow either by enabling an automatic scrollbar or by specifing a minimum size the panels can't be shrink under (which I think is the best one);
2. Add a scrollbar specific to the whole debugger plugin panel, with possible ergonomic issues.

I also checked the GTK+2 behaviour : it simply doesn't honor the plugins minimum panel size. When you shrink the panel so that it becames too small, the controls collapse one above another and the GUI isn't quite usable anymore. This isn't better IMHO.

BTW, on an unrelated note, I'm going to rebase the branch on more current code. No other change beyond that.

elextr:
@bxgaillard have you tried setting sizes and resize and shrink and stuff from CSS, setting minimum sizes in apps is not very portable.  Note also the comment from @LarsGit223 that showed the behaviour was different on different versions of gtk. which did you test on?

bxgaillard:
> @bxgaillard have you tried setting sizes and resize and shrink and stuff from CSS, setting minimum sizes in apps is not very portable.
I have made some research and CSS seems not to be appropriate for setting dimensions. On the other hand the correct way to fix it seems to be the Glade file. Changing <property name="shrink">True</property> to False (Geany release 1.34.1, geany.glade line 8452) doesn't allow overflow for the bottom paned for instance. But as I said just before this has to be fixed in the main Geany repository and has nothing to do with geany-plugins.

> Note also the comment from @LarsGit223 that showed the behaviour was different on different versions of gtk. which did you test on?
I tested on both: with GTK+3 the hidden overflow happens and with GTK+2 the controls collapse (they superpose themselves) and the right button panel from the debugger plugin overflows (instead of the whole bottom panel).

LarsGit223:
@elextr, @bxgaillard: if the behavior is really caused by a different behavior of gtk2 and gtk3 then it is OK for me. As I wrote I couldn't find anything in the code between the two versions which looked like a porting fault. Also, if someone enables the debugger plugin it doesn't really make sense to make the message window that small that you recognize this issue/difference in behavior.

LarsGit223:
This didn't make it into 1.35 but maybe we can put it to milestone 1.36. As written earlier the little overflow problem is acceptable to me.

frlan:
Well...let's go forward.

AnishGRao:
As someone who enjoys using Geany, but is quite new to Linux in general, how would I be able to use this debugger on my machine? I have noted that I cannot use it, and tracked down the issue to here, but am unsure as to how I would use this, or even where to download these plugins from. Thanks for your time, and sorry for any inconvenience.

elextr:
When you post to github or mailing lists (which is where you should ask for help, github is for software improvements and bug reports) you should always provide more than "Linux" which is meaningless, you need to provide your distribution and version and the version of Geany and GTK it uses (see near the top of help debug messages).

AnishGRao:
My apologies, I am on an ubuntu 18.04 system, and noticed that the debugger was not on my plugin list, and through some digging found it was due to having GTK3, rather than GTK2. I want to know how I would download the debugger that is in this repo, and use it on my machine. I am also using Geany 1.36.

elextr:
To repeat "the version of Geany and GTK it uses (see near the top of help debug messages)." Just copy and paste the information from there.

When you are asked for information you need to answer the question or grumpy olde sods like me will tell you off :grin: and as I said above, if its a question do it on the mailing list and also do NOT hijack existing issues for your personal questions.

codebrainz:
@AnishGRao see #942

Issue #794, opened 2018-11-29
Scope breaks nightly with cppcheck 1.85


elextr:
@LarsGit223 [This](https://github.com/geany/geany-plugins/blob/4d99fcd78c11a1615f42845c59637e7a89a2960c/scope/src/memory.c#L393) line seems to break nightlys with gcc 8, maybe needs parens around the argument to `sizeof`.

b4n:
@elextr which nightlies?  The ones on nightly.geany.org currently break because of a cppcheck issue, which I believe to be a cppcheck bug because it builds fine, and always did build fine.

Yes, adding parentheses around might be a workaround, but it still looks like a cppcheck bug to me, unless indeed gcc 8 also fails and that's not a new bug but a new restriction coming from C11 or something -- which I would find odd, but well.

I'll try to investigate what I can, but the code does look legitimate to me.  A reasonable alternative could be using `MAX()` as it's what the code actually does IIUC:
```c
pointer_size = MAX(sizeof(void *), sizeof &memory_init);
```
but I don't see any good reason to change that line apart from working around a cppcheck bug yet.

elextr:
@b4n, sorry to confuse I was just using "with gcc 8" as a way of identifying which nightlys failed, not that gcc failed. 

I didn't try to debug the failure, just identified the line, and noted the missing () on `sizeof`.  Sure `sizeof` doesn't need parens, its not a function, but for both human and software consumption its better to have them so the intention is clear and nobody has to remember the operator priority of anything more than +-*/ :)

b4n:
@elextr OK

Well, looking at the cppcheck error ("AST broken, binary operator '>' doesn't have two operands") I would expect that it's the ternary operator that doesn't get parsed correctly, and thus that it'd likely help more to add parentheses around the test, rather than around `&memory_init`.  But well, as it's a bug, who knows :)

elextr:
>  I would expect that it's the ternary operator that doesn't get parsed correctly,

Well, I was thinking the "suspicious" `sizeof` since thats whats on one side of the '>' operator, but as you say, its a bug, could be anything :grin:

I suppose `MAX` has clearer intention, just its a m-m-m-macro `<breaks out in hives>`

b4n:
Reported upstream: https://trac.cppcheck.net/ticket/8870

LarsGit223:
[...]

PR #795, opened 2018-11-29
scope: small code cleanup. See #794.

LarsGit223:
Fixed ```sizeof``` without parentheses, spaces instead of tab and a compiler warning.


b4n:
As said on #794 there was nothing wrong in those `sizeof` expressions, if anything they weren't as readable as possible; and the real issue is a bug in cppcheck.

This said, the changes here look fine, and will indeed work around that cppcheck issue, and IMO makes that `pointer_size` logic a little clearer. So, LGTM.

Issue #737, opened 2018-04-25
can't save new files under windows 10 


graph:
Steps to reproduce:

1. create a new file using the tree browser
2. give the file a name
3. Try editing the file
4. Try saving.

OS: Windows 10

doing a file->new file via geany works fine.

[original issue](https://github.com/geany/geany/issues/1842)

eadmaster:
i've found some other bugs in the  Tree Browser plugin on Windows 7:
 - cannot create new folders
- "Follow current document" does not work (but it works in the File Browser)

eht16:
I can reproduce the file creation problem and the broken "Follow current document" feature. For the former I have a fix and being testing it right now. The latter I'll check later.

@eadmaster Can you elobarate on "cannot create new folders"? What exactly happens, are there any related messages in `Help->Debug Messages`, did you try in some special path?
For me, this works on Windows 7.

l33tlinuxh4x0r:
This is still an issue on windows 11. Has anyone found a fix yet?

elextr:
The treebrowser plugin is orphaned, nobody is maintaining it.

PR #796, opened 2018-12-02
Issue737 treebrowser follow doc fix

eht16:
This fixes following the current document path.
As mentioned in #737 this was broken on Windows but actually it was broken on non-Windows as well if the start directory (Geany's default open directory or the project base directory) has no common parts with the current document's file path.

For Windows, there was an additional hack necessary: ignoring the drive letter completely in all checks.
This made it working for the moment but obviously breaks a lot of the plugin's functionality if files on another drive than the current are handled.
This should be fixed properly but since the fact that I don't use the plugin and my available time is limited, I won't do myself.


frlan:
@eht16  I assume you like to rebase this before merging?

eht16:
Sure but right now I don't want this merged at all (see my comment above as well as the WIP label).
This needs more work.

frlan:
@eht16  OK. Reopen PR once done.

PR #798, opened 2018-12-04
travis: Run distcheck as well

b4n:
Here again, I'll let it kick in, fix the issue (as there's one with distcheck right now) and then only should it be merged.  The goal is to make sure everything's working as it should.


b4n:
@elextr oh, no!  What will I do! :grin:

PR #809, opened 2019-01-04
geanypy: Use Python callback return value for keybindings return value

### LINKED ISSUE #808: [GTK2] build with gtk2 failed
In OBS https://build.opensuse.org/project/show/home:13ilya:GTK2 I'm build latest version geany and geany-plugins with GTK2.
geany 1.34 succeeded build with gtk2, but geany-plugins not built with geany built with GTK2. at the same time, it does not matter with which GTK2 or GTK3 I am trying to build geany-plugins.
https://build.opensuse.org/package/live_build_log/home:13ilya:GTK2/geany-plugins/openSUSE_Tumbleweed/x86_64

 I: Program returns random data in a function
 E: geany-plugins no-return-in-nonvoid-function geanypy-keybindings.c:36

I do not understand this error.
Perhaps this is a gtk2, geany or compilation issue.
Help to figure it out.

b4n:
This should fix #808.

The approach here is to try and have the Python callback return value be used as the C one is. I'm not totally sure how compatible that is, but IIUC if a Python function doesn't explicitly return it returns None, which should (I guess) fail the `PyObject_IsTrue()` check.

I did *not* test this *at all*: I didn't even try to build it (thanks to a crappy Internet connection tonight, I didn't wait long enough to fetch the dependencies).  This PR should be reviewed as what it is: an attempt at a fix by a guy that doesn't know neither the API nor the plugin :)

However, this should also fix a leak as the return value should likely be `DECREF`'d in any case.


geanypy/src/geanypy-keybindings.c
```diff
@@ -29,10 +29,16 @@ static gboolean call_key(gpointer *unused, guint key_id, gpointer data)
...
 	PyObject *callback = data;
 	PyObject *args;
+	PyObject *py_ret;
+	gboolean ret;
 
 	args = Py_BuildValue("(i)", key_id);
-	PyObject_CallObject(callback, args);
+	py_ret = PyObject_CallObject(callback, args);
 	Py_DECREF(args);
+	ret = (PyObject_IsTrue(py_ret) == 1);
```
b4n: The `==1` is to catch the `-1` on error, which should probably lead to returning `FALSE` rather than `TRUE`.

codebrainz:
I haven't tested yet (on Windows ATM), but this is exactly what I was thinking to do.

The only question is whether the semantics make any sense, as in whether returning nothing in the Python function should default to returning `FALSE` to Geany. The [API Reference](https://www.geany.org/manual/reference/keybindings_8h.html#afb2861d240a298186fe4d84430e5066f) makes it sound like this is not the normal case.

It may be useful in either case to improve [the doctstrings](https://github.com/geany/geany-plugins/pull/809/files#diff-8c19fc53e5ac665274c8ae6877c63a84L145) to clarify this, as a separate change if need be.

b4n:
> The only question is whether the semantics make any sense, as in whether returning nothing in the Python function should default to returning `FALSE` to Geany. The [API Reference](https://www.geany.org/manual/reference/keybindings_8h.html#afb2861d240a298186fe4d84430e5066f) makes it sound like this is not the normal case.

It's a good question, but it will always be better than the current situation that returns a more or less random value -- I'm not sure if it's undefined or implementation defined behavior, but GCC seems to return 0.
In practice, this seems to lead to no change if the Python function doesn't return, but allows it to return a value that will be passed along.

If you want returning `True` as the default case, I'm not sure what the best solution is but maybe simply assuming `None` means *"didn't return a value"* might be good enough, as it's the default return value, and not a valid value for an explicit return.

> It may be useful in either case to improve [the doctstrings](https://github.com/geany/geany-plugins/pull/809/files#diff-8c19fc53e5ac665274c8ae6877c63a84L145) to clarify this, as a separate change if need be.

Yep, definitely.

frlan:
@b4n @codebrainz  I'd volunteer to test it, but not sure about the scenario. Can you give some short list?

b4n:
@frlan The change here allows a GeanyPy plugin to tell Geany it *did* handle a keybinding, preventing any other handler using the same binding to run.

The problem is kinda tricky in real situation because this feature is more or less broken in Geany, as there's no real way to tell which handler will be called first, so it only works if all handlers on the same binding are well-behaved.

Anyway, to test, for example make a GeanyPy plugin with 2 binding handlers that do something visible (log a message or something). Then, bind those both to the same key combo, and see they both run.  Now, try and follow Geany API to return `True` in the handlers, which should prevent the other ones from running.  Before this patch, both still run.  After, only the first one should (given it returns `True`).

PR #218, opened 2015-04-23
geanylua: Update to match latest Geany's Scintilla API

b4n:
This adds new methods, and fixes build removing 2 deprecated (and then missing) ones.

Closes #217

PR #814, opened 2019-01-17
GeanyLua: Update to match latest Geany's Scintilla API 3

Skif-off:
from Scintilla 3.10.2 (see prev. [#218](https://github.com/geany/geany-plugins/pull/218), [#685](https://github.com/geany/geany-plugins/pull/685)).


eht16:
LGTM.
Since the file is generated and contains all Scintilla messages, it's OK even if probably nobody will use them. This way we keep up2date.

PR #815, opened 2019-01-23
Don't rely on a pkg-config output formatting detail

b4n:
Do not require pkg-config to emit spaces around elements in its `--print-requires` output.

This should fix Geany GTK version check on OpenBSD 6.4.

---
@andy5995 could you check this and see if it helps?  Once configured, you should get something like that (obviously with GTK2 version of it if you've got a GTK2 build):
```shell
$ grep ^GP_GTK_ config.log 
GP_GTK_PACKAGE='gtk+-3.0'
GP_GTK_VERSION='3.0'
GP_GTK_VERSION_MAJOR='3'
```


b4n:
@andy5995 thanks for testing!

andy5995:
You're welcome!

PR #816, opened 2019-01-25
debugger/bptree.c:fix compiler warning

andy5995:
I noticed this warning on buster and slack current, both using gcc8.

This fixes the warning (caused by mixed tabs/spaces indentation):

```
bptree.c: In function ‘on_hitscount_changed’:
bptree.c:373:5: warning: this ‘if’ clause does not guard...
[-Wmisleading-indentation]
     if (oldcount != count)
     ^~
bptree.c:376:2: note: ...this statement, but the latter is misleadingly
indented as if it were guarded by the ‘if’
  gtk_tree_path_free(tree_path);
  ^~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/debugger/src/bptree.c
+++ b/debugger/src/bptree.c
@@ -370,9 +370,9 @@ static void on_hitscount_changed(GtkCellRendererText *renderer, gchar *path, gch
 		LINE, &line,
         -1);
         
-    if (oldcount != count)
-    	breaks_set_hits_count(file, line, count);
-	
+	if (oldcount != count)
+		breaks_set_hits_count(file, line, count);
+
 	gtk_tree_path_free(tree_path);
 	g_free(file);
 }                                                        


elextr:
Sigh, TABS SHOULD DIE!!! `</rant>`

Anyway the code is fine its just that its got mixed spaces and tabs in the indentation, open the file in Geany and `View->show whitespace`

andy5995:
Yup, showing whitespace was how I found the cause of the warning. Looked like tabs were used mostly throughout the source, so I replaced some space indentation with tabs.

I assume you'd rather not have the warning showing for gcc8 compilers?

I wonder if gcc8 shouldn't be added to the travis build matrix, such as is done here:
https://github.com/theimpossibleastronaut/rmw/blob/master/.travis.yml cc @b4n 

And if so, may also want that for the main geany repo.

elextr:
@andy5995 note that HACKING defines the style for Geany code, (indent with tabs and align with spaces) but its up to plugins if they follow it.

And even though I HATE that style its more important that Geany be homogenous than exactly what the style is.

andy5995:
@elextr I read a little about the differences between alignment and indentation, and IIUC, this PR is ok, and you're giving me a friendly FYI for future reference. Correct?

elextr:
@andy5995 exactly :)

PR #823, opened 2019-02-17
scope: Improved evaluation on-hover

LarsGit223:
Scope automatically evaluates an expression if the mouse pointer is hovered over it. But before this change it only evaluated words. E.g. if the mouse was hovered over ```structa->item1``` then scope would have evaluated ```structa``` or ```itemb``` depending on the exact position of the mouse pointer. With this change it will instead evaluate the value of ```structa->item1```.

### DIFF
--- a/scope/src/utils.c
+++ b/scope/src/utils.c
@@ -36,6 +36,10 @@
 #ifdef G_OS_UNIX
 #include <fcntl.h>
 
+/* The maximum length of an expression for evaluating the value.
+   (Including the string terminator '\0') */
+#define SCOPE_MAX_EVALUATE_EXPR_LENGTH 256
+
 void show_errno(const char *prefix)
 {
 	show_error(_("%s: %s."), prefix, g_strerror(errno));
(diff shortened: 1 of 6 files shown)


b4n:
Just from the top of my head, wouldn't it be better to evaluate `structa` when over `structa`, and `structa->item1` when over `item1`?  I think it's reasonably useful to be able to get the evaluation of a part of the expression, here the parent.

Also if you wanted to be thorough, the C and C++ syntax allows for any kind of spaces between `structa`, `->` and `item1` :grin:

LarsGit223:
> Just from the top of my head, wouldn't it be better to evaluate `structa` when over `structa`, and `structa->item1` when over `item1`?

Yes, good idea. I'll try to implement that.

> Also if you wanted to be thorough, the C and C++ syntax allows for any kind of spaces between `structa`, `->` and `item1` :grin:

True. But and it might even include linebreaks I think. And for C++ we might need to allow ```::``` for static class members.

LarsGit223:
@b4n, @elextr: I just pushed another improvement for the on-hover-evaluation of expressions. It works like suggested by b4n and also allows spaces but it does only work on the current line.

I wanted to write unit tests for the function that parses the current line and extracts the expression to evaluate. But my unit tests don't build. I have setup an extra folder ```tests```. In it are the test files. If I try to build the tests via ```make check``` then compilation for the scope unit tests fails because ```check.h``` cannot be found. Another thing is that if I remove the comments from ```scope/tests/Makefile.am``` then none of the tests source code files is compiled. It now looks like this:
```
#if UNITTESTS

include $(top_srcdir)/build/vars.build.mk
plugin = scope

check_PROGRAMS = scope-check

scope_check_SOURCES  = unittests.c utils_test.c
scope_check_CPPFLAGS = -DTEST
scope_check_CFLAGS   = $(GEANY_CFLAGS) -DUNITTESTS
scope_check_LDADD    = @GEANY_LIBS@ $(INTLLIBS) @CHECK_LIBS@

TESTS = $(check_PROGRAMS)

#endif
```
So it seems that I have overseen one magic autotools switch which is still missing. Any help would be highly appreciated.

LarsGit223:
> So it seems that I have overseen one magic autotools switch which is still missing. Any help would be highly appreciated.

**Junk!** I simply forgot to install check on my machine (embarrassing). For anyone reading this: make sure you have check installed if you want to write unit tests. Otherwise you might get an error message like this:

gp-check-h-missing

If you execute the command ```./autogen.sh``` it will show if check is present/if unit tests can be created:

gp-unittests-no

In case check is installed it will show ```Unit tests: yes```. If not, building tests via ```make check``` will fail.

LarsGit223:
This looks quite good to me now. Unit tests worked and were run successfully on my machine. Unfortunately scope isn't build in Travis CI as VTE support is missing.

The array in ```scope/src/tests/utils_test.c``` shows which situations are handled.

LarsGit223:
@frlan: I would also like to have this one in milestone 1.36.0 - if no one objects.

LarsGit223:
@b4n, @frlan: If no one objects I will merge this soon also.

LarsGit223:
No complaints, so I merge this now.

scope/src/utils.c
```diff
@@ -36,6 +36,10 @@
 #ifdef G_OS_UNIX
 #include <fcntl.h>
 
+/* The maximum length of an expression for evaluating the value.
+   (Including the string terminator '\0') */
+#define SCOPE_MAX_EVALUATE_EXPR_LENGTH 256
```
eht16: Is it intended that the macro definition is within the `#ifdef G_OS_UNIX` block?
This breaks the Windows builds: https://nightly.geany.org/win32/build_win32_plugins_stderr.log as it used unconditionally later on.

I assume the macro definition should be outside any OS guards?

LarsGit223: @eht16: yes - sorry for the hassle. Fixed in PR #885.

LarsGit223: @eht16: I think I have asked this before but would it be possible to add a Windows build to Travis CI?

eht16: Thanks for #885.
Regarding Windows Travis builds: my comment https://github.com/geany/geany-plugins/pull/787#issuecomment-438838305 is still true. It is possible just needs some work and probably lots of trial-and-error.

LarsGit223: > If someone wants to start, I can provide the necessary GTK stack (i.e. the files we use for the nightly builds).

I added my own geany-plugins repo to Travis CI so I could start there without breaking Travis builds at geany for a probably longer time. I will create a branch for it. Yes - as offered in your previous comment - please provide the files for the nightly builds.

eht16: Here we go:
https://download.geany.org/contrib/gtk/gtk%2B-bundle_2.24-plugins-dependencies.tar.gz
This is a tarball with GTK2 Windows binaries as well as available libraries needed by several plugins. Those libraries are currently used for the nightly builds.
As a start, I hacked a little build script: https://www.geany.org/p/ZnQOw/
This is not nice but should show how it works. I guess https://github.com/geany/geany/blob/master/scripts/cross-build-mingw.sh could be a good base to start with and just adjust it for G-P.

Issue #824, opened 2019-02-17
scope: crash on refreshing "Locals"


LarsGit223:
The following sequence of actions crashes Geany:
1. start debugging a program with the scope plugin
1. wait for a breakpoint
1. in the message window select _"Debug"_ and then select tab _"Locals"_
1. right click inside the tab to show the context menu, select _"Refresh"_
1. Geany crashes (not always but often)

Remark: the crash does not occur if the user first left clicks into the _"Locals"_ tab content area before selecting _"Refresh"_.

PR #827, opened 2019-02-18
scope: prevent usage of invalid iters (GtkTreeeIter)

LarsGit223:
This fixes the crash described in #824 (and two more potential crashes).

### DIFF
--- a/scope/src/break.c
+++ b/scope/src/break.c
@@ -1332,8 +1332,10 @@ static void on_break_apply(const MenuItem *menu_item)
 	if (menu_item || thread_id)
 	{
 		GtkTreeIter iter;
-		gtk_tree_selection_get_selected(selection, NULL, &iter);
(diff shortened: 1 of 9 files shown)


b4n:
This is not really the right fix, `g_return_*if_fail()` is supposed to catch programming error, to check the contract between the caller and the callee.  Admittedly here the values are accessed before the contract is checked, which is probably worth fixing, but the bug itself is that these functions get called with values they are not supposed to.

LarsGit223:
I will try to find the trouble source. Anyway I would like to keep the existing changes for better stability in case of an error.

LarsGit223:
Done. The wrong call came from ```stack_entry()``` because of not checking the result of ```gtk_tree_selection_get_selected()```.

LarsGit223:
@b4n: do you request more changes? I would like to merge this soon.

PR #842, opened 2019-04-14
Workbench: Remove not needed explicit OK and CANCEL button

frlan:


workbench/src/dialogs.c
```diff
@@ -289,8 +289,8 @@ gboolean dialogs_directory_settings(WB_PROJECT_DIR *directory)
 	dialog = gtk_dialog_new_with_buttons(_("Directory settings"),
 										 GTK_WINDOW(wb_globals.geany_plugin->geany_data->main_widgets->window),
 										 flags,
-										 _("_Cancel"), GTK_RESPONSE_CANCEL,
-										 _("_OK"), GTK_RESPONSE_ACCEPT,
+										 GTK_STOCK_CANCEL, GTK_RESPONSE_CANCEL,
+										 GTK_STOCK_OK, GTK_RESPONSE_ACCEPT,
```
LarsGit223: As per the documentation under https://developer.gnome.org/gtk3/stable/gtk3-Stock-Items.html#GTK-STOCK-CANCEL:CAPS stock items are deprecated. I think it should better stay as-is.

frlan: Puh. Whatever is behind this I don't like it. :/

LarsGit223: Found this blog which says it was deprecated to save memory, see https://blogs.gnome.org/tko/2005/10/26/gtk-builtin-stock-icons-removed/. On this quick search I did not find anything from the gtk development team. Anyway, we can't change it.

codebrainz: IIRC this was part of the [Infamous Deprecations of 2013](https://mail.gnome.org/archives/gtk-devel-list/2013-July/msg00000.html). I believe the issue of each project having to provide their own translations for each of these stock items was brought up and dismissed by GNOME designers.

PR #848, opened 2019-05-04
vimode: Separate command and search history

techee:
Vi seems to use a separate history for commands (starting with ':')
and searches (starting with '/' or '?'). Let's do the same.

### DIFF
--- a/vimode/src/excmd-prompt.c
+++ b/vimode/src/excmd-prompt.c
@@ -91,10 +91,11 @@ static gboolean on_prompt_key_press_event(GtkWidget *widget, GdkEventKey *event,
 			case GDK_KEY_KP_Enter:
 			case GDK_KEY_ISO_Enter:
 			{
-				guint index;
+				guint i;
 
-				if (g_ptr_array_find_with_equal_func(history, text + 1, g_str_equal, &index))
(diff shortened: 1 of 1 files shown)


vimode/src/excmd-prompt.c
```diff
@@ -60,16 +74,13 @@ static gboolean on_prompt_key_press_event(GtkWidget *widget, GdkEventKey *event,
...
-			GList *link = g_list_find_custom(history, text, (GCompareFunc)g_strcmp0);
 
-			if (link)
-			{
-				g_free(link->data);
-				history = g_list_remove_link(history, link);
-			}
-			if (g_strcmp0(text, ":") != 0)
-				history = g_list_prepend(history, g_strdup(text));
+			if (g_ptr_array_find_with_equal_func(history, text + 1, g_str_equal, &index))
```
codebrainz: Requires [GLib 2.54](https://developer.gnome.org/glib/stable/glib-Pointer-Arrays.html#g-ptr-array-find-with-equal-func) or greater I guess.

techee: Ah, OK, I didn't notice the checks failed. I'll add some additional commits to this pull request so the command line behaves more like vim.

PR #856, opened 2019-05-12
lipsum: fixed deprecated gtk call for gtk3

LarsGit223:
This fixes a deprecation warning in the lipsum plugin for gtk3. No functional changes.

### DIFF
--- a/lipsum/src/lipsum.c
+++ b/lipsum/src/lipsum.c
@@ -219,7 +219,11 @@ plugin_init(G_GNUC_UNUSED GeanyData *data)
 	g_free(config_file);
 
 	/* Building menu entry */
+#if GTK_CHECK_VERSION(3, 10, 0)
+	menu_lipsum = gtk_menu_item_new_with_mnemonic(_("_Lipsum..."));
+#else
 	menu_lipsum = gtk_image_menu_item_new_with_mnemonic(_("_Lipsum..."));
+#endif
 	gtk_widget_set_tooltip_text(menu_lipsum, _("Include Pseudotext to your code"));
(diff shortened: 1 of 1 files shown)


utils/src/gp_gtkcompat.h
```diff
@@ -51,6 +51,14 @@ G_BEGIN_DECLS
 #define GTK_STOCK_OPEN   _("_Open")
 #define GTK_STOCK_CANCEL _("_Cancel")
 #endif
+
+/* Replace calls to gtk_image_menu_item_new_with_mnemonic() with call to
+   gtk_menu_item_new_with_mnemonic(). Starting from version 3.10.*/
+#if GTK_CHECK_VERSION(3, 10, 0)
+#define gtk_image_menu_item_new_with_mnemonic(icon_info) \
+        gtk_menu_item_new_with_mnemonic(icon_info)
```
codebrainz: This doesn't seem like a good wrapper for general use in the util library since it returns a different type of widget.

While much more work, it might be better to add some kind of real image menu item replacement class that packs an optional icon and an accel label into a box, similar to the example on the [GtkImageMenuItem docs](https://developer.gnome.org/gtk3/stable/GtkImageMenuItem.html).

LarsGit223: Thanks. But I guess I will then postpone that to a later time when it's really required. Here with lipsum there is no icon used at all. I will do the replacement in the lipsum file directly and remove the macro. If I find a use case later which really uses an icon I will go for your advice.

LarsGit223: Done.

PR #858, opened 2019-05-13
extrasel: fixed deprecated gtk call

LarsGit223:
I am a bit uncertain about the version check but if I understood correctly, GDK is part of GTK.


codebrainz:
The check looks OK to me.

Alternatively, you could fix the [similar code in Geany](https://github.com/geany/geany/blob/868da11715ed86d08c1c2c4deaf1b2aa6c209df4/src/utils.c#L917) and expose the function to the plugin API for use in this plugin (and any others which beep).

LarsGit223:
@codebrainz: Thanks, that makes perfect sense - will do that.

Issue #841, opened 2019-04-14
workbench: Changing options are not saved


frlan:
When using the Tools->Workbench->Settings changed options are not saved to configuration file of workbench project

frlan:
@LarsGit223 I think this is something for ;)

LarsGit223:
@frlan: did you choose _"Tools->Workbench->Save"_? The settings are not saved automatically.

frlan:
@LarsGit223  No I did not. Expectation for settings was, that it is saved automatically.

LarsGit223:
Well, then everything works as-designed and it is not a bug. But you could change the label to "enhancement" if you like.

frlan:
Done. While unexpected behaviour is a bug for me ;)

LarsGit223:
> Done. While unexpected behaviour is a bug for me ;)

It is the expected behaviour (it works as implemented/expected by the author). But your expectations doesn't match the actual implementation ;-)

LarsGit223:
@frlan: I now also removed the option _"Save project"_ and projects are also saved automatically if a workbench related project setting is changed. Also I fixed a little bug regarding this issue, see PR #877.

PR #877, opened 2019-06-23
workbench: auto-save project on change

LarsGit223:
This removes the item _"Save project"_ from the popup menu and automatically saves a project if something is changed (e.g. directory added/removed).

This also fixes some situations in which saving of the workbench file was missing (e.g. a change in the workbench bookmarks). See #841.


LarsGit223:
@frlan: thanks for the quick merge.

PR #879, opened 2019-06-27
workbench: added option to let git decide which files to display

LarsGit223:
In the directory settings it is now possible to choose between a workbench or a git file filter. The later is new and lets git decide which files to display depending on the contents of the ```.gitignore``` file.

This also adds a new function to the utils-lib called ```gp_filelist_scan_directory_callback()```.

See this example screenshot:
DirectorySettingsWithGitOption

### DIFF
--- a/build/workbench.m4
+++ b/build/workbench.m4
@@ -3,6 +3,9 @@ AC_DEFUN([GP_CHECK_WORKBENCH],
     GP_ARG_DISABLE([Workbench], [auto])
     GP_CHECK_UTILSLIB([Workbench])
 
+    GP_CHECK_PLUGIN_DEPS([Workbench], [WORKBENCH],
+                         [libgit2 >= 0.21])
+
     GP_COMMIT_PLUGIN_STATUS([Workbench])
     AC_CONFIG_FILES([
         workbench/Makefile
--- a/utils/src/filelist.c
+++ b/utils/src/filelist.c
@@ -39,6 +39,16 @@ typedef struct
 ScanDirParams;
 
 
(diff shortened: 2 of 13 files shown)


LarsGit223:
I need some help or advice regarding Travis CI. The build of the workbench plugin failed with this error message:
```
../../../workbench/src/plugin_main.c:28:18: fatal error: git2.h: No such file or directory
 #include <git2.h>
```
I found this comment in the Travis CI config:
```
    # Ubuntu 12.04 doesn't have libgit2, 14.04 has a too old one, so we build
    # our own for now.
    #- libgit2-dev
```
On my machine I have libgit2-dev installed (Ubuntu 18.04.2). How can I fix this issue?

elextr:
There seems to be a logical error in the `.travis.yml`, the presence of a .pc file for libgit2 is tested [here](https://github.com/geany/geany-plugins/blob/4ddf2e7707d3e0305a3e463d4574e0f715102a0e/.travis.yml#L108) but that doesn't check the version, it just accepts any version.

But that doesn't seem to be the issue here since it doesn't seem to execute the echo in the `then` clause or the build in the `else` clause, weird.

LarsGit223:
@elextr: the issue was caused by an incomplete makefile in the workbench plugin. It missed  ```$(WORKBENCH_CFLAGS)``` in the ```workbench_la_CFLAGS```. The result was that building it locally on my machine worked, but the Travis CI build failed.

LarsGit223:
Waited 3 weeks an Travis builds are fixed. So I merge this now.

PR #882, opened 2019-07-12
Cleanup: fixed usage of '#include "geanyplugin.h"'

LarsGit223:
Nothing more to say.

### DIFF
--- a/addons/src/addons.c
+++ b/addons/src/addons.c
@@ -26,7 +26,7 @@
 # include "config.h"
 #endif
 
-#include "geanyplugin.h"
+#include <geanyplugin.h>
 
 #include "addons.h"
 #include "ao_blanklines.h"
--- a/addons/src/ao_blanklines.c
+++ b/addons/src/ao_blanklines.c
@@ -19,7 +19,7 @@
  *			MA 02110-1301, USA.
  */
 
-#include "geanyplugin.h"
+#include <geanyplugin.h>
 
 #include "addons.h"
 #include "ao_blanklines.h"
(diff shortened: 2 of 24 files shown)


elextr:
IIUC <> is for including system files, and despite its delusions of grandeur I don't think Geany is guaranteed to be in the list of system search directories.

codebrainz:
https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html

> I don't think Geany is guaranteed to be in the list of system search directories.

It is, by the build system.

elextr:
From the link:

```
"#include "file"    This variant is used for header files of your own program.
```

Sure GCC allows you to abuse the convention, but that doesn't mean you should.

LarsGit223:
> #include \<file>
> 
>This variant is used for system header files. It searches for a file named file in a standard list of system directories. You can prepend directories to this list with the -I option (see Invocation).
> 

Isn't this the normal way to do it? The ```geanyplugin.h``` file is not part of geany-plugins but part of geany.

codebrainz:
> Isn't this the normal way to do it?

It is. `geanyplugin.h` is a system include provided by `libgeany`, not local to plugins.

kugel-:
Doesn't really matter if you use `""` or `<>`, the only difference is that `""` first looks in the location of the source file that is being compiled (I believe it's actually the working directory of the compiler), whereas `<>` looks only in paths provided on the command line and default include dirs (/usr/include) (`""` also looks in these paths).

So for geanyplugin.h you need the include path on the command line, so both `<>` and `""` are equally correct. But, by convention, you should prefer `<>` for headers that don't belong to your package, which is the case here.

PR #895, opened 2019-08-10
Fixed document list menu position bug

PatrickGMM:
Fixed document list menu position bug when the tool button has been moved to the toolbar overflow menu
  - tested on builds of Geany 1.35 for Gtk versions 3.22.30 and 2.24.32

### DIFF
--- a/addons/src/ao_doclist.c
+++ b/addons/src/ao_doclist.c
@@ -143,16 +143,13 @@ static void ao_popup_position_menu(GtkMenu *menu, gint *x, gint *y, gboolean *pu
 		window = gtk_widget_get_window (gtk_widget_get_ancestor(widget, GTK_TYPE_TOOLBAR));
 #if GTK_CHECK_VERSION(3, 0, 0)
 		gdk_window_get_geometry(window, &wx, &wy, &width, NULL);
+		gtk_widget_get_preferred_size(widget, &widget_req, NULL);
 #else
 		gdk_window_get_geometry(window, &wx, &wy, &width, NULL, NULL);
+		gtk_widget_size_request(widget, &widget_req);
 #endif
 		gdk_window_get_root_coords(window, wx, wy, &wx_root, &wy_root);
 
-#if GTK_CHECK_VERSION(3, 0, 0)
-		gtk_widget_get_preferred_size(widget, &widget_req, NULL);
-#else
-		gtk_widget_size_request(widget, &widget_req);
-#endif
 		/* Approximate the horizontal location of the overflow menu button */
 		/* TODO: See if there's a way to find the exact location */
 		wx = wx_root + width - (int) (widget_req.width * 1.5);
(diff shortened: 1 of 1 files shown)


addons/src/ao_doclist.c
```diff
@@ -120,23 +124,60 @@ static void ao_doc_list_finalize(GObject *object)
...
+
+		/* Get the root coordinates of the toolbar's window */
+		int wx_root, wy_root, width;
+		window = gtk_widget_get_window (gtk_widget_get_ancestor(widget, GTK_TYPE_TOOLBAR));
+#if GTK_CHECK_VERSION(3, 0, 0)
+		gdk_window_get_geometry(window, &wx, &wy, &width, NULL);
+#else
+		gdk_window_get_geometry(window, &wx, &wy, &width, NULL, NULL);
+#endif
+		gdk_window_get_root_coords(window, wx, wy, &wx_root, &wy_root);
```
eht16: Not completely sure but can't we move the `gdk_window_get_root_coords` call below the others and so merge the two GTK version check blocks?

addons/src/ao_doclist.c
```diff
@@ -120,23 +124,60 @@ static void ao_doc_list_finalize(GObject *object)
...
+		 * the doclist menu item
+		 */
+		priv->in_overflow_menu = FALSE;
+
+	} else {
+		/* Retrieve size and position of both widget and menu */
+		window = gtk_widget_get_window(widget);
+		if (! gtk_widget_get_has_window(widget))
+		{
+
```
eht16: Is this empty line on purpose?

addons/src/ao_doclist.c
```diff
@@ -120,23 +124,60 @@ static void ao_doc_list_finalize(GObject *object)
...
+#endif
+		gdk_window_get_root_coords(window, wx, wy, &wx_root, &wy_root);
+
+#if GTK_CHECK_VERSION(3, 0, 0)
+		gtk_widget_get_preferred_size(widget, &widget_req, NULL);
+#else
+		gtk_widget_size_request(widget, &widget_req);
+#endif
+		/* Approximate the horizontal location of the overflow menu button */
+		/* TODO: See if there's a way to find the exact location */
```
eht16: Do you intend to resolve this TODO?
Otherwise we should probably rather describe that there is no known way to get the exact position.

addons/src/ao_doclist.c
PatrickGMM: Yes, I will change this.

addons/src/ao_doclist.c
PatrickGMM: No, I will delete this.

addons/src/ao_doclist.c
PatrickGMM: I was hoping somebody with more Gtk experience would know how to find this button in the toolbar. If I can assume the drop-down for the overflow will be at the end of the toolbar item list, then I can easily resolve this. But I'm not sure if that is always the case (for instance in right-to-left locales).

PatrickGMM:
> Could you describe in a bit more detail the exact problem?
> I see the overflow menu appears "somewhere" on the screen but it still works basically. It's just offsetted.
> 
> Your changes seems to fix this quite fine, just want to know if there are more problems than the mispositioning.

The main problem was the positioning. And the description in the overflow menu was simply "Index" which I changed to "Document List" which is more descriptive of the plugin feature.

addons/src/ao_doclist.c
PatrickGMM: Never mind on being able to use the last item in the toolbar as the overflow drop-down. I made an incorrect assumption in my previous statement. There doesn't seem to be a way to get the toolbar widget as it's drawn.

eht16:
Ok, thanks for clarifying.
When you changed the remaining points, we can merge it.

eht16:
Thanks.

PR #896, opened 2019-08-11
autoclose: remove pairing quotes

### LINKED ISSUE #687: [New feature] autoclose: remove pairing quotes
Hi.

I often use the feature to remove pairing braces using shift + backspace given by autoclose. This works for brackets and parenthesis as well.
I miss the same feature regarding quotes, both for simple or doble quotes.

LarsGit223:
Using ```Shift```+```Backspace``` at the moment only removes round, square, waved and angle brackets but not quotes. With this PR the feature also works for single, double and back-quotes. Closes #687.

### DIFF
--- a/autoclose/src/autoclose.c
+++ b/autoclose/src/autoclose.c
@@ -367,6 +367,43 @@ improve_indent(
 	return AC_STOP_ACTION;
 }
 
+
+/** Find previous string/last occurence from a position
+ *
+ * The function finds the last occurence of @a text before the given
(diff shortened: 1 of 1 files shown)


LarsGit223:
@scriptum: ping.

@frlan: maybe we should merge this if there is no reaction and so others can test and use it in master.

frlan:
Yepp. Good point. Still some time to go for next release and no feedback by maintainer. Scrolling code did not show big issues to me :)

Issue #819, opened 2019-02-08
Line Operations: Add support for locale-aware sorting


danpla:
Line operations currently use `strcmp()`, so sorting non-ASCII text gives nonsense (because it sorts by bites of UTF-8 sequences). It would be nice to have normal sorting that use `g_utf8_collate()` (or `g_utf8_collate_key()` for better performance) either by default (replacing `strcmp()`) or as separate operations.

smostertdev:
I'll see if I can get `lineoperations` working with non-ASCII characters. Thanks for pointing this out.

elextr:
Probably best to make it separate operations, `strcmp()` isn't nonsense, it sorts by Unicode code point value, which is useful in some cases, and adding sorting by some locale stuff is useful in others. 

Also there is the problem of what locale it uses.  Remember files are shared around the world by github and friends, and having things sort by whatever locale the particular person editing it lives in isn't usually very sensible.

And requiring a user to restart Geany in a different locale to get the sort that a particular file requires is not a very user friendly approach either.

LarsGit223:
@danpla: this enhancement has been implemented by @smostertdev now and will be included in the next release of geany-plugins.

danpla:
Thanks!

PR #897, opened 2019-08-12
lineoperations: add support for locale-aware sorting 

smostertdev:
This PR is in response to #819.

Changes:
- add preferences option: 'Use collation based string compare'
image

- `getcmpfns()` returns either `g_utf8_collate` or `strcmp` function pointer to line operations function based on user preference

Minor:

- add help function to redirect to online documentation
- split files into `lo_prefs` and `lo_fns`
- sorry about all the annoying indentation / spacing changes / file renaming.

I'm open to hear any suggestions/improvements.

### DIFF
--- a/lineoperations/README
+++ b/lineoperations/README
@@ -43,7 +43,7 @@ on bottom of the "Plugin Manager" frame.
 compare strings using linguistically correct rules for the current locale
 (g_utf8_collate).
 
-When it is disabled it will compare strings based off numerical value of the
+When it is disabled it will compare strings based on numerical value of the
 characters (strcmp).
 
 This option is disabled by default.
--- a/lineoperations/README
+++ b/lineoperations/README
@@ -7,7 +7,7 @@ Line Operations
 About
 =====
 
-Line Operations is an assortment of simple line functions that can be 
+Line Operations is an assortment of simple line functions that can be
 applied to an open file, or selection.
 
 Features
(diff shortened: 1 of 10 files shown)


frlan:
Please update POTFILES.in

lineoperations/src/lo_prefs.h
```diff
@@ -0,0 +1,59 @@
...
+#include <geanyplugin.h>
+#include "lo_prefs.h"
+
+
+typedef struct
+{
+	/* general settings */
+	gchar *config_file;
+	gboolean use_collation_compare;
+} AddonsInfo;
```
LarsGit223: Is this intentionally named ```AddonsInfo``` or a copy&paste issue?

lineoperations/src/lo_prefs.h
```diff
@@ -0,0 +1,59 @@
...
+
+
+extern AddonsInfo *lo_info;
```
LarsGit223: Same question here.

lineoperations/README
```diff
@@ -26,23 +26,72 @@ Features
...
+
+To enable preferences ("Tools" > "Plugin Manager"), check the checkbox to
+enable Line Operations, and with the item highlighted click "Preferences"
+on bottom of the "Plugin Manager" frame.
+
+**Use collation based string compare** - When this option is enabled it will
+compare strings using linguistically correct rules for the current locale
+(g_utf8_collate).
+
+When it is disabled it will compare strings based off numerical value of the
```
LarsGit223: Should it be _"based off"_ or _"based on"_?

lineoperations/src/lo_fns.c
```diff
@@ -20,22 +20,40 @@
...
+lo_strcmpfns
+getcmpfns(void)
+{
+	if(lo_info->use_collation_compare)
+	{
+		return g_utf8_collate;
+	}
+	else
+	{
+		return strcmp;
```
LarsGit223: Maybe you like to prefer ```g_strcmp0``` from the glib here? The same as ```strcmp``` just a bit safer but from glib.

lineoperations/src/lo_prefs.h
smostertdev: You are correct, updated

lineoperations/src/lo_prefs.h
smostertdev: You are correct, updated

lineoperations/README
smostertdev: I think they're both [valid](https://english.stackexchange.com/questions/15081/based-on-instead-of-based-off-of). I changed it to "based on", I think it sounds a little better.

lineoperations/src/lo_fns.c
smostertdev: For some reason I changed it a few years ago https://github.com/geany/geany-plugins/commit/2e2e043318fd2630cd05f1ee95c49fcdb019b6f7, I just don't remember why.

I agree with you, I think it is better with glib. I tested it and do not see any problems, so I'm not sure what I was thinking a few years ago.

LarsGit223:
LGBI. The commits should be squashed before merging.

PR #903, opened 2019-09-07
updatechecker: remove duplicate config_file string allocation

smostertdev:
I think that this might be a small memory leak:

The string `config_file` is allocated in `init_configuration()` when the plugin is loaded, and then again in the callback function `on_configure_response()`.

From what I can tell it's not being freed every time the callback is called.


frlan:
Good catch

PR #904, opened 2019-09-11
projectorganizer : add shortcut to focus sidebar project tab

AlicVB:
So one can easily switch between tabs with keyboard.

This is my first contribution here, I hope I've done everything correctly...
Thanks

### DIFF
--- a/projectorganizer/src/prjorg-menu.c
+++ b/projectorganizer/src/prjorg-menu.c
@@ -42,6 +42,7 @@ enum
 	KB_FIND_IN_PROJECT,
 	KB_FIND_FILE,
 	KB_FIND_TAG,
+	KB_FOCUS_TAB,
 	KB_COUNT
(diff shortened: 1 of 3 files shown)


techee:
@AlicVB Isn't Geany's "Focus->Switch to Sidebar" keybinding enough for this purpose?

AlicVB:
well, not really as there's many tabs in sidebar.
I mainly use this new shortcut to switch between sidebar tabs with keyboard... (essentially between symbols and project tabs in my case)

techee:
OK, should be addressed by #920

techee:
Meh, just now I noticed this wasn't a bug report but actually a pull request - could have saved me some work. In my patch I just avoided assigning tab number which is probably better because the number can get wrong when some plugins add/remove new tabs to/from the sidebar.

PR #907, opened 2019-09-22
Delete danish translation

scootergrisen:
Since it contain almost no translated strings.


b4n:
It contains all the useful translations for the *shiftcolumn* plugin, so even though that's only 3 strings, it seems partly useful (not like if only 3 of all strings in Geany itself were translated).
Is this file problematic for any reason I'm not aware of?  Yes, it'd need some work from Danish translators to be more widely useful, but I don't see it as harmful and it provides a better experience for Danish users of the *shiftcolumn* plugin, which seems worth it to me, so long as it's correct (but it looks reasonable from where I stand -- e.g. not speaking a word of Danish).

scootergrisen:
I would say the only 3 translated strings are incorrect with large letter K, V, H in Kolonne, Venstre, Højre which should be small.

    msgstr "Flyt Kolonne"
    msgstr "Flyt til Venstre"
    msgstr "Flyt til Højre"

And a lot of fuzzy with "Flyt det valgte til højre/venstre" which is not correct so the translation would properly spend more time translation those then starting from a empty pot file because the translator have to delete the text first.

I just think not the da.po better indicate the status. That there is no danish translation.
Rather then there is a danish translation with 3 incorrect translated strings.

frlan:
@scootergrisen  Instead of deleting could you maybe start a better one? ;) But of course low level of translated strings with a high failure count does not help

PR #914, opened 2019-10-01
pairtaghighlighter: do not stop scanning on linebreaks

### LINKED ISSUE #775: pairtaghighlighter doesn't find pair tag, if tag is split on a few lines
For example:
```
<button class="date"
        onclick="this.innerHTML=Date()">The time is?</button>

LarsGit223:
Before this change the plugin stopped scanning for a tag on the start or the end of a line. Therefore tags which were split over multiple lines were not recognized. With this change the plugin scans forward or backward up to a certain limit and does not stop on a linebreak. The default limit is 1024 bytes but it can be changed by configuration if required. Fixes #775.


LarsGit223:
@vmkononenko: please have a look.

PR #919, opened 2019-10-05
vimode: Don't use word navigation from Scintilla

### LINKED ISSUE #909: dw/cw incorrect behaviour
Hi,

I use vim quite a lot and noticed a few incorrect behaviours, dw and cw being one.

I'll denote the cursor as a double quote ("), before a letter means it's on that letter.
In geany vimode the following line int the function block:

`void foo(BarClass *pointer) {
    std::"cout << pointer->bar_func() << std::endl;
}`

pressing dw or cw would yield:

`void foo(BarClass *pointer) {
    std::"pointer->bar_func() << std::endl;
}`

in vim either would yield: (it deletes up to some special characters, not the next word)

`void foo(BarClass *pointer) {
    std::"<< pointer->bar_func() << std::endl;
}`

Another example, start again here, with cursor before pointer:

`void foo(BarClass *pointer) {
    std::cout << "pointer->bar_func() << std::endl;
}`

geany yields:

`void foo(BarClass *pointer) {
    std::cout << "bar_func() << std::endl;
}`

vim yields:

`void foo(BarClass *pointer) {
    std::cout << "->bar_func() << std::endl;
}`

Geany vimode also treats dw and cw the same in all cases (apart from dw staying in normal mode and cw entering insert mode, which is correct). In the below example, cursor on B of BarClass:

`void foo("BarClass *pointer) {
    std::cout << pointer->bar_func() << std::endl;
}`

for both, geany would yield:

`void foo("pointer) {
    std::cout << pointer->bar_func() << std::endl;
}`

dw in vim would yield: (asterisk remains)

`void foo("*pointer) {
    std::cout << pointer->bar_func() << std::endl;
}`

cw would yield: (extra space, so when typing new word, space is already there)

`void foo(" *pointer) {
    std::cout << pointer->bar_func() << std::endl;
}`

It's worth noting that the special character seem to be specific, so for example underscore is treated as part of the word in vim:

`void foo(BarClass *pointer) {
    std::cout << pointer->"bar_func() << std::endl;
}`

dw or cw in vim becomes: (deletes underscore too)

`void foo(BarClass *pointer) {
    std::cout << pointer->"() << std::endl;
}`

These seem minor but it is much more often the case that the Alphanum characters and underscore are the ones that the user is considering the 'word' and wants to delete, having to re-type the operators such as '<<' or '-' or '>' etc etc is not the intention.

If it is likely that these changes will get integrated into vimode I would like to offer my services since I work with c and c++ and vim, I think I could get it working just right. Am I ok to dig in?

Thanks,

Lukas

techee:
Scintilla does different things when using SCI_WORDRIGHT, SCI_WORDLEFT,
SCI_WORDRIGHTEND and SCI_WORDLEFTEND than what we want for commands like
"w", "b", "W", "B", "e", "E", etc. For instance, for

ident1 = id|ent2+ident3;

where | represents the cursor, SCI_WORDRIGHT does

ident1 = ident2+|ident3;

while Vim's "w" does

ident1 = ident2|+ident3;

For this reason we have to reimplement all the word navigation by ourselves
to better emulate the behavior of Vim. This patch tries to do that.

Should fix #909


alukreis:
This works well!, I did find one very specific issue though, the cursor will no go back (b) to the very first character of the very first line.

techee:
Should be fixed now. I've also rewritten the patch to make it more maintainable and moved the code to a separate file.

alukreis:
Seems good to me, thanks for doing this by the way.

PR #922, opened 2019-10-06
Disable menu general menu item in case of unsupported file type

### LINKED ISSUE #915: tableconvert: disable menu item on not supported filetypes
Actually the Tableconvert plugin supports the filetypes HTML, LaTeX and SQL. I suggest that the menu item _"Convert to table"_ should be disabled if the current document's filetype is not supported.

frlan:
This PR will fix #915


tableconvert/src/tableconvert.c
```diff
@@ -338,6 +351,31 @@ void cb_table_convert_type(G_GNUC_UNUSED GtkMenuItem *menuitem, G_GNUC_UNUSED gp
 	convert_to_table(TRUE, GPOINTER_TO_INT(gdata));
 }
 
+void cb_table_convert_change_document(G_GNUC_UNUSED GtkMenuItem *menuitem, G_GNUC_UNUSED gpointer gdata)
```
LarsGit223: This is a local function so IMHO it should be static and only declared in ```tableconvert.c``` not in ```tableconvert.h```.

tableconvert/src/tableconvert.c
```diff
@@ -338,6 +351,31 @@ void cb_table_convert_type(G_GNUC_UNUSED GtkMenuItem *menuitem, G_GNUC_UNUSED gp
...
+{
+	set_activate_state();
+}
+
+void set_activate_state()
+{
```
LarsGit223: The same here. This is only a static function.

tableconvert/src/tableconvert.h
```diff
@@ -71,5 +71,7 @@ extern TableConvertRule tablerules[];
 extern void cb_table_convert(G_GNUC_UNUSED GtkMenuItem *menuitem, G_GNUC_UNUSED gpointer gdata);
 extern void cb_table_convert_type(G_GNUC_UNUSED GtkMenuItem *menuitem, gpointer gdata);
 extern void convert_to_table(gboolean header, gint file_type);
+extern void cb_table_convert_change_document(G_GNUC_UNUSED GtkMenuItem *menuitem, gpointer gdata);
+extern void set_activate_state();
```
LarsGit223: I suggest to move this two function declarations above to the beginning of ```tableconvert.c```. And I would remove the not required ```extern```.

tableconvert/src/tableconvert.c
```diff
@@ -338,6 +351,31 @@ void cb_table_convert_type(G_GNUC_UNUSED GtkMenuItem *menuitem, G_GNUC_UNUSED gp
...
+	// getting document
+	GeanyDocument *doc = NULL;
+	doc = document_get_current();
+
+	if (
+		doc != NULL && (
```
LarsGit223: This should also check if there is a current selection or not. The function ```convert_to_table()``` is only doing something if there is a selection (```if (sci_has_selection(doc->editor->sci)) ...```).

tableconvert/src/tableconvert.c
```diff
@@ -35,6 +35,19 @@ PLUGIN_SET_TRANSLATABLE_INFO(
...
+PluginCallback plugin_callbacks[] =
+{
+	{ "editor-notify", (GCallback) &cb_table_convert_change_document, FALSE, NULL },
+	{ "document-activate", (GCallback) &cb_table_convert_change_document, FALSE, NULL },
+	{ "document-filetype-set", (GCallback) &cb_table_convert_change_document, FALSE, NULL },
+	{ "document-new", (GCallback) &cb_table_convert_change_document, FALSE, NULL},
+	{ "geany-startup-complete", (GCallback) &cb_table_convert_change_document, FALSE, NULL },
+	{ "document-close", (GCallback) &cb_table_convert_change_document, FALSE, NULL},
+	{ NULL, NULL, FALSE, NULL }
+};
```
LarsGit223: Do you really need all of this callbacks? I guess at least ```geany-startup-complete``` is not required or am I overseeing something?

LarsGit223:
@frlan: it would be nice if you could prefix the commit message with ```tableconvert: ```

tableconvert/src/tableconvert.c
frlan: Good idea -- but not yet sure it's  really needed.

LarsGit223: Well, the menu item is doing nothing if there is no selection. IMHO it should check the selection or if not the user should get a message that a selection is needed before selecting the menu item.

frlan: It's the wrong place for getting a selection

LarsGit223: Can you explain what you mean please? I mean it would be good to check if there is a selection, not really getting it (there is the ```doc``` pointer and that's all we need?).

frlan: At this point yes. It's either the menu is active or not. The actually handling i done in convert_to_table().

tableconvert/src/tableconvert.c
LarsGit223: I guess that's fine now.

tableconvert/src/tableconvert.c
LarsGit223: Should still be static. Unless you prefer it that way.

tableconvert/src/tableconvert.c
LarsGit223: Ok, when the events are fired, a document is most likely just opened or switched to so there most likely cannot be a selection (yet).

tableconvert/src/tableconvert.h
```diff
@@ -70,6 +70,6 @@ extern TableConvertRule tablerules[];
 
 extern void cb_table_convert(G_GNUC_UNUSED GtkMenuItem *menuitem, G_GNUC_UNUSED gpointer gdata);
 extern void cb_table_convert_type(G_GNUC_UNUSED GtkMenuItem *menuitem, gpointer gdata);
-extern void convert_to_table(gboolean header, gint file_type);
+extern void cb_table_convert_change_document(G_GNUC_UNUSED GtkMenuItem *menuitem, gpointer gdata);
```
LarsGit223: See above. Should be static and removed from the header file.

tableconvert/src/tableconvert.h
```diff
@@ -27,7 +27,7 @@
 
 #include <geanyplugin.h>
 #include <gtk/gtk.h>
-
+#include <geany/filetypes.h>
```
LarsGit223: This is not required. I removed it and can still build Geany-Plugins. Also it makes the Travis-CI build fail for some reason (but I don't know why).

PR #763, opened 2018-08-06
Utilities for Pair Tag Highlighter

sagarchalise:
I think these are useful to working with xml pair tags.

- Editor Menu option for jumping to matching tag
- Selecting to matching tag

PR #926, opened 2019-10-11
Utilities for Pair Tag Highlighter

sagarchalise:
I think these are useful to working with xml pair tags.

- Editor Menu option for jumping to matching tag
- Selecting to matching tag

Related to: https://github.com/geany/geany-plugins/pull/763


LarsGit223:
@vmkononenko: please have a look.

PR #927, opened 2019-10-12
treebrowser: fix icons in the tree and the toolbar

techee:
I noticed several problems with the tree browser:

1. The green toolbar icon with my theme corresponds to a document icon but it should be an action icon
Screen Shot 2019-10-12 at 20 53 41
2. Expanded folder icon is wrong and corresponds to document open action icon instead of a file type icon
Screen Shot 2019-10-12 at 20 54 03
3. On MacOS folder icons are themed differently than e.g. ProjectOrganizer or File browser plugins.

This pull request addresses these issues and makes some cleanups of the code in addition.


treebrowser/src/treebrowser.c
```diff
@@ -180,75 +180,35 @@ tree_view_row_expanded_iter(GtkTreeView *tree_view, GtkTreeIter *iter)
...
-										gtk_widget_get_default_direction(),
-										GTK_STATE_NORMAL, GTK_ICON_SIZE_MENU, NULL, NULL);
-	return NULL;
-}
-#endif
-
-static GdkPixbuf *
-utils_pixbuf_from_path(gchar *path)
+static GIcon *
+utils_gicon_from_path(gchar *path)
```
LarsGit223: The parameter path could be const.

treebrowser/src/treebrowser.c
```diff
@@ -180,75 +180,35 @@ tree_view_row_expanded_iter(GtkTreeView *tree_view, GtkTreeIter *iter)
...
 	ctype = g_content_type_guess(path, NULL, 0, NULL);
 	icon = g_content_type_get_icon(ctype);
 	g_free(ctype);
 
 	if (icon != NULL)
 	{
-		gtk_icon_size_lookup(GTK_ICON_SIZE_MENU, &width, NULL);
-		info = gtk_icon_theme_lookup_by_gicon(gtk_icon_theme_get_default(), icon, width, GTK_ICON_LOOKUP_USE_BUILTIN);
-		g_object_unref(icon);
+		info = gtk_icon_theme_lookup_by_gicon(gtk_icon_theme_get_default(), icon, 16, 0);
```
LarsGit223: What is usually used for icon sizes? You entered 16 but what about using a gtk constant like before?

techee: Notice that the constant GTK_ICON_LOOKUP_USE_BUILTIN  was used for the last parameter which is flags - I believe this is why the icons looked differently on macOS so I used 0.

The width was determined by https://developer.gnome.org/gtk3/stable/gtk3-Themeable-Stock-Images.html#gtk-icon-size-lookup but based on the documentation this function shouldn't be needed and GTK_ICON_SIZE_MENU is 16px. I use the same in ProjectOrganizer and it seems to work fine.

treebrowser/src/treebrowser.c
techee: OK, will do.

treebrowser/src/treebrowser.c
LarsGit223: Ok, then we leave it as-is. I didn't doubt that it works I just wondered if there is a coding style for stuff like this which says 'use constants or literals.

techee:
I've added a few more commits because icons weren't shown in the popup menu on GTK3 - see the new commits for more info.

Issue #545, opened 2017-03-14
[pairtaghighlighter] Make highlight color and opacity configurable


mad-doctor:
When using dark color schemes the pair tag highlighting completely obscures tag attributes. It will be nice if the highlight color and opacity is configurable through configuration file and/or some simple UI (e.g. similar to git-changebar plugin).

![pair_tag_highlight](https://cloud.githubusercontent.com/assets/421991/23891197/d8cb0e6e-089d-11e7-9d3a-7e9cfc8cea39.png)

PatrickGMM:
Added a pull request (#940) to address this since I needed the option as well.

PatrickGMM:
Would anyone be willing to give my pull request a try on Windows and/or Mac? I've tested it on Linux so far.

PR #940, opened 2020-01-06
Pair Tag Highlighter - Added a color preferences dialog (Issue #545)

PatrickGMM:
Added a preferences dialog that allows the highlight colors and overall color alpha to be changed.

* Settings are kept in a .conf file
* Tested on linux mint 19.1 cinnamon, geany 1.35 with gtk2 v2.24 and gtk3 v3.22

### DIFF
--- a/build/pairtaghighlighter.m4
+++ b/build/pairtaghighlighter.m4
@@ -5,5 +5,6 @@ AC_DEFUN([GP_CHECK_PAIRTAGHIGHLIGHTER],
     AC_CONFIG_FILES([
         pairtaghighlighter/Makefile
         pairtaghighlighter/src/Makefile
+        pairtaghighlighter/data/Makefile
     ])
 ])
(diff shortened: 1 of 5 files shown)


elextr:
Afraid the CI fail is pair tag highligher.

Also probably good to follow the style of the original code eg `f()` not `f ()` but this being a plugin its up to the maintainer to enforce (or not :).

codebrainz:
`strtol()` requires including `stdlib.h`, but you could also use one of the GDK functions such as [`gdk_rgba_parse()`](https://developer.gnome.org/gdk3/stable/gdk3-RGBA-Colors.html#gdk-rgba-parse) to parse the colour string or expose [one of the functions](https://github.com/geany/geany/blob/5cc69b3d6728a714637caabf0238759363e91e85/src/utils.h#L284) already in Geany to the plugin API and use that.

PatrickGMM:
I see this is pending fixes. Can someone let me know what fixes I need to perform?

elextr:
@PatrickGMM as a minimum the PR has conflicts and is not mergable as is, not sure what other fixes, but I guess the plugin maintainer should comment too.

PR #947, opened 2020-02-02
cppcheck: Set Geany include path for it to find header files

b4n:
This fixes build with recent versions of *cppcheck* with some plugins using `PLUGIN_VERSION_CHECK()` macro, like the *latex* plugin.

---

@eht16 already tested this with the nightly build environment, but I'm PRing this to make sure the CI tests here also pass.  And extra review is always welcome ;)

### DIFF
--- a/build/compat.m4
+++ b/build/compat.m4
@@ -1,3 +1,9 @@
+dnl taken from Autoconf's m4sh.m4, GPLv3+
+m4_ifndef([AS_VAR_COPY], [
+m4_define([AS_VAR_COPY],
+[AS_LITERAL_WORD_IF([$1[]$2], [$1=$$2], [eval $1=\$$2])])
+])
+
 dnl taken from pkg-config's pkg.m4, GPLv2+
 m4_ifndef([PKG_CHECK_VAR], [
 AC_DEFUN([PKG_CHECK_VAR],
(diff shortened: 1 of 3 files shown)


b4n:
> I'm PRing this to make sure the CI tests here also pass.

…which was a good thing as apparently our CI has a too old pkg-config that doesn't provide `PKG_CHECK_VAR()`.  I thus did similar to what cyptsetup [did in its](https://gitlab.com/cryptsetup/cryptsetup/commit/03581fc8ccf1acdaecbd808c52f48606b785e59f)[ own side](https://gitlab.com/cryptsetup/cryptsetup/commit/eb0293a6378b71e038f47440e3ddf6d68d33d535), that is providing copies when missing.

eht16:
LGBI

build/compat.m4
```diff
@@ -0,0 +1,18 @@
+dnl taken from Autoconf's m4sh.m4, GPLv3+
+m4_ifndef([AS_VAR_COPY], [
+m4_define([AS_VAR_COPY],
+[AS_LITERAL_WORD_IF([$1[]$2], [$1=$$2], [eval $1=\$$2])])
+])
```
b4n: I'm not totally sure about this part: it might not be needed, and if it is I'm not sure `AS_LITERAL_WORD_IF` would be available.  This code is what Autoconf has nowadays, but the inspiration I got from cryptsetup has `AS_LITERAL_IF` instead (but it seems to be buggy in some cases if I trust the documentation).
Anyway, we can fix this if a problem arises

PR #951, opened 2020-02-08
latex: added feature: indent line in new environment (fixes #950)

### LINKED ISSUE #950: latex: new environments don't respect indentation width
I just switched to geany for my main latex editor together with the latex plugin, and I really appreciate the great work you guys do there!

When starting a new environment by typing `\begin{align*}` and pressing enter, the corresponding `\end{align*}` gets automatically created - which is awesome. The current line gets indented, which is also good, but only with one space. My indentation preferences are set to indent with 4 spaces, and it would be cool if the latex plugin would use this preference when opening a new environment. When using the "insert environment" option from the menu, no indentation happens at all. Maybe this can be fixed?

LukeLR:
This adds the feature requested in #950 to respect the user preference for line indentation. This is my first contribution to this project, so feel free to review and give feedback :)

### DIFF
--- a/latex/src/latex.c
+++ b/latex/src/latex.c
@@ -604,6 +604,7 @@ static gboolean on_editor_notify(G_GNUC_UNUSED GObject *object, GeanyEditor *edi
 						gint line_len = sci_get_line_length(sci, line);
 						gint i, start;
 						gint indent;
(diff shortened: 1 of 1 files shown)


frlan:
Looks good from reading. Not yet tested -- stay tuned ;)

PR #952, opened 2020-02-08
commander: Silence expected deprecation warnings

b4n:
We have to use the deprecated API because it is the only way to fetch the correct information from menus that were themselves created with that deprecated API, so we don't care about deprecation here.

Closes #865, and see there for details.


eht16:
@b4n do we want to merge this or carry it around for yet another release cycle? :)

b4n:
Yeah let's get it in not to have other people waste time trying to "fix" it.

This won't remove all deprecation warnings (the plugin still uses `gtk_vbox_new()`) though.  I can very easily fix that now we depend on GTK3 anyway though, but that can be for later.

PR #953, opened 2020-02-11
Commander like file search on keyboard shortcut find file in projects

sagarchalise:
This is mostly my preference but I think it would be quite useful to have file search feature similar to sublime or vscode etc. 
So, I used codebase of commander mostly and modified it.
Basically:

- Find File keyboard shortcut with open a dialog with input and search against files and showing them
- Enter would open the file 
- Workflow from UI would be as is and this is only based on keyboard shortcut

P.S. This is mostly a proof of concept and I think someone with better knowledge of gtk, C and geany api can pull this off better than what I currently have.

### DIFF
--- a/projectorganizer/src/prjorg-menu.c
+++ b/projectorganizer/src/prjorg-menu.c
@@ -49,7 +49,46 @@ enum
 
 static GtkWidget *s_fif_item, *s_ff_item, *s_ft_item, *s_shs_item, *s_sep_item, *s_context_osf_item, *s_context_sep_item;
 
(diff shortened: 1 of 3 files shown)


andy5995:
This is different from the file search already available? https://github.com/geany/geany/discussions/3748

elextr:
#3748 is select among already open files in the document sidebar, IIUC this is for all files in the project organiser tree and opens ones not already open.  But it doesn't seem to have attracted the attention of the PO maintainer, maybe because it isn't complete.

PR #956, opened 2020-02-23
Make libgit2 version preprocessor conditionals compatible with libgit…

Doeme:
…2-0.99

LIBGIT2_SOVERSION is defined as string literal, e.g. "0.99",
from libgit2-0.99 and beyond. Arithmetic checks against this
variable whill hence fail. This patch switches the checks to
compare against the LIBGIT2_VER_* family, which should be more stable.


LarsGit223:
LGBI and build passed, fine for me.

PR #969, opened 2020-04-24
GeanyVC: Add support for Fossil SCM (fossil-scm.org)

### LINKED ISSUE #968: GeanyVC: Add support for Fossil SCM
I'm using [Fossil SCM](https://fossil-scm.org) with most of my projects, as it's free, quite lightweight yet mighty just like Geany :). Apart from distributed Version Control, Fossil also packs Issue tracker, Wiki, Forum, and Server all in single binary executable. So pretty much all I need for _my_ projects.

I recently found myself at an OSX machine without all of my usual toolbox ... that how I discovered Geany (BTW, excellent packaging for OSX)! But to my surprise GeanyVC did not support Fossil...

So here it is, it shouldn't be hard to add Fossil support to GeanyVC, since it mostly has similar commands and  behavior as Git and Bazzar.

nomadbyte:
Fixes #968

### DIFF
--- a/geanyvc/src/vc_fossil.c
+++ b/geanyvc/src/vc_fossil.c
@@ -1,0 +1,590 @@
+/*
+ *      Copyright 2020 Artur Shepilko <nomadbyte@gmail.com>
+ *      Copyright 2007-2011 Frank Lanitz <frank(at)frank(dot)uvena(dot)de>
+ *      Copyright 2007-2009 Enrico Tröger <enrico.troeger@uvena.de>
+ *      Copyright 2007 Nick Treleaven <nick.treleaven@btinternet.com>
+ *      Copyright 2007-2009 Yura Siamashka <yurand2@gmail.com>
+ *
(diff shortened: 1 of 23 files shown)


geanyvc/src/vc_fossil.c
```diff
@@ -0,0 +1,590 @@
...
+		if (std_err)
+		{
+			std_err1 = *std_err;
+			*std_err = NULL;
+		}
+
+		ret = execute_custom_command(base_dir, (const gchar **) argv, (const gchar **) NULL,
+							std_out, std_err, path, list, message);
+
+		if (std_out1)
```
frlan, line 329: ActuallyI I'm unsure what this is doing.

nomadbyte: This appends the output of the second command to the output from the first command.

frlan: Ah. Somehow went confused

PR #974, opened 2020-05-01
Update of Russian translation for the Addons plugin

dmitryunruh:
The translations for the untranslated (into Russian language) strings of the Addons plugin have been added.


Skif-off:
Why "выноску"? Why not "подсказку"?

Geany uses "подсказка" too: https://github.com/geany/geany/blob/master/po/ru.po#L3390

P.S. Thanks! :) I tried to translate (last attempt about a ~year ago), but I was not happy with the result in the ```workbench``` plugin...

dmitryunruh:
I doubted in this word too. But I did not find a better translation of the word "calltip" (with online translators).
I associated the word "подсказка" more with text.
Both of these words are not informative enough.
I suggest the following: 1. Replace the word "выноску" with the word "подсказку". 2. Add a more detailed description to this options (using a tooltip) in one of the next PRs.

Skif-off:
I found my ru.po (or maybe one of them): I used "Копировать путь файла" because the "Copy File Path" command returns full path of file (```path``` + ```name```), not just ```path```, and "значение цвета" instead "цветовой код" (but I don't remember why, "цветовой код" or "код цвета" are used very often).

Also I replaced all "клик" with "щелчок", because "клик" is jargon.

dmitryunruh:
"Копировать путь файла" instead of "Копировать путь к файлу"; also not obvious enough. Maybe "Копировать полный путь файла" will be more suitable in this case?

"Значение цвета" instead of "Цветовой код": "Цветовой код" ("#ABCDEF" for example) is more obvious because such words like "black" is a color value ("значение цвета") too but hovering on it does not show a calltip.

"Клик" instead of "Щелчок": 1. "Клик" is a well-established word in computer documentation (in Russian language). 2. This word is used in the previous option: "Подсвечивать все вхождения слова в документе по двойному клику".

frlan:
@Skif-off  @dmitryunruh Would you mind if I just merge this suggestion and during the next release cycle at latest there will be an further improved PR? ;)

dmitryunruh:
@frlan I don't mind.

PR #980, opened 2020-05-20
Sundry fixes and enhancements to the keystrokes recorder plugin

brackleian:
Uses the Geany "key-press" signal, allowing access to all keys, including bound ones.
Fixes a bug that prevent the use of bound keys without modifiers.
Prevents playback during recording.
Displays recording indicator on the status bar.

### DIFF
--- a/keyrecord/src/keyrecord.c
+++ b/keyrecord/src/keyrecord.c
@@ -42,6 +42,19 @@ static GdkEventKey** recorded_pattern;
 static guint recorded_size;
 int CAPACITY = 2;
 GeanyKeyBinding *record, *play;
+
+static GtkStatusbar *geany_statusbar;
+static GtkWidget *keyrecord_statusbar_box;
+static GtkLabel *keyrecord_state_label;
+
+static void update_status()
+{
+	if (recording)
+		gtk_widget_show(keyrecord_state_label);
+	else
+		gtk_widget_hide(keyrecord_state_label);
+}
+
 static gboolean is_record_key(GdkEventKey *event)
 {
 	return event->keyval == record->key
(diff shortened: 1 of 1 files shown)


codebrainz:
You should compile with at least `-Wall -Wextra -Werror` and make sure there's no warnings/errors. Sometimes it's useful to add `-Wno-unused-parameter` and `-DGLIB_DISABLE_DEPRECATION_WARNINGS ` to filter out the noise. You can also change into the plugin directory and do `make` from there to avoid getting bombarded by messages from other plugins.

brackleian:
On Tue, 19 May 2020 21:48:05 -0700
Matthew Brush <notifications@github.com> wrote:

> You should compile with at least `-Wall -Wextra -Werror` and make sure there's no warnings/errors. Sometimes it's useful to add `-Wno-unused-parameter` and `-DGLIB_DISABLE_DEPRECATION_WARNINGS ` to filter out the noise. You can also change into the plugin directory and do `make` from there to avoid getting bombarded by messages from other plugins. 

Thanks; will do that.  I was just trying to figure out how to reconfigure for GTK2; my previous build had --enable-gtk3, and now I can't seem to make it not use GTK3.

codebrainz:
> my previous build had --enable-gtk3, and now I can't seem to make it not use GTK3.

You should be able to put `--disable-gtk3` to force GTK+2 build. You'll of course have to install the dependencies for GTK+2 if you don't have them installed (ex. on Ubuntu `libgtk2.0-dev` package for GTK+ and `libvte-dev` for Terminal).

brackleian:
OK, GTK2 builds correctly too.

brackleian:
Anything else I need to do, or is it just wait until reviewing is done?

elextr:
Plugin are maintained by individuals (see MAINTAINERS file) so likely have to wait for the maintainer to comment.

PR #984, opened 2020-05-22
Update README - "Mini Script"

PeterMortensen:
"gms" is an abbreviation used in this document - "gms" is not a kind of Mini Script; it ***IS*** Mini Script...

Used uppercase for the abbreviation as it is not the name of an executable. And properly introduced the abbreviation.

Used the name "Mini Script" as that is what it is called in the Plugin Manager - though it is currently not consistently spelled in the software.

Dropped the word "configuring" as there isn't any one-time configuration part - only the options/settings in the Mini Script window.

General copy editing (e.g. ref. <https://en.wikipedia.org/wiki/Perl>, <https://en.wikipedia.org/wiki/Python_%28programming_language%29>, and <https://en.wikipedia.org/wiki/AWK>).

### DIFF
--- a/geanyminiscript/README
+++ b/geanyminiscript/README
@@ -9,9 +9,9 @@ gms
 About
 =====
 
-gms is a Geany Mini-Script filter plugin
+Mini Script (GMS) is a filter plugin.
 
-In order to use it you need Geany 0.15 or later.
+In order to use it, you need Geany 0.15 or later.
 
 Features
 ========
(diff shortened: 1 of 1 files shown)


PeterMortensen:
Can you reject this pull request?

I would like to retract this pull request and submit a new one. The configure part is real - the how-to  just needs to documented.

(Is there a squash thing to hide the first revision?)

frlan:
Just force push to your branch should do the trick

eht16:
You can close your own PRs by yourself at the bottom of the page.
And as @frlan said, you can also squash and force-push to clean up the previous attempts. Google will tell you how to this :).

Issue #944, opened 2020-01-15
Line Operations: please add support for version sorting


ftrebien:
I would like to kindly request support for [version sorting](https://www.gnu.org/software/coreutils/manual/html_node/Details-about-version-sort.html) when using the Line Operations plug-in. It is useful for sorting strings that mix words and numbers, which are quite common, not only software version numbers.

PR #993, opened 2020-08-02
Lineoperations: version compare

smostertdev:
This PR is in response to #944.

Changes:
- Changed preferences to radio buttons to support multiple compare types.
lineops-version-compare-pref
- add "version compare" using filevercmp from coreutils

Notes:
- filevercmp is copied directly from coreutils's sort
https://git.savannah.gnu.org/cgit/gnulib.git/tree/lib/filevercmp.c

I'm open to hear any suggestions/improvements.


elextr:
Unfortunately filevercmp.c is GPL 3 and line operations is GPL 2 and those can't be combined, [see GNU](https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility)

smostertdev:
Can I update the lineoperations license to GPLv3+ without issues? I guess I would have to get permission from anybody who made changes to lineoperations under the old license.
Would changing this license to GPLv3 cause problems in the future?

elextr:
> I guess I would have to get permission from anybody who made changes to lineoperations under the old license.

That is what other projects that have changed licenses did yeah, my glance at git said not many people, and hopefully they will reply yes to a request for permission.

> Would changing this license to GPLv3 cause problems in the future?

Doubt it, there is at least one plugin with GPL3 license already (commander) and Geany itself is GPL2+ which is compatible.

smostertdev:
I am closing this since I cannot merge it until I get this licensing resolved. I will open it if I figure things out.

nomadbyte:
Could the version-sort be done with `strverscmp()` instead of `filevercmp()` from coreutils?

There're some differences, but for the general use like "file-1.9" vs. "file-1.10" the `strverscmp()` should do just fine. And it's part of `glibc` (well, `#define _GNU_SOURCE 1`), so no need to copy the original sources into the project and thus deal with this license harmonization issue.

PR #1015, opened 2020-10-14
TreeBrowser: Avoid double node clear

### LINKED ISSUE #675: [treebrowser] Intermittent crashing when deleting a file
I'm running Geany 1.32 with treebrowser as my only plugin.  I've been using it more heavily than usual today to create files and delete files, and I've had three crashes just today.  Here is a stack trace that I'm grabbing from Fedora's abrt application.

[treebrowser-dump.txt](https://github.com/geany/geany-plugins/files/1636190/treebrowser-dump.txt)
https://retrace.fedoraproject.org/faf/reports/1993243/

ChrisJAllan:
Fixes #675 

`on_treeview_changed` is triggered during `treebrowser_tree_store_iter_clear_nodes`, leading to a double-delete.

Consistent (as far as I can tell) reproduction steps:
1. Create a folder
2. Create two new files in folder
3. Delete second file


elextr:
ping @medovina

medovina:
I can reproduce this crash.  Thanks for the pull request - it's a busy time but I should be able to look at this in the next day or two.

medovina:
The change seems fine to me.  I can confirm that it fixes the crash.

PR #1017, opened 2020-10-17
Debugger: Fix cppcheck false positive warning

### LINKED ISSUE #1014: Debugger plugin: potential cppcheck false positive: uninitvar
cppcheck 2.2 complains about many `ifdef`s in the Debugger plugin:
```
make[5]: Leaving directory '/build/geany-plugins-1.36-1+20201013git06a38f4/debugger/src'
tpage.c:154:62: warning: Uninitialized variable: widgets [uninitvar]
   gtk_container_remove(GTK_CONTAINER(gtk_widget_get_parent(*widgets[i])), *widgets[i]);
                                                             ^
nofile:0:0: warning: Too many #ifdef configurations - cppcheck only checks 12 configurations. Use --force to check all configurations. For more details, use --enable=information. [toomanyconfigs]

make[4]: *** [Makefile:1686: check-cppcheck] Error 2
```

When we remove GTK2 code, this might resolve it.

What should we do until then?
- use `--force`?
- refactor the code?
- ignore?

This breaks the nightly builds for Debian Sid: https://www.geany.org/download/nightly-builds/.

eht16:
This fixes a cppcheck 2.2+ confusion about uninitialized widgets array
and also makes the code a bit more readable.
Closes #1014.

### DIFF
--- a/debugger/src/tpage.c
+++ b/debugger/src/tpage.c
@@ -142,19 +142,16 @@ void tpage_pack_widgets(gboolean tabbed)
 	GList *children = gtk_container_get_children(GTK_CONTAINER(tab_target));
 	if (children)
 	{
-		int i;
-
 		oldroot = (GtkWidget*)children->data;
-		
+
 		/* unparent widgets */
-		i = 0;
-		while (widgets[i])
+		for (int i = 0; widgets[i]; i++)
 		{
-			g_object_ref(*widgets[i]);
-			gtk_container_remove(GTK_CONTAINER(gtk_widget_get_parent(*widgets[i])), *widgets[i]);
-			i++;
+			GtkWidget **widget_ref = widgets[i];
+			g_object_ref(*widget_ref);
+			gtk_container_remove(GTK_CONTAINER(gtk_widget_get_parent(*widget_ref)), *widget_ref);
 		}
-		
+
 		g_list_free(children);
 	}
 	
@@ -265,11 +262,9 @@ void tpage_pack_widgets(gboolean tabbed)
 
 	if (oldroot)
 	{
-		int i = 0;
-		while (widgets[i])
+		for (int i = 0; widgets[i]; i++)
 		{
 			g_object_unref(*widgets[i]);
-			i++;
 		}
 		gtk_container_remove(GTK_CONTAINER(tab_target), oldroot);
 	}


eht16:
An alternative solution would be:
```c
		while (widgets[i])
		{
				g_object_ref(*(widgets[i]));
				gtk_container_remove(GTK_CONTAINER(gtk_widget_get_parent(*(widgets[i]))), *(widgets[i]));
				i++;
		}
```
to make `cppcheck` happy but I think the solution in this PR is more readable for humans.

The change in the second loop is not strictly necessary for `cppcheck` but is more consistent with the first loop where the same array is used.

debugger/src/tpage.c
```diff
@@ -139,22 +139,23 @@ void tpage_pack_widgets(gboolean tabbed)
...
 		{
-			g_object_ref(*widgets[i]);
-			gtk_container_remove(GTK_CONTAINER(gtk_widget_get_parent(*widgets[i])), *widgets[i]);
+			widget = widgets[i];
+			g_object_ref(widget);
+			gtk_container_remove(GTK_CONTAINER(gtk_widget_get_parent(*widget)), *widget);
 			i++;
-		}
-		
+		} while (widget);
```
b4n, line 157: I'd rather see it as
```c
for (i = 0; widgets[i]; i++)
{
	GtkWidget **widget_ref = widgets[i];
	g_object_ref(*widget_ref);
	gtk_container_remove(…);
}
```

It's more idiomatic and understandable to me than relying on the variable set in the loop.

debugger/src/tpage.c
```diff
@@ -266,11 +267,12 @@ void tpage_pack_widgets(gboolean tabbed)
...
 	{
 		int i = 0;
-		while (widgets[i])
+		do
 		{
+			widget = widgets[i];
 			g_object_unref(*widgets[i]);
 			i++;
-		}
+		} while (widget);
```
b4n, line 275: what's the point here?  You're not using the variable you set in the loop.

b4n (changes requested):
OK, so it's a bug in cppckeck that gets operator priority wrong, right?  Because the code was right, `*array[i]` is `*(array[i])`, not `(*array)[i]`.

Anyway, I think there could be a better loop, and there's at least an issue with the `g_object_ref()` calls.

debugger/src/tpage.c
```diff
@@ -139,22 +139,23 @@ void tpage_pack_widgets(gboolean tabbed)
...
+
 		/* unparent widgets */
-		i = 0;
-		while (widgets[i])
+		do
 		{
-			g_object_ref(*widgets[i]);
-			gtk_container_remove(GTK_CONTAINER(gtk_widget_get_parent(*widgets[i])), *widgets[i]);
+			widget = widgets[i];
+			g_object_ref(widget);
```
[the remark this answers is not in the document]
eht16: Ouch, of course.

debugger/src/tpage.c
eht16: Thanks for the idea and snippet, that's way cleaner!

debugger/src/tpage.c
eht16: Being confused by rewriting the code in many variants in too short time or so.
No idea what's happened to me there...:(

eht16:
I rewrote the code again using the `for` loop, fixed the `g_object_ref()` calls and tested again functionality and cppcheck happiness.

debugger/src/tpage.c
```diff
@@ -142,19 +142,18 @@ void tpage_pack_widgets(gboolean tabbed)
 	GList *children = gtk_container_get_children(GTK_CONTAINER(tab_target));
 	if (children)
 	{
-		int i;
-
-		oldroot = (GtkWidget*)children->data;
+		GtkWidget **widget_ref = NULL;
```
b4n, line 145: I'd rather have this declared in the loop

debugger/src/tpage.c
```diff
@@ -265,11 +264,9 @@ void tpage_pack_widgets(gboolean tabbed)
 
 	if (oldroot)
 	{
-		int i = 0;
-		while (widgets[i])
+		for (int i = 0; widgets[i]; i++)
 		{
 			g_object_unref(*widgets[i]);
```
b4n, line 269: Is cppcheck happy with that?  Actually I'm still confused on what it didn't like before, so maybe it's happy. It's correct anyway so whichever way's OK with me.

eht16: It is.
Only `gtk_container_remove(GTK_CONTAINER(gtk_widget_get_parent(*widgets[i])), *widgets[i]);` was problematic, I'm not sure why only here.

debugger/src/tpage.c
eht16: Done in 516277c184f2019e2e504f1d598b008cb6695358.

debugger/src/tpage.c
b4n: Yeah I was just testing manually building cppcheck 2.2 :)

b4n (approved):
LGTM, should not change any behavior, and cppcheck is happy now.

b4n:
BTW I just tested cppcheck master and it's happy without the patch as well, so they fixed it.
Not saying this shouldn't land, it's fine a change.

eht16:
Great and great! :)

b4n:
Wait… nope I messed up, `master` branch is not the tip anymore, they switched to `main`.  I'll do the checking again, but I'm less optimistic that they have it fixed.

b4n:
So it's not fixed in latest cppcheck, but that shouldn't stop merging this, quite the contrary :)

PR #981, opened 2020-05-21
Add link to how to install, fix some English

dandv:
Progress towards https://github.com/geany/geany/issues/2508

PR #1021, opened 2020-11-02
#981: ReadMe install instructions

eht16:
Follow-up of #981 to fix restructuredText formatting and add the install instruction link to other plugins' READMEs as well which are similarly structured.

Closes https://github.com/geany/geany/issues/2508.


dandv:
Can we use Markdown like everyone else?

PR #1024, opened 2020-11-08
PrettyPrinter: Don't try to free const gchar* input_buffer

frlan:


b4n (changes requested):
The problem is not freeing `input_buffer` which *should* be freed, but that the `input_buffer` variable shouldn't be `const` (since 3fad1285599766a7932667688871c04610bd8651).  The change you propose would lead to leaking quite a lot of memory.

frlan:
@b4n  True. Will update it

b4n (approved):
LGTM

PR #1025, opened 2020-11-13
GeanyVC: Support the use of external diff-tool in directory context

### LINKED ISSUE #573: Directory diff ignores git setting
The docs state that if you have a visual diff tool installed, like meld, running VC->*->Diff will use that tool. This appears to work for VC->File->Diff, but not for VC->Directory->Diff, which spits out the default diff output.

Thinking that GeanyVC was using `git diff`, I tried overriding the `git diff` with:

    git config --global diff.external meld

but this has no effect, since GeanyVC appears to ignore git's `diff.external` setting, preventing the use of a visual diff tool.

nomadbyte:
Fixes #573.

### DIFF
--- a/geanyvc/src/geanyvc.c
+++ b/geanyvc/src/geanyvc.c
@@ -201,6 +201,12 @@ free_commit_list(GSList * lst)
 	g_slist_free(lst);
 }
 
+int
+commititem_compare_by_path(const CommitItem * a, const CommitItem * b)
+{
+	return (g_strcmp0(a->path, b->path));
+}
+
 gchar *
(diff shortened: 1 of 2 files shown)


nomadbyte:
I just brought it up-to-date with the current master. Anything else to get this one merged in? ... It's been hanging on for quite a while.

elextr:
@nomadbyte since you are listed as maintainer I guess if you say its ok, it is, and on your head be it :-)

nomadbyte:
:) on my head it has been for over a year by now, so I'd rather not keep it in my head and on my outstanding PR list.

I did test it with Meld today, also tested with some other diff'ers back then.

PR #1028, opened 2020-11-13
GeanyVC:(Git) Fallback to `git rev-parse` for getting the repo/worktree's top directory

### LINKED ISSUE #285: GeanyVC: Support for git worktree
I was using [`git-new-workdir`](http://nuclearsquid.com/writings/git-new-workdir/) and recently switched to using the newer form [`git worktree`](https://github.com/blog/2042-git-2-5-including-multiple-worktrees-and-triangular-workflows) however found this stops the plugin working.

The reason appears to be that instead of  a `.git` directory in the 'worktee directory' there is a `.git` file that contains a reference to the original directory. e.g. `gitdir: /home/ubuntu/origdir.git/.git/worktrees/newdir.git`.

nomadbyte:
Fixes #285.

### DIFF
--- a/geanyvc/src/vc_git.c
+++ b/geanyvc/src/vc_git.c
@@ -28,7 +28,37 @@ extern GeanyData *geany_data;
 static gchar *
 get_base_dir(const gchar * path)
 {
-	return find_subdir_path(path, ".git");
+	gchar *base_dir = NULL;
+	const gchar *argv[] = { "git", "rev-parse", "--show-toplevel", NULL };
+	gchar *dir = NULL;
+	gchar *filename = NULL;
+	gchar *std_out = NULL;
+	gchar *std_err = NULL;
+
+	base_dir = find_subdir_path(path, ".git");
+	if (base_dir) return base_dir;
+
+	if (g_file_test(path, G_FILE_TEST_IS_DIR))
+		dir = g_strdup(path);
+	else
+		dir = g_path_get_dirname(path);
+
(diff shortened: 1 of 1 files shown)


nomadbyte:
@elextr  This one too has not been getting any merge-love for very much long while. Anything else for me to do to get this off my PR list?

Just re-checked, the behavior complies with the use-case listed in the #285 for what it's worth.

elextr:
> Just re-checked, the behavior complies with the use-case listed in the https://github.com/geany/geany-plugins/issues/285 for what it's worth.

To be clear, I don't use geanyvc, so I won't spend time understanding enough of the plugin code to be able to review the change, similarly setting up a sacrificial git repo to test it.  Especially when related to git workflows and stuff.  

Although it would be preferable that some other user (like the OP of #285) tests it, in the end if it sits too long and nobody does then IMHO if the plugin maintainer says so (and the build passes) it should be merged.  

Can you trigger the new CI by pushing an inconsequential change, after all its a year since it was tested, I'm sure there is a comment that needs improving/adding :-)

nomadbyte:
Here it is -- the commit comment now has the use-case in it too.

@elextr Do you favor a VCS other than Git? Which one?

elextr:
I use git but not geanyvc (since I mostly do C/C++ and that not in Geany).

Ok, build passed, so safe(ish) to merge, since the CI builds all plugins, if one of them fails the whole build fails and so subsequent PRs all show as failing even if they are for a plugin other than the one that failed (see recent problems with Geanylua failing other plugins).

PR #1031, opened 2020-11-15
Windows installer: ignore patch level version when checking for Geany

eht16:
Loosen the G-P Windows installer version check for Geany to require only a matching major and minor version, i.e. we now ignore the patch level.
So for example, Geany-Plugins 1.37 can be installed for Geany 1.37.1 without warnings.

As discussed in geany/geany/issues/2655.


frlan:
@eht16  I assume you did a testbuild with it?

eht16:
Yes, I did.

Issue #964, opened 2020-03-25
Bookmark sidebar does not always match actual bookmarks


dbkaynor:
When clearing the bookmarks using the Documents/Remove Markers option the actual bookmarks are cleared but the sidebar list is not. Also sometimes the same bookmark gets listed multiple times in the sidebar list.

eht16:
Remove Markers just removes any markers, this can be also markers for search highlights and maybe more.
It is not intended that this option also clears the bookmark list.

Could you provide more details for the second problem? When exactly does it happen, steps to reproduce?

dbkaynor:
The issue with bookmarks gets listed multiple times in the sidebar list only happens when the numbed bookmark plugin is installed. I will provide more info when I am able to get it.

PR #1040, opened 2020-11-29
Addons: Bookmarks: Prevent duplicating bookmarks when editing line

eht16:
For some filetypes (more specifically some Scintilla lexers) it might
happen that an event with modificationType SC_MOD_CHANGEMARKER is
sent and so the current line is added as bookmark again.
To avoid this, check if the current line has already a bookmark and
update it in this case.

Might be related to #964.


eht16:
To test: set a bookmark in a C file, find the bookmark in the bookmark list sidebar, edit the bookmark line and see how the bookmark is duplicated in the list.

eht16:
*friendly-ping*

PR #1064, opened 2021-02-09
Markdown: Update using known GeanyDocument when available

### LINKED ISSUE #1062: Markdown Plugin Preview not working after starting geany with a file that was already opened the other day
After starting geany 1.37.1 Debian GNU/Linux 10 (buster) with a markdown-file that was opened already yesterday, the preview is not working (see below) until the moment I edit the file again. Phew, quite difficult to explain exactly, sorry.

The shown 'error' in the preview is: _The current document does not have a Markdown filetype_

Not a big issue for me but I still wanted to post it

codebrainz:
For whatever reason, in some cases `document_get_current()` doesn't
return a valid document when it seems like it should, so when updating
the markdown preview from signals where the related GeanyDocument
is available, use that instead of calling `document_get_current()`.

In other cases, continue to use `document_get_current()` as before.

Closes #1062

### DIFF
--- a/markdown/src/plugin.c
+++ b/markdown/src/plugin.c
@@ -51,7 +51,7 @@ static GtkWidget *g_scrolled_win = NULL;
 static GtkWidget *g_export_html = NULL;
 
 /* Forward declarations */
-static void update_markdown_viewer(MarkdownViewer *viewer);
+static void update_markdown_viewer(MarkdownViewer *viewer, GeanyDocument *doc);
 static gboolean on_editor_notify(GObject *obj, GeanyEditor *editor, SCNotification *notif, MarkdownViewer *viewer);
 static void on_document_signal(GObject *obj, GeanyDocument *doc, MarkdownViewer *viewer);
 static void on_document_filetype_set(GObject *obj, GeanyDocument *doc, GeanyFiletype *ft_old, MarkdownViewer *viewer);
@@ -120,7 +120,7 @@ void plugin_init(GeanyData *data)
    * and when reloaded tries to re-register the GTypes. */
   plugin_module_make_resident(geany_plugin);
 
-  update_markdown_viewer(MARKDOWN_VIEWER(viewer));
+  update_markdown_viewer(MARKDOWN_VIEWER(viewer), NULL);
 }
 
 /* Cleanup resources on plugin unload. */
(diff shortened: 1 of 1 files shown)


elextr:
Do you have any further information about when/why/how `document_get_current()` is not returning valid docs? Given its used all through Geany its a bit of a worry if its not right.

codebrainz:
I do not, I just noticed that it was giving invalid/NULL and that I had a valid document pointer already available. Probably one of the callbacks the plugin uses gets triggered early before the document list is fully initialized or something, but I didn't spend much time trying to understand it.

elextr:
It can definitely return NULL for lots of reasons, but if it returns a pointer it should be valid since it tests it and returns NULL if not valid.

markdown/src/plugin.c
```diff
@@ -192,7 +193,7 @@ static gboolean on_editor_notify(GObject *obj, GeanyEditor *editor,
   SCNotification *notif, MarkdownViewer *viewer)
 {
   if (IS_MOD_NOTIF(notif)) {
-    update_markdown_viewer(viewer);
+    update_markdown_viewer(viewer, NULL);
```
b4n, line 196:
```suggestion
    update_markdown_viewer(viewer, editor->document);
```

b4n:
What your change suggests to me is that you'd encounter cases where `document_get_current()` doesn't return the document for which the signal was fired for.  This sounds weird for the `activate`/`new`/`open`/`reload`/`filetype-set` signals you're using, and definitely something that should be fixed if it's indeed not in sync (e.g. the signal is fired before the changes that make `document_get_current()` return the right value happen) -- unless there is an actual reason for that, but still.

I'm sympathetic to the base idea of using the document for which the signal was fired though.  But it really seems like it should be strictly equal to `document_get_current()`.

markdown/src/plugin.c
```diff
@@ -162,9 +162,10 @@ void plugin_help(void)
  * editor's text contents change and not on other editor events.
  */
 static void
-update_markdown_viewer(MarkdownViewer *viewer)
+update_markdown_viewer(MarkdownViewer *viewer, GeanyDocument *doc)
 {
-  GeanyDocument *doc = document_get_current();
+  if (!DOC_VALID(doc))
```
b4n, line 167: Wouldn't simply `doc == NULL` work here?  Do you really get non-NULL doc pointers that have `doc->valid == FALSE`?

Issue #1089, opened 2021-07-05
geanynumberedbookmarks shortcuts do not work with numlock on


Davidy22:
Installed and enabled geany-plugins on linux mint and enabled and tried the numbered bookmarks plugin, CTRL+shift+number and the respective CTRL+number appear to have no effect in geany 1.36. Can provide additional files on request.

elextr:
WFM

```
09:30:53: Geany INFO		: Geany 1.38 (git >= b524a58e), en_AU.UTF-8
09:30:53: Geany INFO		: GTK 3.24.20, GLib 2.64.6
09:30:53: Geany INFO		: OS: Linux Mint 20 (ulyana)
```

elextr:
"Linux Mint" doesn't actually tell us anything, what version and what desktop?  And check that the desktop isn't eating those keycodes for its purposes.

Davidy22:
oh, more detail:
Mint 20.1, mate version. 
Geany built on or after 2020-03-22
Using GTK+ v3.24.20 and GLib v2.64.6 runtime libraries

Had tried mashing the shortcuts in other contexts to see if anything was bound to them, hadn't seen any effect. A config option to set and test the shortcuts could help maybe.

elextr:
To see if those keycodes are available you can try setting something currently unbound in the standard keybindings preferences and see what you get, you can always unbind it again.

Note I get `<primary><shift>exclam` not `<primary><shift>1` in the keybindings, but probably doesn't matter because it works for me.

elextr:
Also just a note that although I am using a git version, nothing has changed in numbered bookmarks since before 1.36.  @Davidy22 This kind of indicates its in your system since it works here, just to check, please post the GTK/Glib versions from `Help->Debug messages` like I did, its not clear where you got the versions you mentioned, unless its actually what Geany prints it may not be what Geany is using since GTK2 is still present on Mint.

Davidy22:
I grabbed it from the about tab under help, but here's the debug messages version:

16:28:01: Geany INFO		: Geany 1.36, en_US.UTF-8
16:28:01: Geany INFO		: GTK 3.24.20, GLib 2.64.6

Didn't run into collision messages when binding those shortcuts in mint's general keyboard shortcuts utility, did check to see if it does give collision messages as well. I do also notice that I get SHIFT+CTRL+!, SHIFT+CTRL+@, SHIFT+CTRL+#, etc, and not the numbers. If there's a log to be produced I can get it.

[...]

PR #1093, opened 2021-07-21
Make numbered bookmarks work with numlock on

Davidy22:
As resolution to #1089, add a case so that plugin still works when num lock is on.

### DIFF
--- a/geanynumberedbookmarks/src/geanynumberedbookmarks.c
+++ b/geanynumberedbookmarks/src/geanynumberedbookmarks.c
@@ -1452,12 +1452,13 @@ static gboolean Key_Released_CallBack(GtkWidget *widget, GdkEventKey *ev, gpoint
 
 	if(ev->type!=GDK_KEY_RELEASE)
 		return FALSE;
-	/* control+shift+number */
+	/* control pressed */
 	if(ev->state & GDK_CONTROL_MASK) {
 		/* could use hardware keycode instead of keyvals but if unable to get keyode then don't
 		 * have logical default to fall back on
 		*/
 		if(ev->state & GDK_SHIFT_MASK) {
+			/* shift pressed */
 			for(i=0;i<10;i++) if((gint)(ev->keyval)==iShiftNumbers[i])
 			{
 				SetBookMark(doc, i);
--- a/geanynumberedbookmarks/src/geanynumberedbookmarks.c
+++ b/geanynumberedbookmarks/src/geanynumberedbookmarks.c
@@ -1453,26 +1453,25 @@ static gboolean Key_Released_CallBack(GtkWidget *widget, GdkEventKey *ev, gpoint
 	if(ev->type!=GDK_KEY_RELEASE)
 		return FALSE;
 	/* control+shift+number */
-	if(ev->state & (GDK_CONTROL_MASK | GDK_SHIFT_MASK)) {
+	if(ev->state & GDK_CONTROL_MASK) {
 		/* could use hardware keycode instead of keyvals but if unable to get keyode then don't
 		 * have logical default to fall back on
 		*/
-		for(i=0;i<10;i++) if((gint)(ev->keyval)==iShiftNumbers[i])
-		{
-			SetBookMark(doc, i);
+		if(ev->state & GDK_SHIFT_MASK) {
+			for(i=0;i<10;i++) if((gint)(ev->keyval)==iShiftNumbers[i])
+			{
+				SetBookMark(doc, i);
+				return TRUE;
+			}
+		} else {
+			/* control and number pressed */
+			i=((gint)(ev->keyval))-'0';
+			if(i<0 || i>9)
+				return FALSE;
+
+			GotoBookMark(doc, i);
 			return TRUE;
 		}
-
-	}
-	/* control and number pressed */
-	if(ev->state & GDK_CONTROL_MASK)
-	{
-		i=((gint)(ev->keyval))-'0';
-		if(i<0 || i>9)
-			return FALSE;
-
-		GotoBookMark(doc, i);
-		return TRUE;
 	}
 
 	return FALSE;
--- a/geanynumberedbookmarks/src/geanynumberedbookmarks.c
+++ b/geanynumberedbookmarks/src/geanynumberedbookmarks.c
@@ -1464,7 +1464,7 @@ static gboolean Key_Released_CallBack(GtkWidget *widget, GdkEventKey *ev, gpoint
 		return TRUE;
 	}
 	/* control+shift+number */
-	if(ev->state==5) {
+	if(ev->state==5 || ev->state==21) {
 		/* could use hardware keycode instead of keyvals but if unable to get keyode then don't
 		 * have logical default to fall back on
 		*/
--- a/geanynumberedbookmarks/src/geanynumberedbookmarks.c
+++ b/geanynumberedbookmarks/src/geanynumberedbookmarks.c
@@ -1452,19 +1452,8 @@ static gboolean Key_Released_CallBack(GtkWidget *widget, GdkEventKey *ev, gpoint
 
 	if(ev->type!=GDK_KEY_RELEASE)
 		return FALSE;
-
-	/* control and number pressed */
-	if(ev->state==4)
-	{
-		i=((gint)(ev->keyval))-'0';
-		if(i<0 || i>9)
-			return FALSE;
-
-		GotoBookMark(doc, i);
-		return TRUE;
-	}
 	/* control+shift+number */
-	if(ev->state==5 || ev->state==21) {
+	if(ev->state & (GDK_CONTROL_MASK | GDK_SHIFT_MASK)) {
 		/* could use hardware keycode instead of keyvals but if unable to get keyode then don't
 		 * have logical default to fall back on
 		*/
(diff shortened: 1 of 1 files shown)


Davidy22:
What's this about a julia lexxer failing the build? The build worked on my machine and I made the most innocent looking change, were there catastrophic side effects to a || I didn't know about?

elextr:
The Geany-plugins CI is still using an olde distro to build on, Trusty, Geany itself has been updated to use a supported distro because of problems with Trusty, and it was upgraded to Bionic.  (Xenial having reached end of standard life the month before the upgrade).  

What has happened is that the new Julia filetype has leaked a C++17 object `std::string_view` from its Scintilla implementation, but thats ok with Bionic, and the versions used for nightlys, but not Trusty.  Oh well C++17 was bound to leak at some point :-D

@frlan do you want to bring G-P CI distro up to the same standard as Geany?

elextr:
@Davidy22   I would suggest that you use the official mask https://developer.gnome.org/gdk3/unstable/gdk3-Windows.html#GdkModifierType to remove all but the wanted modifiers, any of them can change at any time in any version of GTK, or with any new keyboard with extra keys, so the current implementation is very fragile.

Davidy22:
Now that you mention it, should have figured there would be mask docs when numlock code was exactly 16 away, added

geanynumberedbookmarks/src/geanynumberedbookmarks.c
```diff
@@ -1452,19 +1452,8 @@ static gboolean Key_Released_CallBack(GtkWidget *widget, GdkEventKey *ev, gpoint
...
-		i=((gint)(ev->keyval))-'0';
-		if(i<0 || i>9)
-			return FALSE;
-
-		GotoBookMark(doc, i);
-		return TRUE;
-	}
 	/* control+shift+number */
-	if(ev->state==5 || ev->state==21) {
+	if(ev->state & (GDK_CONTROL_MASK | GDK_SHIFT_MASK)) {
```
elextr, line 1456: This tests for control or shift, not control and shift

eht16:
> The Geany-plugins CI is still using an olde distro to build on, Trusty, Geany itself has been updated to use a supported distro because of problems with Trusty, and it was upgraded to Bionic. (Xenial having reached end of standard life the month before the upgrade).
> 
> What has happened is that the new Julia filetype has leaked a C++17 object `std::string_view` from its Scintilla implementation, but thats ok with Bionic, and the versions used for nightlys, but not Trusty. Oh well C++17 was bound to leak at some point :-D
> 
> @frlan do you want to bring G-P CI distro up to the same standard as Geany?

/me is not @frlan (😄) but anyway: #1094

geanynumberedbookmarks/src/geanynumberedbookmarks.c
Davidy22: Oh word, that wasn't the way I thought it'd work. Thought it was checking right too when I was checking ctrl+number.

elextr:
Just a note that this plugin will only work for keyboards that are similar to US keyboards where the shift characters of the numbers match the majik numbers in the `iShiftNumbers` array ie `)!@#$%^&*(`. 

Since this is a plugin such limitations can be accepted but perhaps the initialisation of that array could be changed to use character literals to be a bit more obvious.

Davidy22:
I actually thought the magic numbers in iShiftNumbers were the punctuation keys too when I first looked at it, but when I was writing the fix I noticed the chunk of code starting line 1503 that seems to be finding out on load what the numbers turn into when you hit shift, the Fraser fellow seems to have thought of that. Still a slightly obtuse magic number list though

elextr:
Ok, the magic number list is only the fallback default and is British keyboard (`"` and &pound; not `@` and `#`).  The code you pointed to is a good try and probably good enough for most cases, but it does assume that the digits are level 0 and shift is level 1, which may not be the case on all keyboards, but probably few enough it won't be a problem very often.

elextr:
@Davidy22 I tried to rerun CI after merging #1094, but although it said the request was queued it doesn't seem to have happened, maybe you could push some inconsequential change (to a comment maybe) to re-trigger it.

Davidy22:
Ey, it passed. Noticed a meaningful comment correction to make a commit out of.

eht16:
Looks good to me, tested and works (on my German keyboard) with and without Numlock turned on.

elextr:
Can anybody with a French keyboard (at least I think its a French keyboard I'm remembering) with numbers on the top and symbols below check if it works, perhaps even though they are on top the numbers will still be level 0?

andy5995:
I just tried this plugin and I have the same problem. I found the open issue.

When I disable numlock the numbered bookmarks work. I have a Logitech K120 USB wired keyboard.

Thank you for working on this @Davidy22 . Does this patch need further testing before it's merged?

elextr:
> Does this patch need further testing before it's merged?

[Yes](https://github.com/geany/geany-plugins/pull/1093#issuecomment-894628779) with keyboards with non-US intl layout, or those could be ignored since nobody with one has bothered to test in two and a half years.

b4n:
I just discovered this, and I wonder why the code isn't "simply" using [`gtk_accelerator_get_default_mod_mask()`](https://docs.gtk.org/gtk3/func.accelerator_get_default_mod_mask.html)? Shouldn't this fix all modifiers issues?  I'll look at this on a computer proper later though.

b4n (commented):
This will accept more modifiers than just <kbd>Control</kbd> and <kbd>Shift</kbd>, it just checks those are active, not that no other are (e.g. <kbd>Meta</kbd>, <kbd>Super</kbd> or others).

I believe the "right" fix for not getting confused by numlock is this:
```diff
diff --git a/geanynumberedbookmarks/src/geanynumberedbookmarks.c b/geanynumberedbookmarks/src/geanynumberedbookmarks.c
index 206c460c..9f8324b9 100644
--- a/geanynumberedbookmarks/src/geanynumberedbookmarks.c
+++ b/geanynumberedbookmarks/src/geanynumberedbookmarks.c
@@ -1482,6 +1482,7 @@ static gboolean Key_Released_CallBack(GtkWidget *widget, GdkEventKey *ev, gpoint
 {
 	GeanyDocument *doc;
 	gint i;
+	GdkModifierType state = keybindings_get_modifiers(ev->state);
 
 	doc=document_get_current();
 	if(doc==NULL)
@@ -1491,7 +1492,7 @@ static gboolean Key_Released_CallBack(GtkWidget *widget, GdkEventKey *ev, gpoint
 		return FALSE;
 
 	/* control and number pressed */
-	if(ev->state==4)
+	if(state == GDK_CONTROL_MASK)
 	{
 		i=((gint)(ev->keyval))-'0';
 		if(i<0 || i>9)
@@ -1501,7 +1502,7 @@ static gboolean Key_Released_CallBack(GtkWidget *widget, GdkEventKey *ev, gpoint
 		return TRUE;
 	}
 	/* control+shift+number */
-	if(ev->state==5) {
+	if(state == (GDK_CONTROL_MASK | GDK_SHIFT_MASK)) {
 		/* could use hardware keycode instead of keyvals but if unable to get keyode then don't
 		 * have logical default to fall back on
 		*/
```

b4n:
Apart from this, this doesn't actually *work* for me with my French keyboard and layout: numbers on the top row are found using <kbd>Shift</kbd>, so it's not actually possible to produce <kbd>Ctrl+"keysym 1"<kbd>, as it'll always include the shift modifier. The equivalent is <kbd>Ctrl+&</kbd>.

One partial improvement would be supporting the keypad.  Here, it's actually the other way around: <kbd>Ctrl+KP_1</kbd> works, but <kbd>Ctrl+Shift+KP_1</kbd> will produce <kbd>Ctrl+KP_End</kbd>.

I see two and a half solutions
* Implement support for the keypad.  With my layout, I could then use the top row to add bookmarks, and the keypad to go to them. Not great, but well.
* Make the keybindings configurable like usual ones.  This might require the users to adjust them for their layout, but that would be flexible and probably fairly easy.
* Do some kaymap magic to figure out the keycodes that correspond to the number keys, and use that instead.

So, the fixes for numlock are good (although see review above), but I don't think that's actually enough.  But as @elextr mentioned above, maybe it's good enough because nobody cared to report it?

techee:
First, sorry for making everyone busy regarding this issue - I just went through various g-p issues and tried to fix the crashers or strange behavior ones and this one seemed to be "fixed for a long time" so I thought we might just merge it.

> I believe the "right" fix for not getting confused by numlock is this:

Agree. I'll try to prepare a PR with this one plus the KP support.

> Apart from this, this doesn't actually work for me with my French keyboard and layout: numbers on the top row are found using Shift, so it's not actually possible to produce...

Wouldn't

https://docs.gtk.org/gdk3/method.Keymap.translate_keyboard_state.html

help here? I remember I suggested we could use it for Geany in https://github.com/geany/geany/pull/1633 and I think this is the kind of situation where it could help. I'll try.

Apart from that, I tested what happens on Windows with the current g-p state and Ctrl+Shoft+num launches some system applications so it may even be possible that this keybinding will be unavailable there.

Davidy22:
Word some amount happened here. So, am I folding in the state changes, or should I just wait for this to get folded into the other change techee is looking at?

techee:
@b4n You are right, `keymap_translate_keyboard_state()` wouldn't work here - I assumed that the `consumed` argument would contain the really consumed modifiers for the key press but after testing it, it contains only "potentially consumed" modifiers so it's useless here.

However, the trick I used is that in addition to `iShiftNumbers` I also precompute `iNoShiftNumbers` (that is the "bottom row" of the key with a number) and based on into which of these two groups the pressed key falls, determine whether Shift was pressed or not.

The result works for me both with the Czech keyboard (numbers in the top row after pressing Shift) and the US keyboard (numbers in the bottom row without Shift). In addition, these `iShift` values aren't computed once inside `plugin_init()` but after every Ctrl+key keystroke so one can change keyboard layout while using Geany and it still works.

See https://github.com/geany/geany-plugins/pull/1458.

> Word some amount happened here. So, am I folding in the state changes, or should I just wait for this to get folded into the other change techee is looking at?

@Davidy22 What I propose in https://github.com/geany/geany-plugins/pull/1458 is something completely different than this PR. The trouble of this PR is that it's not universal enough for all keyboard layouts.

PR #1096, opened 2021-08-07
geanynumberedbookmarks: Fix README typos and add headings

eht16:
Also add license information (based on COPYING).


eht16:
Just noticed the README of the geanynumberedbookmarks plugin had quite some typos and no heading at all. Felt like this should be fixed.

PR #1098, opened 2021-08-13
Geany ctags fixes

### LINKED ISSUE #1042: Geanyctags: Re-generating ctags file fails
Using Geanyctags 1.38 on Ubuntu 20.04.

On first use the 'Project > Generate tags' menu command completes correctly and generates a new .tags file for the project. After that, trying to re-generate the tags (repeat 'Generate tags') fails with the following error in the Messages pane:

```
find -L . -not -path '*/\.*' | ctags --totals --fields=fKsSt --extra=-fq --c-kinds=+p --sort=foldcase --excmd=number -L - -f '/home/user/work/test-project.tags'
ctags: "/home/developer/work/geany-plugins-develop.tags" doesn't look like a tag file; I refuse to overwrite it.
```
Looking inside the .tags file, reveals that the first line of it is blank. Removing the blank line allows the regeneration as expected.

Geany info:
```
geany -V
geany 1.36 (git >= 825acb21) (built on 2020-04-28 with GTK 3.22.30, GLib 2.56.4)
```

techee:
Fixes #1042

### DIFF
--- a/geanyctags/src/geanyctags.c
+++ b/geanyctags/src/geanyctags.c
@@ -183,7 +183,7 @@ static gchar *generate_find_string(GeanyProject *prj)
 {
 	gchar *ret;
 
-	ret = g_strdup("find -L . -not -path '*/\\.*'");
+	ret = g_strdup("find -L . -not -path '*/\\.*' -type f");
 
 	if (!EMPTY(prj->file_patterns))
 	{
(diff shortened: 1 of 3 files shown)


eht16:
LGTM, compiled and tested on Windows (even the change in the `find` command is not relevant there) and still works fine.

techee:
@eht16 Thanks for giving it a try. Yeah, the find changes should do nothing on Windows but the updated readtags library might so good to know it didn't cause any problems.

Issue #1060, opened 2021-02-03
yank paste incorrect behaviour


howdev:
aaa
bbb

when you yank 2 lines from the top and press p at line bbb it should be.
aaa
bbb
aaa
bbb
but instead it is
aaa
bbbaaa
bbb
It should paste from new line but it didn't

I really like to use geany because is fast, but currently vim mode is very broken even with the latest 1.37.1
I don't like to use the mouse/keyboard since I learned to use vim and is so efficient with the simple editing functions.

A lot of the vim mode is missing. for instance, delete line number doesn't work.

elextr:
On behalf of the overly modest creator of the vimode plugin let me remind everyone:

"Vimode is a Vim-mode plugin for Geany written by a guy who does not use Vim. Expect problems unexpected by a Vim user and, please, report them.".

> A lot of the vim mode is missing. for instance, delete line number doesn't work.

Given the note above, probably best if you explained what you expect to happen when, not just list things.  And check known issues [here](https://plugins.geany.org/vimode.html).

And given that neither Geany nor Scintilla is Vim how well some Vim specifics will work is unknown.

howdev:
@elextr  we are aware vim mode is create by a non-vim user. I did report the bugs. I am just reporting there are too many vim broken functions to be usable.

elextr:
@howdev I am making the point that its unhelpful to just say "there are too many vim broken functions to be usable".  Thats just putting down the guy who spent his own time to implement it as far as he has, and who acknowledges it has erorrs and is incomplet. He has made his efforts available for people to use and improve, not for them to criticise its acknowledged shortcomings.  Remember he gets nothing from it as he isn't a Vim user.  

Bug reports should not be accompanied by general negative criticism, its better if you help to improve it by just describing how missing functionality is supposed to work and help to implement it.  The plugins project is open source, anyone can propose a pull request to add missing features.  

Users who know and use Vim are more likely to know how Vim features are expected to work and which ones are important to them, and will actually gain by improving the plugin.

howdev:
why criticised the bug reporter of reporting the state of the plugin? Reporting the state of the plugin is not criticise. Yes the point of Open Source is you expect many to use it, so that they can report the bugs. But if it is too broken how do you use it? Are you then, claiming we are and should use an Alpha part of the software? 
Just a reminder, I did report  what is not working as far as I found. People who report bugs are just as important, they are part of the testing team are they not?

howdev:
[...]

PR #1105, opened 2021-09-02
vimode: add some more ex commands

techee:
These patches add the following ex commands:

```
:y[ank]
:pu[t]
:u[ndo]
:red[o]
:<
:>
:d[elete]
:j[oin]
:co[py]
:m[ove]
```

Fixes the `:delete` command missing part of #1060.

### DIFF
--- a/vimode/README
+++ b/vimode/README
@@ -550,12 +550,22 @@ a new command, please do not forget to update the table below.::
     tag           command         action 
     ------------------------------------------------------------------------------
     :&            :&              repeat last ":substitute"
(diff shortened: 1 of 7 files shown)


techee:
@frlan Any chance of merging the vimode plugin pull requests before the release? It would be nice to have them in the next release so vi users have a better experience with the plugin.

techee:
Thanks Frank! Yeah, I think it's the plugin developer's responsibility to make sure the commits do the right thing (I guess your responsibility as the geany-plugins maintainer is just to make sure it compiles fine as a whole).

Issue #1027, opened 2020-11-13
GeanyLua/GTK3: g_object_get_data: assertion 'G_IS_OBJECT (object)' failed


Skif-off:
Xubuntu 20.04
Geany 1.37.1, GTK 3.24.20, GLib 2.64.3
```
(gdb) run
Starting program: /usr/local/bin/geany
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff548c700 (LWP 80774)]
[New Thread 0x7ffff4be7700 (LWP 80775)]
[New Thread 0x7fffecc9e700 (LWP 80776)]
[New Thread 0x7fffe7fff700 (LWP 80777)]
[Detaching after fork from child process 80778]
[Thread 0x7fffe7fff700 (LWP 80777) exited]

(geany:80770): GLib-GObject-CRITICAL **: 08:43:38.441: g_object_get_data: assertion 'G_IS_OBJECT (object)' failed

Thread 1 "geany" received signal SIGSEGV, Segmentation fault.
0x00007fffedc844e5 in gsdlg_option (label=0x555556413338 "DICT", value=0x555556413308 "dict",
    key=0x555556412e18 "proto", dlg=0x555556403ed0) at gsdlg.c:479
479		cw=g_object_get_data(G_OBJECT(hbox), DataKey);
```
For test use [complex-dialog.lua](https://github.com/geany/geany-plugins/blob/master/geanylua/examples/dialogs/complex-dialog.lua)

Also I tried GTK2-version (Geany 1.36, GTK 2.24.32 and GLib 2.64.3): works fine.

elextr:
The search in gsdlg.c looks fragile, its possible GTK3 changed the widget hierarchy so it doesn't return what was expected.

xiota:
@Skif-off When I use geany/geany-plugins from the Ubuntu 21.04 repository, I get the segfault you describe.  However, when I build geany/geany-plugins from git, I do not.  So the problem is probably already fixed.

Skif-off:
@xiota 
Thanks for checking. I have tried too and I think I will wait v1.38 and close this issue (if Geany team members will not close it).

PR #1112, opened 2021-09-20
GeanyLua: Enable running scripts for all signals

xiota:
Geany can send 20 different signals to plugins, but GeanyLua is only able to run scripts for seven of them.  This PR makes it so that GeanyLua scripts can respond to all signals sent by Geany.  The script file names now also match the signal names so that it is clear exactly what signal they are responding to.

(Note: I wrote this because I needed some signals that GeanyLua doesn't currently handle.)

### DIFF
--- a/geanylua/.clang-format
+++ b/geanylua/.clang-format
@@ -1,0 +1,4 @@
+BasedOnStyle: LLVM
+TabWidth: 3
+IndentWidth: 3
+UseTab: ForContinuationAndIndentation
--- a/geanylua/docs/geanylua-intro.html
+++ b/geanylua/docs/geanylua-intro.html
@@ -22,6 +22,7 @@ the <i style="font-family:sans-serif">Tools->Lua Scripts</i> menu in Geany.
 You can create additional scripts yourself, but all script files must have
 names that end with the <tt><b>.lua</b></tt> file extension. &nbsp;
 The menu labels are generated from the filenames like this:
+</p>
 <ol>
   <li> The &nbsp;<tt><b>.lua</b></tt>&nbsp; filename extension is removed.
   <li> If the filename <i>begins</i> with two decimal digits, followed by a dot,
@@ -59,23 +60,45 @@ user data files, modules, etc.
 <a name="events"><br></a>The special directory  &nbsp;<tt>./plugins/geanylua/events/</tt>&nbsp;
 can contain some scripts that
 will be invoked whenever the respective event is triggered by Geany.
-</p><p>
-Currently, there are nine valid filenames for the event scripts:
+(See <a href="https://www.geany.org/manual/reference/pluginsignals_8c.html">Plugin Signals</a> for more information.)
+</p
+
 <dl compact>
- <dt><tt>init.lua &nbsp; </tt></dt><dd>      -- Runs when the plugin is loaded.</dd>
- <dt><tt>cleanup.lua &nbsp; </tt></dt><dd>      -- Runs when the plugin is unloaded.</dd>
-
- <dt><tt>saved.lua &nbsp; </tt></dt><dd>      -- Runs when an open document is saved.</dd>
- <dt><tt>created.lua  &nbsp; </tt></dt><dd>    -- Runs when a new document is created.</dd>
- <dt><tt>opened.lua  &nbsp; </tt></dt><dd>     -- Runs when an existing file is opened.</dd>
- <dt><tt>activated.lua</tt></dt><dd>  -- Runs when switching documents (notebook tabs).</dd>
-
- <dt><tt>proj-opened.lua &nbsp; </tt></dt><dd>      -- Runs when a project is opened.</dd>
- <dt><tt>proj-saved.lua &nbsp; </tt></dt><dd>      -- Runs when a project is saved.</dd>
- <dt><tt>proj-closed.lua &nbsp; </tt></dt><dd>      -- Runs when a project is closed.</dd>
+ <dt><tt>init.lua</tt></dt>
+ <dd>      -- Runs when the GeanyLua plugin is loaded.</dd>
+ <dt><tt>cleanup.lua</tt></dt>
+ <dd>      -- Runs when the GeanyLua plugin is unloaded.</dd>
 </dl>
+
+<dl compact>
+ <dt><tt>build-start.lua</tt></dt>
+ <dt><tt>editor-notify.lua</tt></dt>
+ <dt><tt>geany-startup-complete.lua</tt></dt>
+ <dt><tt>key-press.lua</tt></dt>
+ <dt><tt>update-editor-menu.lua</tt></dt>
+</dl>
+<dl compact>
+ <dt><tt>document-activate.lua</tt></dt>
+ <dt><tt>document-before-save.lua</tt></dt>
+ <dt><tt>document-close.lua</tt></dt>
+ <dt><tt>document-filetype-set.lua</tt></dt>
+ <dt><tt>document-new.lua</tt></dt>
+ <dt><tt>document-open.lua</tt></dt>
+ <dt><tt>document-reload.lua</tt></dt>
+ <dt><tt>document-save.lua</tt></dt>
+</dl>
+<dl compact>
+ <dt><tt>project-before-close.lua</tt></dt>
+ <dt><tt>project-close.lua</tt></dt>
+ <dt><tt>project-dialog-close.lua</tt></dt>
+ <dt><tt>project-dialog-confirmed.lua</tt></dt>
+ <dt><tt>project-dialog-open.lua</tt></dt>
+ <dt><tt>project-open.lua</tt></dt>
+ <dt><tt>project-save.lua</tt></dt>
+</dl>
+
 <p>
-Note that if any of these nine files exists, it will be executed each time
+Note that if any of these files exists, it will be executed each time
 its corresponding event occurs, which will result in some increase in disk activity.
 In particular, you might notice a delay in startup time with the &nbsp;<tt>opened.lua</tt>&nbsp;
 script when you open a bunch of files at once from the commmand line, etc.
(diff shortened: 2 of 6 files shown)


Skif-off:
Thank you, it looks interesting!

If you have free time and you are familiar with GTK3 API, can you look [this](https://github.com/geany/geany-plugins/issues/1027) bug?

Skif-off:
Excuse my curiosity :) Are ```key_press``` and ```update_editor_menu``` really needed?
As I understood plugin doesn't work with any popup menu and I'm not sure about the first one.

xiota:
> Are key_press and update_editor_menu really needed?

@Skif-off  I don't know.  It's more effort to pick and choose than to just include all of the signals.  No one who doesn't need them needs to use them, and someone might need them someday.

Skif-off:
@xiota

> No one who doesn't need them is forced to use them, and someone else might need them someday.

This is partly true, but let's try to be realists:

```update_editor_menu```: GeanyLua doesn't work with any (his or someone else) popup menu, so it's just not needed and unnecessary checking/processing.

```key_press```: _each_ key/signal will call ```g_file_test``` with ```G_FILE_TEST_IS_REGULAR```: GeanyLua  just rape HDD.
For example, Addons, Autoclose, Vimode or Keyrecord use this signal, but they don't touch file system.

**Comment by elextr** [comment]
I suggest _all_ the signal scripts be checked for existence at plugin start, not checking the file system every signal.

The filesystem will probably be cached (at least on linux) but its still system call overhead.

The at startup means no UI is needed to cause a rescan, just disable and re-enable the plugin.  Or "somebody" could add a Tools menu item to rescan for the convenience of people developing scripts.

**Comment by xiota** [comment]
@Skif-off

* `update_editor_menu` ­– GeanyLua is capable of sending signals to other Geany components.  So it could send a signal to a component that could interact with the menu.  Or a user may wish to take some other action that doesn't require direct interaction with the menu.  Lack of creativity does not make it useless.

* `key_press` – Are you having a *real* problem with the signal, or is this just hypothetical?  On a properly functioning system, checking whether the file exists takes a negligible amount of time.

  On my computer, it takes just under a second on average to check for a nonexistent file a million times (code at the end of thie message).  That's about 0.001 *nano*seconds per keystroke.  That amounts to about 4.38 min of processing for someone typing at 100 wpm for a year *non-stop*.  I've already lost more time just reading your messages than I would have lost to the plugin over *many* years of normal use.

```
#include <time.h>

clock_t begin = clock();

for (int i = 0 ; i < 1000000 ; ++i) {
  char * file_does_not_exist = g_strconcat(geany->app->configdir, "file_does_not_exist", NULL);
  g_file_test(file_does_not_exist, G_FILE_TEST_IS_REGULAR);
}

clock_t end = clock();
double time_spent = (double)(end - begin) / CLOCKS_PER_SEC;

msgwin_status_add("time_spent = %f", time_spent);
```

xiota:
@elextr

> I suggest all the signal scripts be checked for existence at plugin start, not checking the file system every signal.

What problem does this really solve?  Only 0.001 nanoseconds is spent per keystroke checking that the file exists.  That results in 4.38 min for a year of non-stop typing at 100 wpm.  Even pro-writers would be hard pressed to lose more than 30sec per year of normal use to the checks.  Meanwhile, performing the check once at startup eats up programming time, introduces the potential for bugs, and wastes users' time when their scripts don't work and they have to spend time researching the command to update the plugin state.

elextr:
@xiota running a test in a tight loop like that the inode _will_ be cached by Linux, and probably even by windows (although its file operations are notoriously slower than Linux) all you are testing is the cost of the allocate for the string concat and the stat system call to copy existing data, not the cost to access the file system.  This is confirmed when I run the same test on a memory filesystem, an SSD filesystem and a spinny hard disk filesystem and get the same answer.

But when the calls are not in the same closed loop the caching may not happen so well, or even not at all with some mount options.  

And for a remote file on a network (where the default is to ask for new metadata if the requests are more than 3 secs apart) there could be intermittent delays while typing.

Since the cost is incurred only when the plugin is enabled, ultimately the decision is up to the the plugin maintainer, Geany doesn't really care.

xiota:
@elextr 

> ... the inode will be cached by Linux...
> But when the calls are not in the same closed loop the caching may not happen so well...

The inode will also be cached when even the slowest two-finger typists are actively typing.  The access time would have to measure over 100 ms to be noticeable to someone typing over 120wpm.  If that were to happen there is probably something else seriously wrong.

> or even not at all with some mount options.

What mount option?  In this case, the file simply does not exist, so there isn't even an atime to update.

> And for a remote file on a network (where the default is to ask for new metadata if the requests are more than 3 secs apart) there could be intermittent delays while typing.

I suppose someone forced to keep their Lua scripts on a high-latency remote server would have this problem.  In this case it's not enough to just check for file existence at startup.  The scripts themselves should also be cached.  I'm willing to work on this *if* I hear from someone with commit privileges.

> Since the cost is incurred only when the plugin is enabled, ultimately the decision is up to the the plugin maintainer, Geany doesn't really care.

I'm not inclined to make changes to fix hypothetical problems that no one will ever have.  At this time, the estimated max number of users of this patch is 3.  If neither you nor @Skif-off are the plugin maintainer, the chances of this PR being merged is currently nil.

elextr:
> If neither you nor @Skif-off are the plugin maintainer, the chances of this PR being merged is currently nil, regardless of the changes I might make.

Actually Geanylua is orphaned, see the [MAINTAINERS](https://github.com/geany/geany-plugins/blob/master/MAINTAINERS) file, so nobody is responsible for checking any changes to it.  Note its not the job of the collection maintainer to check unlike Geany itself.

But given that @Skif-off has made quite a number of contributions to the plugin, committers like me will probably defer to his advice.

> At this time, the estimated max number of users of this patch is 3. 

Yeah, with no spyware we don't know how many users there are for any Geany features or plugins, maybe we should add it. :smiling_imp:

> The scripts themselves should also be cached. 

Indeed, irrespective of the metadata examination cost, executing a script off disk for each keystroke is potentially a bigger issue.
 
I would have thought a normal design would have scripts executing off signals cached in memory by default, but I don't know how much effort that is for embedded Lua.  I don't expect scripts executing off Geany signals would be very large so memory usage should not be excessive.

Just a note that when we talk about remote filesystems, that includes network appliances and it surprised me that tens of millions of systems per annum were sold when last I looked.  So its likely not as rare as we think.

xiota:
> Yeah, with no spyware we don't know how many users there are for any Geany features or plugins, maybe we should add it. smiling_imp

So this is the path to the dark side...

> Indeed, irrespective of the metadata examination cost, executing a script off disk for each keystroke is potentially a bigger issue.
> 
> I would have thought a normal design would have scripts executing off signals cached in memory by default, but I don't know how much effort that is for embedded Lua. I don't expect scripts executing off Geany signals would be very large so memory usage should not be excessive.

Normally, scripts would be cached by the operating system, at least on Linux.  It's the network device scenario where programs have to do their own caching that would be a problem.

Unfortunately, since I'm starting to think about caching, I might end up implementing it...

> Actually Geanylua is orphaned... so nobody is responsible for checking any changes to it.

There also seem to be a lot of plugins that just won't build anymore because dependencies are unavailable.  Lua 5.1 is well over a decade old, and its final update was about 9 years ago.  So it may not be long before distros stop including it.

I recently saw, but haven't tried, another scripting plugin for Geany ([peasy](https://github.com/kugel-/peasy)).  Since it uses gobject introspection, it presumably has access to all of the API.

elextr:
> There also seem to be a lot of plugins that just won't build anymore because dependencies are unavailable. Lua 5.1 is well over a decade old, and its final update was about 9 years ago. So it may not be long before distros stop including it.

Is there any newer Lua that could plug in?

Unfortunately Geanypy has now expired due to one of its dependencies being stuck on GTK2.  

Because plugins are made by individuals with few additional contributors (mostly) they tend to not have the effort available to keep up with changing dependencies and expire when the dependency does.

> I recently saw, but haven't tried, another scripting plugin for Geany (peasy). Since it uses gobject introspection, it presumably has access to all of the API.

That is made by a regular Geany contributor and (IIUC) makes plugins written in scripting languages "first class" plugins and could replace Geanypy and Geanylua albeit with a different API since both of those attempted to make the API more OO and "native" in the respective scripting language and Peasy is direct translation of the Geany C API.  

Currently its not part of the Geany project itself and (according to Google) its not packaged by anything other than Arch AUR.  I have said I think it should be one of the plugins distributed with Geany itself (like filebrowser, save actions and splitwindow) so its available everywhere Geany is, but its up to the developer if they want to contribute it.

Skif-off:
Yesterday I thought about something like ```get_list_events_scripts``` with multiple calls of ```g_file_test``` that can be call inside ```glspi_init``` and ```glspi_rescan``` only once. Of couse, Geany is not a microcontroller firmware in very limited conditions and does not require _very strict_ optimizations, but checking a boolean value in memory inside GeanyLua looks less wasteful.
But I'm not sure I can do it :) So, I agree to any decision.


@elextr
> has made quite a number of contributions to the plugin

Oh, my contribution is quite modest :)

xiota:
@Skif-off I don't understand what you wrote.  Are you describing a way of turning on/off responding to events from within the init script?  So if someone knows ahead of time that they don't need want to respond to key-press events, they could just turn it off?

xiota:
@elextr @Skif-off

Force push was because of rebase to current master.  I made the following changes:

* Each callback function now keeps track of its own script, rather than storing the filenames in a struct.
* The old event scripts will be used if a new one doesn't exist.  A deprecation warning will be sent to the status window.
* Each callback function disables itself if the associated scripts don't exist.  This reduces overhead to the function call plus a few boolean checks.
* Each callback function keeps a copy of its script in memory.  To clear the cache currently requires restarting Geany.  (Unloading/reloading plugins seems to have some general issues.)

PR #1113, opened 2021-09-29
Update da.po (making it completly untranslated)

scootergrisen:


frlan:
Why this step?

scootergrisen:
Because completely untranslated is better then having 3 translated strings and some incorrect fuzzy strings.

frlan:
Are you planning to restart a new translation?

scootergrisen:
> 
> 
> Are you planning to restart a new translation?

no

frlan:
I had some discussion with ppl I know about this. Even though I'm not a fan of this I'm going to merge this.

PR #1115, opened 2021-10-01
Update Turkish translation

seqizz:


seqizz:
Still not complete but makes it slightly better

PR #1122, opened 2021-10-18
Project Organizer: Set header filetype to match source filetype

### LINKED ISSUE #1121: Project Organizer: Add option/ability to set filetype of header files?
@techee

I opened a similar issue at [geany/#2946](https://github.com/geany/geany/issues/2946).  I was informed that the feature might be a better fit with one of the project-related plugins.  Since the Project Organizer already has some file extension-related settings, setting filetype for ambiguous header files may be within its scope.

The issue is that `.h` files are ambiguous with respect to C and C++.  Identifying the right type most visibly affects syntax highlighting.  But it may also affect whether class structures and other symbols are correctly identified.

1. Many header files have corresponding source files.  When a pair is found, the header type can be changed if needed.

2. Projects may have sufficient information about file locations to find related files in separate directories?

3. If pairing files is too troublesome, or wasteful of cpu cycles, whichever language predominates the project could be used.

4. The project organizer could have a setting that identifies the main project language.  Ambiguous header files could then be set accordingly.

What do you think?  Any other ideas?

xiota:
Resolves #1121.  Resolves geany/geany#2946.

If file is a header and a corresponding source file is found in the same directory, set the header filetype to match the source filetype.  Otherwise, search the project for the corresponding source file.  This is useful for `.h` files, which are ambiguous for C/C++.  Does not address the filetypes of isolated header files.

### DIFF
--- a/projectorganizer/src/prjorg-utils.c
+++ b/projectorganizer/src/prjorg-utils.c
@@ -26,6 +26,7 @@
 
 #include <geanyplugin.h>
 
+#include "prjorg-project.h"
 #include "prjorg-utils.h"
 
 extern GeanyData *geany_data;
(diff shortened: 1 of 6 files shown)


techee:
Thanks. Could you squash the first commit "Project Organizer: Set header filetype to match source filetype" with the third commit "Combine common code between swap header/source..." so this change is easier to review?

I must say I'm not a big fan of the last commit "Set header filetype to match source filetype". Project organizer, as the name suggests, serves for working with projects and isn't meant to do anything when projects aren't open. There's the "Code navigation plugin" for non-project header/source swapping. So I'd prefer this commit to be removed from this pull request.

xiota:
@techee I'll combine commits when I get a chance.

The reasons to make the function work when a project isn't loaded:

* swapping headers isn't inherently project specific.
* If two plugins provide basically the same functionality, one for projects, the other not, they would need multiple keybindings and users would have to keep track of which plugin/keybinding to use when a project is or isn't loaded.

techee:
Hmm, well, yeah, I can probably survive if this patch stays even though I don't like much when a plugin does something that's not obvious that it does...

xiota:
If a general config is added for the open folder and terminal commands, it would be less not obvious that the plugin can do some things that are not strictly project oriented.  I'd expect it would be remain limited to features that benefit from project integration.

xiota:
I combined the commits.  The change that lets it work when a project isn't loaded is:
```C
	if (prj_org)
	{
		header_patterns = get_precompiled_patterns(prj_org->header_patterns);
		source_patterns = get_precompiled_patterns(prj_org->source_patterns);
	}
	else
	{
		// use default header/source patterns if project isn't open
		gchar **headers, **source;
		headers = g_strsplit(PRJORG_PATTERNS_HEADER, " ", -1);
		source = g_strsplit(PRJORG_PATTERNS_SOURCE, " ", -1);

		header_patterns = get_precompiled_patterns(headers);
		source_patterns = get_precompiled_patterns(source);

		g_strfreev(headers);
		g_strfreev(source);
	}
```

If a general config is added to set commands for open folder/terminal, an option to set preferred header/source file extensions could be added as well.  Rather than mess with it in this or the open folder/terminal PRs, I'd prefer to open (yet another) PR specifically to add general preferences.

projectorganizer/src/prjorg-utils.c
```diff
@@ -241,3 +242,193 @@ GtkWidget *menu_item_new(const gchar *icon_name, const gchar *label)
...
+	{
+		/* use default patterns when project isn't open */
+		gchar ** patterns = g_strsplit(PRJORG_PATTERNS_HEADER, " ", -1);
+		header_patterns = get_precompiled_patterns(patterns);
+		g_strfreev(patterns);
+	}
+
+	gchar * doc_basename = g_path_get_basename(doc->file_name);
+	gboolean is_header = patterns_match(header_patterns, doc_basename);
+	gchar * full_name = is_header ? find_header_source(doc) : NULL;
```
techee, line 423: Just a style thing - could you move the declarations of the 3 variables to the beginning of the function and only assign the variables here?

Apart from this change the patch looks good to me.

xiota: I'll make the change.  Do you want the commits squashed again?

techee:
> @techee I made the change you requested. Let me know if you want the commits squashed again.

Yes, please. Apart from this change there's no more work to do I think and this PR will be ready for merging.

xiota:
@techee I assume squashed commits if your preference, so I will also squash the commit I made to another PR.

techee:
> @techee I assume squashed commits if your preference, so I will also squash the commit I made to another PR.

Well, when it's a feature commit, I prefer them separately, but when it's just a fix based on review, it's better not to clutter the geany-plugins git history with these.

Anyway, the pull request looks good and can be merged I think (can take some time based on Frank's availability),

techee (approved):

xiota:
@techee Thank you.

techee:
Let's merge these finally. @xiota Thanks!

PR #1140, opened 2021-11-24
Remove GTK2 jobs from Travis CI

eht16:
While we don't support GTK2 anymore, the Travis CI config still tried to build it but actually built against GTK3 twice because Geany is always built against GTK3.

I'm going to migrate this soon to Github Actions and so this is only a quick'n'dirty attempt to prevent unnecessary builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,21 +5,14 @@ sudo: false
 compiler:
   - gcc
 
-env:
-  matrix:
-    - GTK3=no
-    - GTK3=yes
-
 cache:
   ccache: true
   directories:
     - $HOME/geany/gtk3
-    - $HOME/geany/gtk2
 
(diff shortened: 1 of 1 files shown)


eht16:
Even the build stills fails (because of GeanyLua vs Scintilla), the changes of this PR seem to work and we build against GTK3 only once.

elextr:
Can the m4 for geanylua test for Scintilla < 4 and disable it?  That can then be changed in any PR that fixes it.

eht16:
> Can the m4 for geanylua test for Scintilla < 4 and disable it if fail? That can then be changed in any PR that fixes it.

I guess it is easier to get #1123 ready which is already about to fix the GeanyLua Scintilla problem.

PR #1154, opened 2022-01-12
geanypg: Update to Scintilla 5.1.5 SCI_GETSELTEXT change

techee:
See Scintilla release notes here

https://www.scintilla.org/ScintillaHistory.html

and https://github.com/geany/geany/issues/3095

### DIFF
--- a/geanypg/src/helper_functions.c
+++ b/geanypg/src/helper_functions.c
@@ -133,24 +133,17 @@ void geanypg_load_buffer(gpgme_data_t * buffer)
 {
     /* gpgme_data_new_from_mem(buffer, text, size, 0); */
     GeanyDocument * doc = document_get_current();
-    char * data = NULL;
-    unsigned long size = 0;
+    char * data;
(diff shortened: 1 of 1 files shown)


elextr:
Should not use send message directly, should use `sci_get_selection_contents()`

techee:
> Should not use send message directly, should use sci_get_selection_contents()

Good point, then it doesn't need the API bump change.

But I think we should actually do ABI bump because the previous version won't work correctly, what do you think?

elextr:
> But I think we should actually do ABI bump because the previous version won't work correctly, what do you think?

Well thats part of Geany, and discussions elsewhere, not this PR, and as @kugel- pointed out elsewhere that won't help if the plugin gets re-compiled, but its better than nothing.

kugel-:
LGTM

kugel- (approved):

techee:
I'm merging this as I think the patch is OK - it should be the last piece of the Scintilla 5.1.5 aftermath.

PR #1159, opened 2022-01-26
Migrate from Travis CI to Github Actions

eht16:
Adaption of the previous Travis CI build steps to Github Actions.

Geany is built from current GIT master except it already exists in the build cache, based on the last commit short hash. This is the same behavior as with the Travis CI script.

Additionally, ccache is used to speed up consecutive builds.


elextr:
Question/challenge, can CI be made to build each plugin individually to avoid the current Geany_lua situation where one plugin failing stops any succeeding?

eht16:
Interesting idea. In theory, this could be easily possible with a matrix configration (it's like you start many builds with parameters).

Though the matrix approach could be very time consuming as all build steps are replicated for each plugin. I think such situations as with GeanyLua won't happen that often, hopefully.

For now, I will concentrate on a more 1:1 migration which is already pretty good.
Then Windows builds will follow with a resulting ready to use installer.

elextr:
> all build steps are replicated for each plugin

What I was thinking of was to have the CI setup and build Geany done once then 

```
./configure --disable-all-plugins --enable-xxx; make install
```
where xxx cycles through each plugin.

My ulterior motive was then to read something like `[xxx]` from the PR and only do that one :-)

> For now, I will concentrate on a more 1:1 migration which is already pretty good.

Sure, as you say its rarely been a problem

elextr:
> My ulterior motive was then to read something like [xxx] from the PR and only do that one :-)

Or maybe CI could ask git which directories (== which plugins) are modified and build those.

Or maybe an AI bot could work it out, I here Sweet Geany is unemployed ;-P

eht16:
> > My ulterior motive was then to read something like [xxx] from the PR and only do that one :-)
> 
> Or maybe CI could ask git which directories (== which plugins) are modified and build those.
> 
> Or maybe an AI bot could work it out, I hear Sweet Geany is unemployed ;-P

:)

As said, I would leave it for now and get the current full build in first to finally move away from Travis.

Maybe we could move your suggestion to a seperate issue, maybe someone else wants to work on it.

elextr:
> Maybe we could move your suggestion to a seperate issue, maybe someone else wants to work on it.

Done #1163

eht16:
Great.

@elextr @kugel- any objections?
This is the last repository where Travis-CI is used.

kugel-:
No objections, go for it!

elextr:
Do it, and thanks for the effort.

eht16:
Cool, merged.
I'll keep the Travis CI integration enabled for some time, so existing PRs can still be built without having to rebase to master. At least as long as the remaining credits will not run out again.

PR #1162, opened 2022-01-30
Feature remote debugger

pascal-fb-martin:
This is a small set of changes in the Debugger plugin to support gdb's remote mode (using gdbserver on the remote side). An additional cross development mode launches gdb-multiarch instead of gdb, so that an amd64 workstation can be used to debug a target running on an ARM board, or any other combination supported by gdb-multiarch.

Large portion of the changes is to adjust the debugger Target page, so that the debug mode is explicit and visible. In addition the label flip-flop between Command line Arguments and Remote Address to reflect how the field is actually used.

The `--multi` option of gdbserver is not supported. Doing so might require modifying the GUI, among other impacts.

A French translation is provided for the new labels. Since I am not familiar with the internationalization tools, I just hand-made a few more entries. Feel free to correct if that was not the right way.

### DIFF
--- a/debugger/ChangeLog
+++ b/debugger/ChangeLog
@@ -1,3 +1,7 @@
+01-29-2022 	Pascal Martin <pascal.fb.martin@gmail.com>
+
+	* add remote and cross development debugging mode.
+
 14-06-2012 	Alexander Petukhov <devel@apetukhov.ru>
 
 	* fixed reverse children order in watch/autos
--- a/debugger/ChangeLog
+++ b/debugger/ChangeLog
@@ -1,4 +1,4 @@
-01-29-2022 	Pascal Martin <pascal.fb.martin@gmail.com>
+29-01-2022 	Pascal Martin <pascal.fb.martin@gmail.com>
 
 	* add remote and cross development debugging mode.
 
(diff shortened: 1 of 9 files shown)


elextr:
Although the debugger plugin is marked as maintained, AFAICT there has been no input from the "maintainer" for a long time, so probably in practice it is orphaned.

So anybody (other than @pascal-fb-martin :-) who is interested in this needs to code review and test.

pascal-fb-martin:
As a side note: I am experiencing issues with the gdb's thread handling when using gdb-multiarch. I implemented a quick workaround. I found that I am not the only one with this thread issues, so it most likely not related to geany or the Debugger plugin itself.

Also gdb is not finding OS libraries or header files for the target OS (not finding a thread library might trigger the first issue, actually). I think that might be because I compiled the code under debug native on the Raspberry Pi and the cross compiler is not present on my desktop PC (this is a lot simpler than installing a cross compiler :-).

I use geany on the desktop because the Raspberry Pi has no display, and no GUI environment is installed. Not to mention that a Raspeberry Pi Zero may be slow for geany..

Still the debugger is usable, when cautious (most problems are with step, breakpoints seem OK).

pascal-fb-martin:
These two last commits resolve all the issues that I have experienced so far in remote/cross debugging mode. The Debugger plugin now behaves nicely (at least to _my_ taste).

(The fact that gdb does not retrieve on the desktop machine the sources and libraries that it needs is not the plugin's fault, only an artefact of my weird setup.)

PR #1170, opened 2022-04-03
Debugger: Fix debugging of multi-threaded programs

### LINKED ISSUE #1069: How to debug a multithread application [ Debugger / gdb ]
I am not sure where I should open this issue, here or in geany-plugins.

I need to debug a multithreaded app with Geany with debugger plugin (gdb) but Geany hangs inside the thread.
I have been using Geany for a long time but I can't find a way to debug a thread.

To reproduce the problem, build and debug the sample thread.c, setting a break-point at the line *22* and *34* and then **run**
* Build
        
       gcc -g -O0 -o thread thread.c -lpthread  

* thread.c

```
#include<stdio.h>
#include<string.h>
#include<pthread.h>
#include<stdlib.h>
#include<unistd.h>

/*
 * Build with:
 *
 * gcc -g -O0 -o thread thread.c -lpthread
 *
*/

pthread_t tid[2];

void* doSomeThing(void *arg)
{
    unsigned long x = 0;
    int j = 0;
    int i = (int)arg;

    pthread_t id = pthread_self();

    printf("Thread %d processing...\n", i);
    if(pthread_equal(id,tid[0])) {
        printf("Inside First thread\n");
    } else  {
        printf("Inside Second thread\n");
    }

    for(x=0; x<(0xFFFFFFFF);x++) {
        j++;
    }
    printf("Thread %d: 0x%lx [ x = 0x%X ]\n",i, id, j);

    return NULL;
}

int main(void)
{
    int i = 0;
    int err;

    while (i < 2) {
        err = pthread_create(&(tid[i]), NULL, &doSomeThing, (void*)i);
        if (err != 0)
            printf("can't create thread :[%s]", strerror(err));
        else
            printf("Thread %d created successfully\n", i);
        i++;
    }

    sleep(10);
    return 0;
}

```

Geany correctly stops at the first break-point, but If i click "Step over", Geany switches to the second thread, hitting "Step over" again hangs Geany. And eventually, a crash occurs if you try to close Geany.

The correct behavior should be to stay in the same thread and walk through the code while "Step over" is hit, line by line.

nomadbyte:
- By the original design, the debugger allows thread context switching only
  interactively, while in the Call Stack pane (in Stopped mode). This is
  being handled directly through GtkTreeView::cursor-changed event.
- However, this event was also getting triggered when clearing the latest
  thread's Call Stack after continuing the execution (Running mode).
- This behavior does not seem to have being intended, and it lead to a
  number of issues: stepping hangs, corruption of GDB output processing,
  unnecessary attempts at opening of source files corresponding to thread's
  call-stack frames.
- To avoid all of these issues, GtkTreeView::cursor-changed event is left
  unhandled while clearing the frames. Also the handler for thread-context
  switching is allowed processing only in debugger's Stopped mode to enforce
  the intended behavior.

Fixes #1069

### DIFF
--- a/debugger/src/stree.c
+++ b/debugger/src/stree.c
@@ -57,6 +57,9 @@ static select_frame_cb select_frame = NULL;
 static select_thread_cb select_thread = NULL;
 static move_to_line_cb move_to_line = NULL;
 
+/* signal handler ids */
+static guint cursor_changed_hid = 0;
+
 /* tree view, model and store handles */
 static GtkWidget *tree = NULL;
 static GtkTreeModel *model = NULL;
@@ -353,6 +356,7 @@ static void on_render_address (GtkTreeViewColumn *tree_column, GtkCellRenderer *
 	}
 }
 
+
 /*
  *	inits stack trace tree
  */
@@ -382,7 +386,7 @@ GtkWidget* stree_init(move_to_line_cb ml, select_thread_cb st, select_frame_cb s
 	gtk_tree_view_set_show_expanders(GTK_TREE_VIEW(tree), FALSE);
 	
 	/* connect signals */
-	g_signal_connect(G_OBJECT(tree), "cursor-changed", G_CALLBACK (on_cursor_changed), NULL);
+	cursor_changed_hid = g_signal_connect(G_OBJECT(tree), "cursor-changed", G_CALLBACK (on_cursor_changed), NULL);
 
 	/* for clicking on already selected frame */
 	g_signal_connect(G_OBJECT(tree), "button-press-event", G_CALLBACK(on_msgwin_button_press), NULL);
@@ -464,7 +468,14 @@ void stree_add(GList *frames)
  */
 void stree_clear(void)
 {
+	/* removing nodes from tree triggers the "cursor-changed" event;
+	 * in this context we don't need to process it,
+	 * so the signal handler should be disconnected before the clear
+	 * and reconnected properly afterwards
+	 */
+	g_signal_handler_block(G_OBJECT(tree), cursor_changed_hid);
 	gtk_tree_store_clear(store);
+	g_signal_handler_unblock(G_OBJECT(tree), cursor_changed_hid);
 }
 
 /*
@@ -566,8 +577,17 @@ void stree_remove_frames(void)
 	if (find_thread_iter (active_thread_id, &thread_iter) &&
 	    gtk_tree_model_iter_children(model, &child, &thread_iter))
 	{
+		/* removing nodes from tree triggers the "cursor-changed" event;
+		 * in this context we don't need to process it,
+		 * so the signal handler should be disconnected before the removes
+		 * and reconnected properly afterwards
+		 */
+		g_signal_handler_block(G_OBJECT(tree), cursor_changed_hid);
+
 		while(gtk_tree_store_remove(GTK_TREE_STORE(model), &child))
 			;
+
+		g_signal_handler_unblock(G_OBJECT(tree), cursor_changed_hid);
 	}
 }
 
--- a/debugger/src/debug.c
+++ b/debugger/src/debug.c
@@ -966,6 +966,10 @@ static void on_select_thread(int thread_id)
 {
 	gboolean success;
 
+	/* interactive use - only when in stopped state */
+	if (debug_state != DBS_STOPPED)
+		return;
+
 	if (stack)
 		remove_stack_markers();
 


avafinger (approved):
Compiled and tested with Geany 1.37 (git >= 5cc69b3d)

nomadbyte:
@elextr  Anyone to review this to get it merged? I don't think it's being maintained by anyone current, so I worded in much detail in the commit message to explain both the cause and the fix.

elextr:
@nomadbyte yeah, its over 10 years since the maintainer listed contributed.  

We don't usually summarily remove maintainers, but since there are no contact details in the file to check if they are still interested I think the plugin could be marked orphaned so someone else can take over, if nobody objects (and someone reminds me) in a couple of weeks I'll do that @frlan @eht16 @b4n

debugger/src/stree.c
```diff
@@ -464,7 +480,14 @@ void stree_add(GList *frames)
...
 void stree_clear(void)
 {
+	/* removing nodes from tree triggers the "cursor-changed" event;
+	 * in this context we don't need to process it,
+	 * so the signal handler should be disconnected before the clear
+	 * and reconnected properly afterwards
+	 */
+	disable_cursor_changed_handler();
 	gtk_tree_store_clear(store);
+	enable_cursor_changed_handler();
```
b4n, line 490: You probably should use [`g_signal_handler_block()`](https://docs.gtk.org/gobject/?q=g_signal_handler)-[family of functions](https://docs.gtk.org/gobject/?q=g_signal_handler) instead of rolling your own.

b4n:
> @nomadbyte yeah, its over 10 years since the maintainer listed contributed.
> 
> We don't usually summarily remove maintainers, but since there are no contact details in the file to check if they are still interested I think the plugin could be marked orphaned so someone else can take over, if nobody objects (and someone reminds me) in a couple of weeks I'll do that @frlan @eht16 @b4n

I'd look at who committed recently and ask whether they'd be willing to maintain it, but as far as removing maintainership information I really it's fine 10 years later :)

nomadbyte:
> ... If you are interested in knowing more about the issue, please let me know. 

@avafinger,  you could try to see if the same steps work properly in command-line GDB without Geany.

If the issue is persistent  in Geany, then you may describe it in a new issue against the Debugger plugin.

avafinger:
@nomadbyte 
I will do as you suggested.
There are many signal handlers there that might trigger some weird thing.

By the way, your patch cannot be applied over the last one, i had to manually change the code. It works as expected. ;)

avafinger:
@nomadbyte,
I think i found the reason. The problem is with [Step into] and not [Step over] as i mentioned.
Please, see if you can advise how to overcome the error.
In fact, the debugger closes the debugging section and not Geany.

Below  is the error:

**sudo gdb --args ./minicom -D /dev/ttyUSB1**
```
GNU gdb (Ubuntu 8.1.1-0ubuntu1) 8.1.1
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./minicom...done.
(gdb) b 77
Breakpoint 1 at 0x5365: file minicom.c, line 77.
(gdb) s
The program is not being run.
(gdb) run
Starting program: /apps/minicom-2.7.1/debian/minicom/usr/bin/minicom -D /dev/ttyUSB1
Lockfile is stale. Overriding it..

Breakpoint 1, port_init () at minicom.c:77
77	  m_setparms(portfd, P_BAUDRATE, P_PARITY, P_BITS, P_STOPB,
(gdb) s
78	             P_HASRTS[0] == 'Y', P_HASXON[0] == 'Y');
(gdb) s
77	  m_setparms(portfd, P_BAUDRATE, P_PARITY, P_BITS, P_STOPB,
(gdb) s
m_setparms (fd=3, baudr=0x5555557866e0 <mpars+4032> "1500000", 
    par=0x555555786800 <mpars+4320> "N", bits=0x555555786770 <mpars+4176> "8", 
    stopb=0x555555786890 <mpars+4464> "1", hwf=1, swf=0) at sysdep1.c:396
396	{
(gdb) s
397	  int spd = -1;
(gdb) s
399	  int bit = bits[0];
(gdb) s
408	  if (portfd_is_socket)
(gdb) s
413	  tcgetattr(fd, &tty);
(gdb) s
__GI___tcgetattr (fd=3, termios_p=0x7fffffffde30)
    at ../sysdeps/unix/sysv/linux/tcgetattr.c:34
34	../sysdeps/unix/sysv/linux/tcgetattr.c: No such file or directory.
(gdb) s
38	in ../sysdeps/unix/sysv/linux/tcgetattr.c
(gdb) s
34	in ../sysdeps/unix/sysv/linux/tcgetattr.c
(gdb) s
38	in ../sysdeps/unix/sysv/linux/tcgetattr.c
(gdb) s
40	in ../sysdeps/unix/sysv/linux/tcgetattr.c
(gdb) s
46	in ../sysdeps/unix/sysv/linux/tcgetattr.c
(gdb) s
63	in ../sysdeps/unix/sysv/linux/tcgetattr.c
(gdb) s
42	in ../sysdeps/unix/sysv/linux/tcgetattr.c
(gdb) s
63	in ../sysdeps/unix/sysv/linux/tcgetattr.c
(gdb) s
42	in ../sysdeps/unix/sysv/linux/tcgetattr.c
(gdb) n
Warning:
Cannot insert breakpoint 0.
Cannot access memory at address 0x1ec4635941ad2a3e

__longjmp () at ../sysdeps/x86_64/__longjmp.S:45
45	../sysdeps/x86_64/__longjmp.S: No such file or directory.
(gdb) 
```

elextr:
@avafinger `tcgetattr()` is a linux library call so you can't set a breakpoint in it as its write only.

avafinger:
Yes, but the breakpoint is not in tcgetattr(). The s (Step into) wants the source.
I would expect that gdb would just skip the sysdeps function if not found.
The GDB way is to tell gdb where the sysdeps source code is, but i think must be a better way. 

```
(gdb) s
413	  tcgetattr(fd, &tty);
(gdb) s
__GI___tcgetattr (fd=3, termios_p=0x7fffffffde30)
    at ../sysdeps/unix/sysv/linux/tcgetattr.c:34
34	../sysdeps/unix/sysv/linux/tcgetattr.c: No such file or directory.
(gdb) info source
Current source file is ../sysdeps/unix/sysv/linux/tcgetattr.c
Compilation directory is /build/glibc-CVJwZb/glibc-2.27/termios
Source language is c.
Producer is GNU C11 7.5.0 -mtune=generic -march=x86-64 -g -O2 -O3 -std=gnu11 -fgnu89-inline -fmerge-all-constants -frounding-math -fstack-protector-strong -fPIC -ftls-model=initial-exec -fstack-protector-strong.
Compiled with DWARF 2 debugging format.
Does not include preprocessor macro info.
(gdb) frame 1
#1  0x000055555557556f in m_setparms (fd=3, 
    baudr=0x5555557866e0 <mpars+4032> "1500000", 
    par=0x555555786800 <mpars+4320> "N", bits=0x555555786770 <mpars+4176> "8", 
    stopb=0x555555786890 <mpars+4464> "1", hwf=1, swf=0) at sysdep1.c:413
413	  tcgetattr(fd, &tty);

```

elextr:
The missing source is ok, in [above](https://github.com/geany/geany-plugins/pull/1170#issuecomment-1148048143) you stepped into `tcgetattr()` stepped along a bit which was fine even if no source was shown, then did `next` not `step`.  I would guess `tcgetattr()` does a trap to the kernal, and thats probably not tracable by GDB, which then gets very confused trying to do `next`.  Debugging _should_ work ok if you step over the call to `tcgetattr()` not into it.

avafinger:
You are right, n can step over without throwing an error.
I need to find a way to install the sources in case i inadvertently hit [Step into]. :)

elextr:
I doubt the source will make it work any better, thats only displayed for the human, GDB doesn't use it. :-)

nomadbyte:
@avafinger, If the minicom-bound issue is not directly related to this multi-threaded debugging issue, I'd rather see this topic discussed in a separate issue for clarity and completeness.

nomadbyte:
Looking back, this may also fix the symptoms in #309 (Error 'Can't find a source file')

avafinger:
@nomadbyte and others,

I opened a new issue to discuss the issue [Can't find a source file and no directory or file) there. Please, make your considerations there.
https://github.com/geany/geany-plugins/issues/1179

nomadbyte:
I wonder if this PR is ready for the merge?

elextr:
@nomadbyte I guess its waiting for "somebody" to review and test, IIRC @avafinger applied changes manually, it would be nice for someone to test the PR explicitly.

LGB(very cursory)I

nomadbyte:
> ... IIRC @avafinger applied changes manually, it would be nice for someone to test the PR explicitly.

Sure. However, as you probably could see, the changes are very much limited, so applying manually should be just as effective. Of course, I did the testing of the issue, as I described above.

It's been almost a year since this issue is in DONE state. Is there something that I could do to get this finally merged?

b4n:
I just tried this using #1069 test case, and it's clearly better.  The second example program (more reasonable) seems to work OK.

With the first example it gets confused and state desynchronizes (the plugin doesn't notice the program stopped), leading to the plugin giving weird-looking errors (though they aren't). I got fairly random results though, and believe once I had a hang, but it's a bit too random to tell.  Anyway, I guess the issue here is that the threads are dying and thus just vanish, and something isn't coping well with that.  Admittedly, an interactive GDB session is a bit confusing and random as well, but makes sense.

So there's something to improve here, but it does help quite a bit indeed :+1:

elextr:
@nomadbyte thanks (and thanks for good commit messages)

PR #1183, opened 2022-06-28
Use standard gettext, following Geany core.

kugel-:
Standard gettext should be used over glib's glue these days.
AM_GLIB_GNU_GETTEXT and glib-gettextize are deprecated. This helps
possible meson build (no concrete plans yet) as well as working with
intltool is harder over there. gettext, on the other hand, is supported
out of the box in meson.

Beware, autopoint is a new dependency when building from git,
should not affect tarballs. Also, it's already required by Geany core.

Additionally, autogen.sh follows Geany's autogen.sh more closely,
allowing for out-of-tree builds.

### DIFF
--- a/build/i18n.m4
+++ b/build/i18n.m4
@@ -1,16 +1,14 @@
 AC_DEFUN([GP_I18N],
 [
-    if test -z "$conf_dir" ; then
-        conf_dir="."
-    fi
-    ALL_LINGUAS=`cd "$conf_dir/po" 2>/dev/null && ls *.po 2>/dev/null | $AWK 'BEGIN { FS="."; ORS=" " } { print $[1] }'`
     GETTEXT_PACKAGE=geany-plugins
     AC_SUBST(GETTEXT_PACKAGE)
     AC_DEFINE_UNQUOTED(
-        [GETTEXT_PACKAGE],
+    o    [GETTEXT_PACKAGE],
         ["$GETTEXT_PACKAGE"],
         [The domain to use with gettext])
     LOCALEDIR="${datadir}/locale"
     AC_SUBST(LOCALEDIR)
-    AM_GLIB_GNU_GETTEXT
+
+    AM_GNU_GETTEXT_VERSION([0.19.8])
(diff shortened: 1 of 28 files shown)


build/i18n.m4
```diff
@@ -1,16 +1,14 @@
...
 [
-    if test -z "$conf_dir" ; then
-        conf_dir="."
-    fi
-    ALL_LINGUAS=`cd "$conf_dir/po" 2>/dev/null && ls *.po 2>/dev/null | $AWK 'BEGIN { FS="."; ORS=" " } { print $[1] }'`
     GETTEXT_PACKAGE=geany-plugins
     AC_SUBST(GETTEXT_PACKAGE)
     AC_DEFINE_UNQUOTED(
-        [GETTEXT_PACKAGE],
+    o    [GETTEXT_PACKAGE],
```
eht16, line 6: Is this a typo or some special M4 syntax?

configure.ac
```diff
@@ -16,9 +16,11 @@ AM_PROG_CC_C_O
 AC_DISABLE_STATIC
 AC_PROG_LIBTOOL
 
-dnl i18n
-IT_PROG_INTLTOOL([0.35.0])
-GP_I18N
```
eht16, line 21: If we don't call `GP_I18N` here anymore, couldn't we remove `build/i18n.m4` completely?

If so, we probably need to declare `LOCALEDIR` here or still use `GP_I18N`.

eht16:
Apart from the comments above, LGBI.
I would test it on Linux and Windows when my questions are answered.

build/i18n.m4
kugel-: typo, file will be removed

eht16:
Alright. Roughly tested on Linux and seems to work.

@kugel- could you remove the `build/i18n.m4` file and fix the merge conflicts?

I think then we can merge.

eht16:
@kugel- do you think you will find time to finish this or should anyone pick this up?

eht16:
Superseded by #1434.

PR #1191, opened 2022-08-31
scope: Use 0 instead of NULL for gboolean

kraj:
Fixes warnings with clang 15+

scope/src/stack.c:168:11: error: incompatible pointer to integer conversion initializing 'gboolean' (aka 'int') with an expression of type 'void *' [-Wint-conversion]
        gboolean entry = NULL;
                 ^       ~~~~
1 error generated.

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/scope/src/stack.c
+++ b/scope/src/stack.c
@@ -165,7 +165,7 @@ void on_stack_follow(GArray *nodes)
 gboolean stack_entry(void)
 {
 	GtkTreeIter iter;
-	gboolean entry = NULL;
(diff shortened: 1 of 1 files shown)


kraj:
this CI build failure looks unrelated to this change

eht16:
Why not `FALSE` as default? It's technically the same as `0` but more readable.

kraj:
> Why not `FALSE` as default? It's technically the same as `0` but more readable.

sure. Done

PR #1193, opened 2022-08-31
geanygendoc: Remove all sanity checks testing if tag is tm_tag_file_t

techee:
tm_tag_file_t has never been used by Geany so the check is unnecessary.
In the latest version of Geany this tag was renamed to tm_tag_local_var_t
and used for local variables which breaks compilation.

Those things that appear as big diffs on github are just unindents after the removed `if` check.

### DIFF
--- a/geanygendoc/src/ggd-tag-utils.c
+++ b/geanygendoc/src/ggd-tag-utils.c
@@ -56,16 +56,12 @@ tag_cmp_by_line (gconstpointer a,
   gint          direction = GPOINTER_TO_INT (data);
   gint          rv;
   
-  if (t1->type & tm_tag_file_t || t2->type & tm_tag_file_t) {
-    rv = 0;
+  if (t1->line > t2->line) {
+    rv = +direction;
+  } else if (t1->line < t2->line) {
+    rv = -direction;
   } else {
-    if (t1->line > t2->line) {
-      rv = +direction;
-    } else if (t1->line < t2->line) {
-      rv = -direction;
-    } else {
-      rv = 0;
-    }
+    rv = 0;
   }
   
   return rv;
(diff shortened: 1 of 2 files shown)


elextr:
@techee thanks!!

LGBI

Fixes the CI build

Unless there is communication from @b4n by end of the weekend will commit, it can always be reverted if he wants to do something else.

geanygendoc/src/ggd-tag-utils.c
```diff
@@ -274,7 +267,7 @@ static const struct {
   { tm_tag_externvar_t,       "extern"    },
   { tm_tag_macro_t,           "define"    },
   { tm_tag_macro_with_arg_t,  "macro"     },
-  { tm_tag_file_t,            "file"      }
+  { tm_tag_local_var_t,       "local variable" }
```
techee, line 270: Just noticed that this change depends on the new API with renamed `tm_tag_file_t` to `tm_tag_local_var_t` so we should require that API version (and also bump API version in Geany which we didn't do).

Alternatively, and this might be a better solution, we could just remove this line. If I understand correctly the purpose of this plugin, it serves for generating documentation stubs for symbols at the top level and local variables aren't interesting for it. I checked if something in the plugin doesn't rely on the exact size of this array but this array is only accessed from `ggd_tag_type_get_name()` and `ggd_tag_type_from_name()` below and they use the real size of the array.

elextr: If its not used then there seems no point in leaving it in the array and then the plugin is API agnostic.

techee: `tm_tag_file_t` isn't used but `tm_tag_local_var_t` to which we renamed it is now. But I think `tm_tag_local_var_t` isn't interesting for the plugin so it doesn't have to be listed here (so we don't have to depend on a new Geany API version).

elextr: Agree, I meant "not used by the plugin".

eht16:
Looks good, also your reasoning!

> Those things that appear as big diffs on github are just unindents after the removed `if` check.

Just in case someone didn't know:
on the "Files" tab there is settings icon and when clicking on it, you'll get an option "Hide whitespace". This makes the diff a bit easier to read.

techee:
> Just in case someone didn't know:
on the "Files" tab there is settings icon and when clicking on it, you'll get an option "Hide whitespace". This makes the diff a bit easier to read.

Ahh, that someone happened to be me :-). Thanks for the tip.

PR #1197, opened 2022-09-12
Add --library=gtk to cppcheck

### LINKED ISSUE #1196: cppcheck throws error - warning: There is an unknown macro here somewhere G_DEFINE_TYPE
Recently cppcheck started failing for G-P, e.g.:
```
ao_bookmarklist.c:93:1: warning: There is an unknown macro here somewhere. Configuration is required. If G_DEFINE_TYPE is a macro then please configure it. [unknownMacro]
G_DEFINE_TYPE(AoBookmarkList, ao_bookmark_list, G_TYPE_OBJECT)

overviewprefs.c:86:1: warning: There is an unknown macro here somewhere. Configuration is required. If G_DEFINE_TYPE is a macro then please configure it. [unknownMacro]
G_DEFINE_TYPE (OverviewPrefs, overview_prefs, G_TYPE_OBJECT)
```

I wonder why cppcheck doesn't find the macro anymore.

In the bugtracker and forums I did not find anything relevant:
https://sourceforge.net/p/cppcheck/discussion/general
https://trac.cppcheck.net/

Maybe we need to pass more include locations but then again, why now? What changed?

A full log can be found here (this is with cppcheck 2.9):
https://nightly.geany.org/debian/geany-plugins_1.38-1+20220909git4d08987-1_sid-amd64.log
(the link might break in the future, if so use the latest logfile on https://nightly.geany.org/debian/ for geany-plugins and sid)

I'm pretty sure it is cppcheck, I reproduced locally with cppcheck 2.9 while 2.8 works fine.

Does anyone has a clue?
@kugel- @elextr @techee @b4n

eht16:
Recent versions of cppcheck (2.9+) seem to require this configuration. The gtk.cfg configuration file is part of cppcheck itself.

Closes #1196.

### DIFF
--- a/build/cppcheck.mk
+++ b/build/cppcheck.mk
@@ -4,6 +4,7 @@ if HAVE_CPPCHECK
 check-cppcheck: $(srcdir)
 	$(CPPCHECK) \
 		-q --template gcc --error-exitcode=2 \
+		--library=$(top_srcdir)/build/cppcheck-geany-plugins.cfg \
 		-I$(GEANY_INCLUDEDIR)/geany \
 		$(AM_CPPCHECKFLAGS) $(CPPCHECKFLAGS) \
 		$(srcdir)
--- a/codenav/src/goto_file.c
+++ b/codenav/src/goto_file.c
@@ -151,7 +151,7 @@ directory_check(GtkEntry* entry, GtkEntryCompletion* completion)
     static GtkTreeModel *old_model = NULL;
    	GtkTreeModel* completion_list;
     static gchar *curr_dir = NULL;
-    gchar *new_dir, *new_dir_path; 
+    gchar *new_dir, *new_dir_path = NULL; 
     const gchar *text;
     
     text = gtk_entry_get_text(entry);
@@ -174,8 +174,10 @@ directory_check(GtkEntry* entry, GtkEntryCompletion* completion)
     new_dir = g_strndup (text, dir_sep+1);
     /* I've already inserted new model completion for sub-dir elements? */
     if ( g_strcmp0 (new_dir, curr_dir) == 0 )
+    {
+        g_free(new_dir);
         return;
-
+    }
     if ( curr_dir != NULL )
         g_free(curr_dir);
 
@@ -188,7 +190,7 @@ directory_check(GtkEntry* entry, GtkEntryCompletion* completion)
     log_debug("New completion list!");
 
     if ( g_path_is_absolute(new_dir) )
-        new_dir_path = new_dir;
+        new_dir_path = g_strdup(new_dir);
     else
         new_dir_path = g_build_filename(directory_ref, new_dir, NULL);
 
@@ -196,6 +198,8 @@ directory_check(GtkEntry* entry, GtkEntryCompletion* completion)
     completion_list = build_file_list(new_dir_path, new_dir);
   	gtk_entry_completion_set_model (completion, completion_list);
     g_object_unref(completion_list);
+
+    g_free(new_dir_path);
 }
 
 
--- a/geniuspaste/src/geniuspaste.c
+++ b/geniuspaste/src/geniuspaste.c
@@ -102,6 +102,10 @@ PLUGIN_SET_TRANSLATABLE_INFO(LOCALEDIR, GETTEXT_PACKAGE, PLUGIN_NAME,
                              PLUGIN_VERSION, "Enrico Trotta <enrico.trt@gmail.com>")
 
 
+static void show_msgbox(GtkMessageType type, GtkButtonsType buttons,
+                        const gchar *main_text,
+                        const gchar *secondary_markup, ...) G_GNUC_PRINTF (4, 5);
+
 static void pastebin_free(Pastebin *pastebin)
 {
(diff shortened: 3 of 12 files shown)


elextr:
Needs another free [here](https://github.com/geany/geany-plugins/blob/4d08987a505411a3690447a1b1faea5ee46373c4/debugger/src/dbm_gdb.c#L762)

eht16:
> Needs another free [here](https://github.com/geany/geany-plugins/blob/4d08987a505411a3690447a1b1faea5ee46373c4/debugger/src/dbm_gdb.c#L762)

Interesting. cppcheck 2.9 doesn't find this on my machine. The CI job uses a much older cppcheck version. This is confusing.
Anyway, I'm going to fix it.

elextr:
> cppcheck 2.9 doesn't find this on my machine.
> EDIT: the CI job uses cppcheck 1.82.

Interesting, wonder why your machine doesn't find it (worries what else it misses)

geniuspaste/src/geniuspaste.c
```diff
@@ -102,6 +102,10 @@ PLUGIN_SET_TRANSLATABLE_INFO(LOCALEDIR, GETTEXT_PACKAGE, PLUGIN_NAME,
                              PLUGIN_VERSION, "Enrico Trotta <enrico.trt@gmail.com>")
 
 
+static void show_msgbox(GtkMessageType type, GtkButtonsType buttons,
```
eht16, line 105: This is to trick cppcheck: it complains that it doesn't know `G_GNUC_PRINTF` when it is used in the function definition. But it is ok (probably ignored) if it is used in the function declaration.

This macro is used in other places in G-P in function declarations as well, so I guess it is fine.

geniuspaste/src/geniuspaste.c
```diff
@@ -377,7 +381,10 @@ static gchar *pastebin_get_language(const Pastebin *pastebin,
     gchar *lang = g_key_file_get_string(pastebin->config, PASTEBIN_GROUP_LANGUAGES,
                                         geany_ft_name, NULL);
 
-    return lang ? lang : pastebin_get_default(pastebin, "language", "");
+    if (lang)
```
eht16, line 384: It seems cppcheck doesn't like the tertiary operator in this case.

scope/src/debug.c
```diff
@@ -347,7 +348,7 @@ static void gdb_finalize(void)
 	if (send_channel)
 	{
 		g_io_channel_shutdown(send_channel, FALSE, NULL);
-		g_io_channel_unref(send_channel);
+		g_io_channel_unref(send_channel); // cppcheck-suppress doubleFree // should be fine
```
eht16, line 351: I have no idea if this correct, maybe the unref call is indeed too much?

elextr: AFAICT the unref is needed, the `g_io_channel_shutdown()` documentation, erm ...source ;-) shows it doesn't unref it.

build/cppcheck.mk
```diff
@@ -4,6 +4,7 @@ if HAVE_CPPCHECK
 check-cppcheck: $(srcdir)
 	$(CPPCHECK) \
 		-q --template gcc --error-exitcode=2 \
+		--library=$(top_srcdir)/build/cppcheck-geany-plugins.cfg \
```
eht16, line 7: Is `$top_srcdir` correct here?

eht16:
@elextr @techee @kugel- anyone for review? I guess we all agree that we want a succeeding CI again :).

codenav/src/goto_file.c
```diff
@@ -188,14 +190,16 @@ directory_check(GtkEntry* entry, GtkEntryCompletion* completion)
...
+        new_dir_path = g_strdup(new_dir);
     else
         new_dir_path = g_build_filename(directory_ref, new_dir, NULL);
 
     /* Build the new file list for completion */
     completion_list = build_file_list(new_dir_path, new_dir);
   	gtk_entry_completion_set_model (completion, completion_list);
     g_object_unref(completion_list);
+
+    g_free(new_dir_path);
```
techee, line 202: Shouldn't also `curr_dir` be freed here?

techee:
I just went through the patch and it seems OK to me except the minor comment above. (I'm not the right person to review anything related to autoconf/automake but I'd say if it works, it's alright)

elextr:
LGBcusoryI and CI passes, so I would say merge it (since CI is already broken, it can't make it worse :-) and trigger all PRs to re-CI and see what happens.

codenav/src/goto_file.c
eht16: `curr_dir` is declared static and freed before it is assigned with a new value at line 182. But of course, since it is static within the function, the last instance will never be freed.
Since I don't know the code and don't want to risk to break more than it helps, it would keep it as is.

eht16:
Thanks for the review. I commented on @techee's comment.
If you agree, we could merge it then in a few days.

codenav/src/goto_file.c
techee: Oh yeah, I missed the `static`. It's fine then.

Issue #1213, opened 2022-12-14
Outdated gpgme.m4 is blocking Debian package migration


rdipardo:
According to [Debian's bug tracker]:

> geany-plugins ships [an outdated version of gpgme.m4][0] which cannot handle the transition from gpgme-config to [gpgrt-config].

Apparently the stale macros can be simply pruned out with no impact on the build: https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=1024483;filename=remove-outdated-gpgme.m4.diff;msg=5

But given the proposal to [drop geany-plugins from distribution][1], it seems Debian is expecting a dependency migration to libgpg-error-dev, with updated macros conforming to the gpgrt spec.

There are examples around the Web that might be copied/adapted, e.g.,

- [GPGME](https://dev.gnupg.org/rMf9cbf2c8a839d2c3df20157cfe65b67a4b0bcb5a)
- [libgpg-error (unofficial fork)](https://github.com/ShiftMediaProject/libgpg-error/blob/master/src/gpgrt.m4)

The proper division of package maintenance roles is still mostly a mystery to me. This is just a heads-up, in case anything can be done from Geany's end.

[Debian's bug tracker]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1024483
[gpgrt-config]: https://manpages.debian.org/testing/libgpg-error-dev/gpgrt-config.1.en.html
[0]: https://github.com/geany/geany-plugins/blob/0fd74d766f231e2aa84ca67af4da4773552eceb6/build/bundled/gpgme.m4#L1
[1]: https://tracker.debian.org/pkg/geany-plugins#:~:text=Version%201.38%2Bdfsg%2D1.1%20of%20geany%2Dplugins%20is%20marked%20for%20autoremoval%20from%20testing%20on%20Mon%2019%20Dec%202022.

elextr:
@hyperair as the Debian package maker, and the one who bundled gpgme.m4?

hyperair:
Thanks for the ping, looks like I missed the Debian bug report.

I think I added it years ago in https://github.com/geany/geany-plugins/commit/2535603ae1a211d68d5f2f1122fb5e8972e0b754 to avoid a hard dependency on `gpgme.m4` being present in the system generating the `dist` tarball, because that would result in a dist tarball that is unable to enable `geanypg` via `./configure` without rerunning `./autogen.sh`.

[...]

PR #1214, opened 2022-12-15
Update bundled gpgme.m4

hyperair:
`gpgme.m4` was bundled in geany-plugins 12 years ago in https://github.com/geany/geany-plugins/commit/2535603ae1a211d68d5f2f1122fb5e8972e0b754, and now it's sufficiently outdated that it's causing trouble for the migration from `gpgme-config` to `gpgrt-config`.

This PR updates it to the current version from [gpgme git master](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gpgme.git;a=blob;f=src/gpgme.m4;hb=2ba11f9bfb1f51a481eb1240b8cf2c8ad260c222)

Fixes: #1213


eht16:
Thanks @hyperair!
Tested successfully with the nightly build setup for Windows, Debian Sid and Debian Bullseye.

PR #1217, opened 2023-01-10
Remove unusable plugins: DevHelp, GeanyPy, Multiterm

eht16:
DevHelp depends on unavailable WebKitGtk 1.x.
GeanyPy and MultiTerm are GTK 2 only.

After we agreed on this, I will take care of updating the website, closing related issues and maybe more references.


eht16:
It feels weird to delete plugins from @b4n, @kugel- and @codebrainz :upside_down_face: 

I guess you do not intend to update any of those plugins?

elextr:
Well to be fair, Geanypy has a dependency that has stated it will never be newer than GTK2, so "upgrading" would be a rewrite.

So since it was a rewrite @kugel- replaced Geanypy with Peasy, which is also better, just its not been added to GP.  I think its in some distros (Arch AUR at least), but not Debian or Fedora and their derivatives.

Similarly IIUC Webkit2 (now just webkit) is a rewrite, and not API compatible, but how much work a port would be I don't know, not simple I guess or it would be done :sob:

Probably removing these is the best thing, then anybody who wants/can "upgrade" them can get the source from git tag 1.38 as a reference or starting point and submit a shiny new plugin witjhout having to handle any backwards compatibility.

kugel-:
From my side you can remove geanypy. Regarding peasy, I would rather ship it with Geany itself rather than g-p. It's not a useful plugin on its own, just enables other plugins written in non-C languages (which can be hosted here). Also because it wraps all of Geany's APIs it's much more tied to Geany relases than normal plugins.

elextr:
> Also because it wraps all of Geany's APIs it's much more tied to Geany relases than normal plugins.

Good point.

b4n:
re WebHelper, I kind of plan to port it to something not dead someday (webkit2), but it's been a long while and it didn't happen, mostly as @elextr mentioned because it means some non-trivial rewrites.  Also, I don't actually use that plugin very much (otherwise I probably would have picked this up by then), and mostly wrote it as a feature request if I recall.

If people care, I could possibly try and revamp an old webkit2 port attempt (I think Debian had something back in the days as well), but I wasn't wholly happy about t because it didn't have full feature parity.  And who knows, maybe it'd give somebody else ideas.
Yet if nobody cares, you can also drop it, I won't cry on its demise :shrug:

b4n:
@eht16 well done for tricking me into giving it another shot ;)  Though you only got bare minimum resurrection from old work, but that doesn't actually don't look that bad, not sure why I never submitted it.

b4n:
wrt the others, FWIW I guess that
* devhelp is kind of a lost cause short of bundling libdevhelp
* geanypy as well, and I don't know feature difference, but indeed peasy is probably more current anyway
* multiterm could likely be salvaged with relatively low effort?

elextr:
> multiterm could likely be salvaged with relatively low effort?

Just needs "somebody" to do it ;-)

eht16:
> @eht16 well done for tricking me into giving it another shot ;) Though you only got bare minimum resurrection from old work, but that doesn't actually don't look that bad, not sure why I never submitted it.

This wasn't completely intended by me but nice that it worked :).
Thanks for pushing the old uncompletely work, if anyone wants to pick it up, now they can.

If nobody stops me, I will merge this in a week so that we finally get a green CI again.

b4n:
> Thanks for pushing the old uncompletely work, if anyone wants to pick it up, now they can.
> 
> If nobody stops me, I will merge this in a week so that we finally get a green CI again.

Actually if it's alright with you, I'd rather merge that PR and see how it goes.  I just went over it and fixed a couple of things, and it's actually in a pretty decent shape. And if anybody want to improve what could be missing they could still do.

elextr:
@b4n, "that PR"? Which PR?

kugel-:
Probably #1224

eht16:
@b4n Alright, Webhelper is back in the game.

I would like to restart the "merge in about a week counter" to get this done.

b4n:
@eht16 Fine with me, I now merged the webhelper part, so at least I got the ball moving again (and I think it's in a fair enough state anyway).

BTW I just gave a quick look at multiterm, and it's less trivial than I expected, so it'd require somebody actually motivated to do it.

PR #1231, opened 2023-03-04
GeanyLua: Use luaL_Reg instead of luaL_reg

xiota:
The real name of the struct in Lua 5.1 is luaL_Reg.  Lua 5.1 contains a define for compatibility, but it is not present in LuaJIT or later 5.x releases.

### DIFF
--- a/geanylua/glspi_app.c
+++ b/geanylua/glspi_app.c
@@ -615,7 +615,7 @@ static gint glspi_reloadconf(lua_State* L)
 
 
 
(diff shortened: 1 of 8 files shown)


xiota:
This was a straightforward find/replace to update the name of a struct.  It is a necessary precursor to other updates/fixes of this plugin.  Please merge.  Thank you.

b4n (approved):
Builds fine, and indeed the lua 5.1 header seems to define this as compat only

PR #1233, opened 2023-03-04
GeanyLua: Fix with-lua-pkg configure flag, Add option to use LuaJIT

### LINKED ISSUE #1133: [GeanyLua] For version 5.4.3?
Any change this plugin will be upgraded to Lua 5.4.3 any time soon?

### LINKED ISSUE #1228: [GeanyLua] Switch to LuaJIT?
Lua 5.1 has been deprecated (in a few years, will be two decades).

[LuaJIT](https://github.com/LuaJIT/LuaJIT) appears to be a drop-in replacement for Lua 5.1 that would require only build script or other minor changes.  Based on Git activity, LuaJIT is actively maintained.

xiota:
Modifies `build/geanylua.m4`:

* Fixes `configure --with-lua-pkg=...`.  The flag was previously ignored.
* Falls back on LuaJIT if Lua 5.1 is not available.

When both Lua 5.1 and LuaJIT are installed, LuaJIT can be forced with:
````
./configure --with-lua-pkg=luajit
````

Note: Building with LuaJIT requires #1231.

Resolves #1228.  Possibly also resolves #1133.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,21 @@ Geany Plugins 1.39 (not release date yet)
 
     t.b.p.s.
 
+    GeanyLua
+    * 
+    * 
+    * Add LuaJIT build support (PR #1231, #1233)
+
+      The build default will continue to be Lua 5.1 until the next release,
+      after which the default will switch to LuaJIT.  To force a specific
+      Lua package, configure with the `with-lua-pkg` flag:
+
+          ./configure --with-lua-pkg=luajit
+          ./configure --with-lua-pkg=lua51
+
+    * 
+    * 
+    * 
 
 Geany Plugins 1.38 (2021-10-09)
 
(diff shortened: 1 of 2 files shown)


elextr:
IMO the default should stay lua5.1 for one release to allow packagers to go through the process of changing dependencies.  For example I think Debian requires an approval process to change them.  Also "somebody" needs to integrate it into the Windows bundle.

We can document that it will change in the release after and as soon as the next release happens we can push the change of default.  Of course distros that want to change immediately can just set the CL option to luajit.

xiota:
> ... the default should stay lua5.1...

Done.

> We can document that it will change in the release after...

Is that something I should change/add to this PR?  The `NEWS` file?

elextr:
> Is that something I should change/add to this PR? The NEWS file?

Yeah, its not commonly done, but putting a note in the 1.39 section of the `NEWS` so its remembered is a good idea.

PR #1234, opened 2023-03-04
GeanyLua: Fix geany.activate()

### LINKED ISSUE #1229: [GeanyLua] geany.activate not activating document
`geany.activate()` is supposed activate a document and return true.  If it is unable, it returns false.  However, it is not activating the document.  Accepted arguments are a tab id (negative numbers), document id (positive numbers), or full document path.  Test code attempting to call `geany.activate()` with the document id (test1) and path (test2) follows:
```
  test1 = geany.activate(geany.documents(geany.filename()))

  doc_id = geany.documents(geany.filename())
  test2 = geany.activate(doc_id)

  for key,val in pairs({ test1, test2 }) do
    if val then
      geany.status("true")
    else
      geany.status("false")
      geany.status(geany.documents(geany.filename()))
      geany.status(geany.filename())
    end
  end
```
Results in status window are:
```
false
1
/tmp/test
false
1
/tmp/test
```

For reference, an excerpt from the GeanyLua [documentation](https://plugins.geany.org/geanylua/geanylua-ref.html#activate):

> **`geany.activate ( tab_id )`**
>
> Activates the document specified by tab_id.
>
> If `tab_id` is a number, it activates the document at that index.   A positive number refers to the internal document array, and a negative number refers to the (absolute) GtkNotebook index.  
> Note that the indices begin at (1) and (-1) respectively.
>
> If `tab_id` is a string, the function will try to activate the notebook tab whose filename matches that string.  
> ( The full pathname is required! )
>
> Returns `true` on success, or `false` if the tab could not be found.

xiota:
Resolves #1229.

### DIFF
--- a/geanylua/glspi_doc.c
+++ b/geanylua/glspi_doc.c
@@ -71,15 +71,6 @@ static gint filename_to_doc_idx(const gchar*fn)
 }
 
 
-/* Converts a geany->documents_array index to a notebook tab index */
-static gint doc_idx_to_tab_idx(gint idx)
-{
-	return (
-		(idx>=0) && ((guint)idx<geany->documents_array->len) && documents[idx]->is_valid
-	) ? gtk_notebook_page_num(NOTEBOOK, GTK_WIDGET(documents[idx]->editor->sci)):-1;
-}
-
-
 
 /* Returns the filename of the specified document, or NULL on bad index */
 static const gchar* doc_idx_to_filename(gint idx) {
@@ -92,34 +83,48 @@ static const gchar* doc_idx_to_filename(gint idx) {
 
 
 
-/* Actvate and focus the specified document */
+/* Switch to the specified document tab */
 static gint glspi_activate(lua_State* L)
 {
-	gint idx=-1;
-	if (lua_gettop(L)>0) {
-		if (lua_isnumber(L,1)) {
-			idx=(lua_tonumber(L,1));
-			if (idx<0) { /* Negative number refers to (absolute) GtkNotebook index */
-				idx=(0-idx)-1;
-				if (idx>=gtk_notebook_get_n_pages(NOTEBOOK)) { idx=-1;}
-			} else { /* A positive number refers to the geany->documents_array index */
-				idx=doc_idx_to_tab_idx(idx-1);
-			}
+	gint tab_idx = -1;
 
-		} else {
-			if (lua_isstring(L,1)) {
-				idx=doc_idx_to_tab_idx(filename_to_doc_idx(lua_tostring(L, 1)));
-			} else {
-				if (!lua_isnil(L,1)) { return FAIL_STR_OR_NUM_ARG(1); }
+	if (lua_gettop(L) > 0) {
+		if (lua_isnumber(L, 1)) {
+			gint lua_idx = (lua_tointeger(L, 1));
+
+			/* Note: lua_idx == 0 is undefined */
+			if (lua_idx < 0) {
+				/* Negative number refers to (absolute) GtkNotebook index */
+				tab_idx = (0 - lua_idx) - 1;
+
+				if (tab_idx >= gtk_notebook_get_n_pages(NOTEBOOK)) {
+					tab_idx = -1;
+				}
+			} else if (lua_idx > 0) {
+				/* Positive number refers to the geany->documents_array index */
+				gint doc_idx = lua_idx - 1;
+				if ((doc_idx < geany->documents_array->len) && documents[doc_idx]->is_valid) {
+					tab_idx = document_get_notebook_page(documents[doc_idx]);
+				}
 			}
+		} else if (lua_isstring(L, 1)) {
+			/* String is full document path */
+			gint doc_idx = filename_to_doc_idx(lua_tostring(L, 1));
+			if (doc_idx >= 0) {
+				tab_idx = document_get_notebook_page(documents[doc_idx]);
+			}
+		} else if (!lua_isnil(L, 1)) {
+			return FAIL_STR_OR_NUM_ARG(1);
 		}
 	}
-	if (idx>=0) {
-		if (idx!=gtk_notebook_get_current_page(NOTEBOOK)) {
-			gtk_notebook_set_current_page(NOTEBOOK, idx);
-		}
+
+	/* Switch to document tab if found */
+	if (tab_idx >= 0) {
+		gtk_notebook_set_current_page(NOTEBOOK, tab_idx);
 	}
-	lua_pushboolean(L, (idx>0));
+
+	/* Return True if tab found.  False if not found. */
+	lua_pushboolean(L, (tab_idx >= 0));
 	return 1;
 }
 
--- a/geanylua/docs/geanylua-ref.html
+++ b/geanylua/docs/geanylua-ref.html
@@ -39,7 +39,7 @@ The Geany module provides these functions and variables...<br><br><br>
 <table style="font-family:monospace" summary="function index" width="90%">
 <tr class="odd">
   <td>&nbsp; function <a href="#activate"><b>activate</b></a> ( tab_id )<br></td>
-  <td class="desc">-- Focus the specified editor tab.</td>
+  <td class="desc">-- Activate the specified editor tab.</td>
 </tr>
 <tr class="even">
   <td>&nbsp; function <a href="#appinfo"><b>appinfo</b></a> ()<br></td>


elextr:
LGBI and :+1: for fixing the doc.

Skif-off:
Forgive my curiosity :) Why not
```c
if (tab_idx >= 0) {
	if (tab_idx != gtk_notebook_get_current_page(NOTEBOOK)) {
		gtk_notebook_set_current_page(NOTEBOOK, tab_idx);
	}
}
```
?

xiota:
I haven't looked at this in a while.  Behavior originally intended is to "Actvate and focus" a document.  The original inner conditional `if (idx != ...)` prevented both.  Also, `gtk_notebook_set_current_page` most likely already performs necessary checks, so additional checks wastes processor cycles to produce incorrect behavior.

xiota:
Confirmed original issue described in #1229 and this PR fixes it.  Used `list-open-files.lua` to test.

Force pushed to rebase.  Also corrected `> 0` to `>= 0`.  I had misunderstood @Skif-off's previous comment.

@b4n Should be ready for review and merge.

Skif-off:
Any news?
In fact, the plugin is broken, because one of its functions is not working (maybe this is not the main function, but then what is the main function? :))).

xiota:
> the plugin is broken, because one of its functions is not working

Anything that needs changes to this PR?

> Any news?

I'm waiting for someone to review and commit.  Maybe @frlan or @elextr ?

Skif-off:
@xiota I meant `geany.activate()`, this function is still broken in upstream.

> I'm waiting for someone to review and commit.

Yes, I understand, I just thought that an extra reminder in notifications would remind contributors once again about the unresolved issue. And I hope I wasn't too annoying :))

xiota:
@Skif-off No problem.  Might help to have additional confirmation that this PR is working as expected.  You've tested it after the rebase?

Skif-off:
> You've tested it after the rebase?

Hmm... I thought you tested it. Joke :))
I am not a contributor, so my voice has little weight, but just in case I compiled and tested it 20 minutes ago: works fime, I didn't find any problems.

xiota:
> my voice has little weight ... I didn't find any problems.

Thanks for testing.  Helpful to for you to have checked in case I miss something else.

Skif-off:
@b4n, can you add this to the [2.1.0](https://github.com/geany/geany-plugins/milestone/17) milestone?

I do not know when the new version of Geany is scheduled to be released, there is no alternative solution and it will be sad if all forget about this PR... :)

xiota:
@frlan Any chance you could take a look and merge?  This fixes a bug that has been affecting at least a couple users.  Thanks.

geanylua/glspi_doc.c
```diff
@@ -92,34 +83,51 @@ static const gchar* doc_idx_to_filename(gint idx) {
...
-			if (idx<0) { /* Negative number refers to (absolute) GtkNotebook index */
-				idx=(0-idx)-1;
-				if (idx>=gtk_notebook_get_n_pages(NOTEBOOK)) { idx=-1;}
-			} else { /* A positive number refers to the geany->documents_array index */
-				idx=doc_idx_to_tab_idx(idx-1);
+	gint lua_idx = -1;
+	gint tab_idx = -1;
+
+	if (lua_gettop(L) > 0) {
+		if (lua_isnumber(L, 1)) {
```
b4n, line 93: shouldn't that be [`lua_isinteger()`](https://www.lua.org/manual/5.4/manual.html#lua_isinteger) rather than [`lua_isnumber()`](https://www.lua.org/manual/5.4/manual.html#lua_isnumber) as it's converted using [`lua_tointeger()`](https://www.lua.org/manual/5.4/manual.html#lua_tointeger) in the body?  Note that I don't know the details, but there seem to be more difference than float vs. integer.

geanylua/glspi_doc.c
```diff
@@ -92,34 +83,51 @@ static const gchar* doc_idx_to_filename(gint idx) {
...
-	gint idx=-1;
-	if (lua_gettop(L)>0) {
-		if (lua_isnumber(L,1)) {
-			idx=(lua_tonumber(L,1));
-			if (idx<0) { /* Negative number refers to (absolute) GtkNotebook index */
-				idx=(0-idx)-1;
-				if (idx>=gtk_notebook_get_n_pages(NOTEBOOK)) { idx=-1;}
-			} else { /* A positive number refers to the geany->documents_array index */
-				idx=doc_idx_to_tab_idx(idx-1);
+	gint lua_idx = -1;
```
b4n, line 89: The scope of `lua_idx` could be reduced, and it probably would make things a bit more clear (is that used for anything else, and what is that `-1` about while `-1` is supposed to mean "first tab"?)

geanylua/glspi_doc.c
```diff
@@ -92,34 +83,51 @@ static const gchar* doc_idx_to_filename(gint idx) {
...
+				tab_idx = document_get_notebook_page(documents[doc_idx]);
 			}
-
 		} else {
-			if (lua_isstring(L,1)) {
-				idx=doc_idx_to_tab_idx(filename_to_doc_idx(lua_tostring(L, 1)));
-			} else {
-				if (!lua_isnil(L,1)) { return FAIL_STR_OR_NUM_ARG(1); }
+			if (!lua_isnil(L, 1)) {
+				return FAIL_STR_OR_NUM_ARG(1);
```
b4n, line 119: that could be the last `else if` instead of an `else { if }` while you're at it.  Just nitpicking here though

geanylua/glspi_doc.c
xiota: I wouldn't mind reducing scope, but I've been told changing previously established functionality should be avoided because it's unknown how many people might depend on the old behavior.

Indexing of notebook and documents array don't necessarily match, so might have to split into two separate or create a new lua function to convert between document and notebook indices.

Regarding the calculation...

-1 ⇒ first tab ⇒ 0
0 - (-1) = 1
1-1 = 0

geanylua/glspi_doc.c
xiota: I can change this pending an opinion about change of scope.

geanylua/glspi_doc.c
xiota: I think `lua_idx` is initialized to avoid compiler warnings / errors.  Don't think the particular value matters.

`tab_idx` has to be initialized to a negative number to ensure correct return value.

b4n: By reducing scope I just mean moving the variable declaration inside the `if (lua_isnumber(L, 1)) {}` block where it's used.

As for the initialization, the compiler probably wouldn't complain as the initial value is never used, but I get that some people like initializing everything. I just think that `-1` is a poor choice here as IIUC `-1` means something specific in the code below, while `0` would have been better as it's the only clearly invalid value for that same code.

On the end it wouldn't change anything on how this behaves, it would just make the implementation a hair clearer (IMO).

geanylua/glspi_doc.c
b4n: OK. Did you check how it behaves if called with a string representing a number or a real number rather than an integer?
It's probably harmless but that might change behavior of this function with such values.

geanylua/glspi_doc.c
b4n: > But since it looks like this function never behaved as intended,

It probably worked in the past but for the return value: part of the problematic code was likely "fine" before the internals of Geany's editor tabs was modified a few years ago.

xiota: That type of scope... I'll change it.  I probably just left the variable declaration where it was when I split the variable.  A lot of code in geany seems to declare variables at the beginning of functions.

geanylua/glspi_doc.c
xiota: I tested with following:
```
test = geany.activate("2.6")
if (test)
then
  geany.status("geany.activate success")
else
  geany.status("geany.activate fail")
end
```
It truncates to `2`.

xiota:
@b4n Thank you for reviewing.  I made the suggested changes.

geanylua/glspi_doc.c
b4n: Which Lua version? But anyway using a string here seems horrible anyway so that's just for compatibility

b4n: BTW, unrelated to this, but it seems an awful idea to allow a string representation of a number, and then a string for the file name: this means that if the filename is e.g. `"1.2"` it's not gonna work.
Swapping the integer/string checks would be slightly incompatible but probably a future change worth considering.

xiota: Right now, geanylua supports only 5.1.  I plan to look at that after another bugfix #1397 is merged.

xiota: From comment in code and brief test, string has to be full document path to work.

b4n: Ah, OK, fair enough

b4n:
I squashed the commits, and merged this, hopefully it's been tested well enough :)
Thanks!

PR #1235, opened 2023-03-05
geanylua: Migrate to Lua 5.3

techee:
This PR migrates geanylua to Lua 5.3 (my oldish distro doesn't provide lua 5.4 so I didn't test, maybe it works too). There were mostly just simple changes. From my limited testing, everything seems to work.

Posting as an (in my opinion better) alternative to #1233


techee:
OK, I just tried on another VM and it compiles and runs with Lua 5.4 too, without any modifications. The only time some modifications were needed was for 5.1->5.2 transition, the 5.2->5.3 transition didn't require any change either.

PR #1236, opened 2023-03-06
GeanyLua: Fix geany.keygrab and remove X11 dependency

### LINKED ISSUE #1230: GeanyLua has an X11 dependency
Geany Lua depends on getting X11 events [here](https://github.com/geany/geany-plugins/blob/2665511f731e613944815ab62885ecf52e3037e6/geanylua/glspi_app.c#L469).  This means it won't work on Windows, Macos or Wayland AFAICT.

[Edit: on closer inspection there is a `#else` clause for Windows, so its handled]

Should be using the "official" plugin keybinding interface instead, making it portable.

xiota:
* Make platform agnostic.  Resolves [#1230](https://github.com/geany/geany-plugins/issues/1230).
* Fix [infinite loop](https://github.com/geany/geany-plugins/issues/1230#issuecomment-1455210392).
* Update documentation.

Original function wasn't working on X11.  Rewritten function behaves as described in the original documentation, with the exception of allowing only a single instance of `geany.keygrab` to run at a time.  Subsequent instances return `nil` instead of a key name.  This restriction prevents Geany/GeanyLua from losing keyboard access and hanging.

### DIFF
--- a/geanylua/docs/geanylua-ref.html
+++ b/geanylua/docs/geanylua-ref.html
@@ -638,39 +638,38 @@ This function raises an error if the command argument is invalid.
 
 
 
-<a name="keygrab"></a><hr><h3><tt>geany.keygrab ( [prompt] )</tt></h3><p>
+<a name="keygrab"></a><hr><h3><tt>geany.keygrab ( [prompt] )</tt></h3>
 
-Intercepts the next keyboard key pressed, preventing the editor from receiving
-that keyboard event.</p>
-<p>
-The function returns the value of the pressed key as a string,
-alphanumeric keys will return their keyboard value, such as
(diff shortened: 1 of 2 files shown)


techee:
Thubms up for having a look at this! I didn't have a detailed look at the code but I shamelessly stole the patch and applied during the macOS binary build here

https://github.com/geany/geany-osx/blob/master/geany_patches/08-geany_plugins_geanylua_remove_x11_dependency.patch

and this function seems to work as advertised.

techee:
I just marked this PR with the 2.0 milestone as it's needed on macOS which doesn't have X11. From my point of view this PR can be merged.

elextr:
Since Geanylua is orphaned if it works for you might as well merge it, can't be any worse.

PR #1239, opened 2023-03-07
Update GitHub workflow to Ubuntu 20.04

xiota:
I recently saw this message after creating a pull request.

> This is a scheduled Ubuntu-18.04 brownout. The Ubuntu-18.04 environment is deprecated and will be removed on April 1st, 2023. For more details, see https://github.com/actions/runner-images/issues/6002
>
> Linux Build
> GitHub Actions has encountered an internal error when running your job.
>
> CI Build: .github#L1
> The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002


.github/workflows/build.yml
```diff
@@ -127,7 +127,7 @@ jobs:
             # markdown
             libmarkdown2-dev
             # markdown/webhelper
-            libwebkitgtk-3.0-dev
+            libwebkit2gtk-4.0-dev
```
elextr, line 130: IIUC Webkit2 is not a drop in replacement for Webkit, their APIs diverge, so unless those plugins are updated to webkit2 this is wrong.

xiota: `libwebkit2gtk-4.0-dev` is the dependency being used in the current Ubuntu [`geany-plugins`](http://archive.ubuntu.com/ubuntu/pool/universe/g/geany-plugins/geany-plugins_1.38+dfsg-1build1.debian.tar.xz) package.  `libwebkitgtk-3.0-dev` doesn't exist anymore after 18.04.

xiota:
Getting a distcheck failure related to `vte`.  It may not build anymore on Ubuntu 20.04 because it's not in the repository.

.github/workflows/build.yml
elextr: I was aware that webkit was no longer available, but IIUC the plugins have not yet been updated (see #1217 and #1224)

elextr:
See #1217

.github/workflows/build.yml
xiota: The Markdown plugin does work with `libwebkit2gtk-4.0-dev`.   GeanyPy is gone for other reasons.  I don't know about MultiTerm, I vaguely recall it had other issues.  WebHelper, I understood was using an even older version of webkitgtk than Markdown.

elextr: Ok, since markdown is already updated (but didn't update this) and webhelper is updated by #1224 (and it doesn't update this either) I guess its ok to make this change.  Certainly leaving webkit1 there is useless since its not in the distro.

P.S. I confused webhelper here and devhelper in #1217, ignore references to it other than multiterm

elextr:
Looks like distcheck doesn't respect the `--disable*` options.

elextr:
Thinking about it it makes sense that distcheck doesn't look at configure options, it makes a tarball that is unconfigured (but _is_ `autogen.sh`ed).

xiota:
> Thinking about it it makes sense that distcheck doesn't look at configure options, it makes a tarball that is unconfigured (but _is_ `autogen.sh`ed).

Is `autogen.sh` why multiterm is getting in there when it shouldn't?

elextr:
> Is autogen.sh why multiterm is getting in there when it shouldn't?

It just creates the `configure` shell script IIUC, so tarballers don't need autotools installed.  `configure` needs to be run on the users build system.

elextr:
It looks (being aware of my limitations with autotools and lack of Vala knowledge) that the autotools vala dist tries to pre-compile vala to C and will distribute that so no valac is needed?  Vala knowledgable users answers on a postcard please.

xiota:
When I run distcheck locally (not using Ubuntu though), the problem is vala cannot find vte.  Maybe there is a separate package that vala needs to use vte?

The reason the `--disable-*` flags don't work is distcheck runs configure *after* whatever it's doing that's failing.

I'm going to try disabling multiterm in `Makefile.am`.  The lines are just commented out, so if multiterm is ever fixed, the lines can be easily restored.

xiota:
distcheck is evil:
```
config.status: error: cannot find input file: `multiterm/Makefile.in'
```

elextr:
Sigh!!!

The best thing to do is to update #1217 to not remove webhelper (since #1224 fixes it) and then remove multiterm, and we might get our CI back again.

xiota:
I'll test a more complete removal of multiterm to see what happens.  Can revert if it goes bad or otherwise unwanted.

xiota:
Changed to `actions/checkout@v3` and `actions/cache@v3` because of message:

> Node.js 12 actions are deprecated. ...

There are some more messages.

> The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
>
> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

eht16:
@xiota before you are continuing, please note that most of the what you
are trying to do is already done in:
https://github.com/geany/geany-plugins/pull/1201/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L33

And the PR #1201 is only waiting for #1217 which will remove the
problematic plugins.

Since both PRs already exist and are almost finished, I think you could
stop working on this one.

I'll try to update #1217 soon so that it can be merged and then also
#1201 and we are done.

eht16:
Done in #1201.

PR #1252, opened 2023-05-22
ProjectOrganizer: Initialise "is_header" variable

eht16:
Older cppcheck versions (e.g. 1.90) complain about this while newer versions (e.g. 2.10) do not.

@techee is `FALSE` a proper default here?


elextr:
The original looks safe, `is_header` is set at https://github.com/geany/geany-plugins/blob/6e697f2582b03db6749b91da2f247c6e60b016e2/projectorganizer/src/prjorg-utils.c#L327 except when `known_type` is set `FALSE`.  Then all uses of `is_header` are guarded by `known_type` being true so its safe.

Probably olde cppcheck simply objected to all uninitialised variables, but new cppcheck is smarter and looks at actual usage.

PR #1257, opened 2023-06-04
Addons: Delay updating of tasks list until document has been colorised

### LINKED ISSUE #1254: TODO tasks need the autorefresh button when opening a file to be visible 1.38 kubuntu 22.04, plasma 5.24
Hi,

When opening a file which has tasks, they won't show unless you click the refresh button, I'm not sure if it is default and why?
(1.38 kubuntu 22.04, plasma 5.24)

Kind regards,

eht16:
Closes #1254.

This is not completely perfect but will do the trick to delay the parsing of tasks in the current document until the main loop has triggered the also delayed colourising of the current document which is necessary for parsing the tasks.

### DIFF
--- a/addons/src/ao_tasks.c
+++ b/addons/src/ao_tasks.c
@@ -568,18 +568,22 @@ static void create_task(AoTasks *t, GeanyDocument *doc, gint line, const gchar *
 static gboolean update_tasks_for_doc_idle_cb(gpointer data)
 {
 	AoTasksUpdateTasksForDocArguments *arguments = data;
-	AoTasks *t = arguments->t;
-	GeanyDocument *doc = arguments->doc;
-	gboolean clear = arguments->clear;
+	AoTasksPrivate *priv;
+	GeanyDocument *doc;
 	gint lexer, lines, line, last_pos = 0, style;
 	gchar *line_buf, *display_name, *task_start, *closing_comment = NULL;
 	gchar **token;
-	AoTasksPrivate *priv = AO_TASKS_GET_PRIVATE(t);
+
+	if (! arguments)
+		return FALSE;
+
+	priv = AO_TASKS_GET_PRIVATE(arguments->t);
+	doc = arguments->doc;
 
 	if (doc->is_valid && priv->active && priv->enable_tasks)
 	{
-		if (clear)
-			ao_tasks_remove(t, doc);
+		if (arguments->clear)
+			ao_tasks_remove(arguments->t, doc);
 
 		display_name = document_get_basename_for_display(doc, -1);
 		lexer = sci_get_lexer(doc->editor->sci);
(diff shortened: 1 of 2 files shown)


addons/src/ao_tasks.c
```diff
@@ -556,15 +565,22 @@ static void create_task(AoTasks *t, GeanyDocument *doc, gint line, const gchar *
 }
 
 
-static void update_tasks_for_doc(AoTasks *t, GeanyDocument *doc)
+static gboolean update_tasks_for_doc_idle_cb(gpointer data)
 {
+	AoTasksUpdateTasksForDocArguments *arguments = data;
+	AoTasks *t = arguments->t;
+	GeanyDocument *doc = arguments->doc;
```
elextr, line 572: Needs to check valid since its an unknown period and unknown actions before it happens (isn't asynchronous fun!)

addons/src/ao_tasks.c
```diff
@@ -556,15 +565,22 @@ static void create_task(AoTasks *t, GeanyDocument *doc, gint line, const gchar *
 }
 
 
-static void update_tasks_for_doc(AoTasks *t, GeanyDocument *doc)
+static gboolean update_tasks_for_doc_idle_cb(gpointer data)
 {
+	AoTasksUpdateTasksForDocArguments *arguments = data;
+	AoTasks *t = arguments->t;
```
elextr, line 571: Will the lifetime of AoTasks object guarantee its still valid (see below)

elextr (commented):
Perhaps get the doc pointer first and check valid, then link the tasks object lifetime to the doc (if it isn't already).

Otherwise get current doc in the callback, since tasks doesn't write to the doc it won't do any harm if it scans tasks for the visible doc, not the one initially requested so long as its re-scanned when it becomes visible.

The chances of the doc being deleted between the idle being scheduled and the callback is unlikely, but who knows what them plugins will do.  Better not to use dangling pointers.

addons/src/ao_tasks.c
eht16: I don't know if there is more than checking `data` for `NULL`. BUt at least I've added this.

addons/src/ao_tasks.c
eht16: I think so. The `AoTasks` object exists as long as the Addons plugin is loaded, independently of opened documents and independent of whether the Tasks addon is enabled or not.

While in theory possible I think it's highly unlikely that a user will manage to unload the Addons plugin or even only to quit Geany so that the `AoTasks` object will be destroyed while an idle callback is still waiting to be executed.

eht16:
@elextr fine by you now?

addons/src/ao_tasks.c
elextr: I was thinking of DOCVALID()

eht16:
I'd like to merge this in a few days if there are no objections or further remarks.

elextr:
Oops, missed the changes being added, fine by me.

PR #1259, opened 2023-07-01
vimode: Don't build the viw binary on Windows

### LINKED ISSUE #1255: Compiling geangy-plugins appeared errors
I  use msys2 to compile geany-plugin-1.38 source code ;when I excute "make j2" in msys2 ;it shows:
"collect2.exe:error :ld return 1 exit status " "make[2]: *** [makefile:762:viw.exe] Error 1";
I want to know how to solve it

techee:
With the latest version of msys, the viw binary fails to link (most probably it should be built as a C++ binary instead of the current plain C binary).

Since this binary is used just for testing purposes, there's no need to build it on Windows so this patch skips the build there to avoid this problem.

Fixes #1255 (hopefully)

### DIFF
--- a/vimode/src/Makefile.am
+++ b/vimode/src/Makefile.am
@@ -3,8 +3,6 @@ plugin = vimode
 
 geanyplugins_LTLIBRARIES = vimode.la
 
-noinst_PROGRAMS = viw
-
 vi_srcs = \
 	vi.h \
 	vi.c \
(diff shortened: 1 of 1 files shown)


elextr:
CI passes!!!!!

techee:
> CI passes!!!!!

Was the original problem reproducible with our CI (and the version of msys2 our CI uses)?

elextr:
Ahh, I misremembered #1255 that @eht16 updated CI to latest msys, but on rereading I see now it was only his local build, so yeah need to wait until end of summer holidays so he can check.

eht16:
> end of summer holidays so he can check.

Unfortunately no holidays but I was unable to find enough time anyway :).

> > CI passes!!!!!
> 
> Was the original problem reproducible with our CI (and the version of msys2 our CI uses)?

Let's just celebrate the CI passes, even without any reason! :)

Finally to the topic:
@elextr @techee I can confirm the changes here fix the Windows build. Good job!

eht16:
Thank you @techee.

PR #1242, opened 2023-03-24
Porting devhelp plugin for gtk3. Working display in sidebar and main …

liomarhora:
Porting devhelp plugin for gtk3. Working display in sidebar and main window. Missing showing SearchBar in sidebar or other tweaks.

PR #1260, opened 2023-07-03
Update pt_BR.po

liomarhora:


eht16:
Are these changes intended?

liomarhora:
> Are these changes intended?
 
Yes, I'm removed the old library devhelp2.

eht16:
I don't get it.
Currently, there is no Devhelp in GIT master. So, strictly the *.po files should be updated to remove the previous translations.

But since you are already about to re-add the plugin, updating the translation files would be done best after your changes from #1242 have been merged.

So, why do you want to comment one (to me random) translation string in GIT master from the not yet available Devhelp plugin?

elextr:
Won't it be commented automatically anyway if devhelp is still missing when the message files are re-merged before next release?

IMO leave it to the tools.

liomarhora:
I must have duplicated it due to using two computers
I use 2 PCs to update the plugin.
It seems to me that I created 2 forks for the same plugin.
should I keep both or migrate to the one with the highest number of commits?

eht16:
I suggest to simply close this PR and focus on #1242 which has the same changes.

liomarhora:
> Sugiro simplesmente fechar este PR e focar no [nº 1242](https://github.com/geany/geany-plugins/pull/1242) , que tem as mesmas alterações.
Ok.

liomarhora:
have you tested the plugin in your Geany IDE

PR #1263, opened 2023-07-27
Update POTFILES.in

liomarhora:
removed old library devhelp from archive POTFILEs.in


eht16:
Do we need this PR at all?
It seems the requested changes are also handled in #1242.

liomarhora:
> Do we need this PR at all? It seems the requested changes are also handled in #1242.

I must have applied due to using two computers.

PR #1277, opened 2023-10-11
Spellecheck: Update style mappings

eht16:
Implicitly require Geany 2.0 for G-P as the SpellCheck plugin makes uses of new Scintilla API.

Is this OK?


b4n:
> Implicitly require Geany 2.0 for G-P as the SpellCheck plugin makes uses of new Scintilla API.
> 
> Is this OK?

It is, especially as I just merged #1281 which also uses newer API :smile:

eht16:
Yesterday I noticed that the GeanyLua plugin also uses Scintilla 5.x API and so the version requirement bump is more than justified I think.
Adding conditionals in mutiple plugins only for a hypothetical backwards compability seems not very reasonable.

PR #1280, opened 2023-10-12
git-changebar: Re-set first visible line after Scintilla size request

### LINKED ISSUE #1279: Change popup not working in git-changebar after updating to Scintilla to 5.3.7
To keep a healthy amount of breakages before the release, here's one for git-changebar. After updating Scintilla to 5.3.7, the popup  that shows the diff when a mouse cursor is placed on the marker in the sidebar shows a wrong thing:

Screenshot 2023-10-11 at 20 51 03
Screenshot 2023-10-11 at 20 50 27

It appears to be showing the same value for all diffs in the file.

The commit introducing this issue is

19336d22946b77bbcb5131544ea0e6317cdcf6b5 Update Scintilla to version 5.3.7

The commit before works as expected.

@b4n Any idea what might be wrong?

techee:
After updating to Scintilla 5.3.7 the Scintilla popup with the diff against git shows the wrong number.

It seems that gtk_widget_set_size_request() scrolls Scintilla view so despite it was set previously, it isn't set correctly after this call and has to be re-set.

Note that the call to SCI_SETFIRSTVISIBLELINE has to stay at the original location too, otherwise subsequent calculation using SCI_POINTXFROMPOSITION doesn't work because the line may not be visible.

@b4n Does this look OK to you? Maybe there's a better way to fix this problem so take this only as a hint where the problem is.

Fixes #1279

### DIFF
--- a/git-changebar/src/gcb-plugin.c
+++ b/git-changebar/src/gcb-plugin.c
@@ -878,6 +878,9 @@ get_widget_for_buf_range (GeanyDocument *doc,
                                MIN (width + 2, alloc.width),
                                MIN (height + 1, alloc.height));
   
+  /* Size request seems to scroll Scintilla view so we have to re-set visible lines again */
+  scintilla_send_message (sci, SCI_SETFIRSTVISIBLELINE, line_start, 0);
+  
   return GTK_WIDGET (sci);
 }
 


git-changebar/src/gcb-plugin.c
```diff
@@ -878,6 +878,10 @@ get_widget_for_buf_range (GeanyDocument *doc,
                                MIN (width + 2, alloc.width),
                                MIN (height + 1, alloc.height));
   
+  /* Size request seems to scroll Scintilla view so we have to re-set visible lines again */
+  scintilla_send_message (sci, SCI_SETENDATLASTLINE, 0, 0);
```
b4n, line 882: any reason to re-set `ENDATLASTLINE`?  I don't see it being needed on my end, did you see an issue without it?

b4n (commented):
Thanks a lot for debugging this up to that point!  I somehow didn't notice this until fairly recently, and was puzzled (and frightened) at what it could be, yet didn't have time to dig in.  And now I see you did, yay!

I don't see a problem re-asking for the line; although at some point I'll try and understand what changed in Scintilla that would lead to this.

git-changebar/src/gcb-plugin.c
techee: No, it doesn't seem to be necessary, I just copied the two lines above. Repushed without this line.

b4n (approved):
LGTM.

I don't understand what's actually going on, especially as from my tests it seems to work alright without this for lines <= 182 for me (yes, that feels random). Changes on later lines are stuck starting at line 182…

Anyway, this works and shouldn't cause any more problems.

techee:
> Thanks a lot for debugging this up to that point!

Essential plugin, life would be unbearable without it :-).

>  I somehow didn't notice this until fairly recently, and was puzzled (and frightened) at what it could be, yet didn't have time to dig in.

I was worried it could be some "unsolvable" conflict with the changebar that is now part of Scintilla. But then all the printfs I added seemed to return the right value and then I displayed line numbers in the Scintilla popup and it was clear it wasn't set correctly.

> I don't see a problem re-asking for the line; although at some point I'll try and understand what changed in Scintilla that would lead to this.

Yeah, who knows.

It's actually kind of surprising that it works now because `SCI_POINTXFROMPOSITION` didn't work when I removed `SCI_SETFIRSTVISIBLELINE` in front of the size request and moved it behind - so `SCI_POINTXFROMPOSITION` works only for the displayed editor area but in this case it isn't displayed yet and `sci` has some who-knows-what size before the size request. But I don't know how this could be fixed.

Anyway, good, it works now, the world seems to be saved for a while :-)

b4n:
> It's actually kind of surprising that it works now because `SCI_POINTXFROMPOSITION` didn't work when I removed `SCI_SETFIRSTVISIBLELINE` in front of the size request and moved it behind - so `SCI_POINTXFROMPOSITION` works only for the displayed editor area but in this case it isn't displayed yet and `sci` has some who-knows-what size before the size request. But I don't know how this could be fixed.

Ugh… yeah, maybe someday I'll have to figure that out, because it doesn't seem very safe indeed. :crossed_fingers: 

> Anyway, good, it works now, the world seems to be saved for a while :-)

Thanks! :detective:

PR #1281, opened 2023-10-13
GitChangebar: disable Scintilla's Change History in diff tooltip

eht16:
If the "Change history" feature is enabled in Geany, it might lead to styling issues in the tooltip displaying the diff of the current hunk. So always disable this feature for the tooltip where it is not needed.

I don't think we ever need the "Change history" in the diff tooltip, as the tooltip itself is already the improved version of the change history :).

gcb_200
This happens when the "Change history" feature is enabled in Geany, so it gets also enabled in the Scintilla widget for the tooltip where we don't need it.
For some reason, by setting the margin widths to 0 in https://github.com/geany/geany-plugins/blob/master/git-changebar/src/gcb-plugin.c#L840 the background colors in the Scintilla widget are reset to white (or something else happens with a similar result, I don't understand completely the relation between the margin visibility and the line background color).
Anyway, disable change history prevents this issue.

### DIFF
--- a/git-changebar/src/gcb-plugin.c
+++ b/git-changebar/src/gcb-plugin.c
@@ -837,6 +837,7 @@ get_widget_for_buf_range (GeanyDocument *doc,
   /* hide stuff we don't wanna see */
   scintilla_send_message (sci, SCI_SETHSCROLLBAR, 0, 0);
   scintilla_send_message (sci, SCI_SETVSCROLLBAR, 0, 0);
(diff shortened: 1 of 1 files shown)


b4n:
Thanks! I didn't see it behave so badly as your screenshot, but it also looks weird if the inline option is enabled, so disabling it altogether in the popup seems the way to go -- and no, we don't need it anyway :)

eht16:
FTR: the used theme on the screenshot is Inkpot.

PR #1284, opened 2023-10-17
updatechecker: Fix logic comparing version numbers

techee:
The condition is wrong and for version 2.0 it will constantly report there's an update available when upgrading from 1.x (where x > 0).

I had a "visionary" moment regarding possible 2.0 problems and had a look at the updatechecker plugin if it handles major versions - id does, but wrongly (in addition, there are two identical conditions for the "minor" version but the "mini" version check is missing).

I haven't tested this at all (not sure how to test when there's no actual release).

@b4n @eht16 I think this or some other variant of this patch should go to 2.0.

### DIFF
--- a/updatechecker/src/updatechecker.c
+++ b/updatechecker/src/updatechecker.c
@@ -166,16 +166,22 @@ version_compare(const gchar *current_version)
     parse_version_string(current_version, &geany_current.major,
         &geany_current.minor, &geany_current.mini, &geany_current.extra);
 
-    if ((geany_running.major < geany_current.major) ||
-        (geany_running.minor < geany_current.minor) ||
-        (geany_running.minor < geany_current.minor))
-    {
+    if (geany_running.major < geany_current.major)
         return TRUE;
-    }
-    else
-    {
+    if (geany_running.major > geany_current.major)
         return FALSE;
-    }
+
+    if (geany_running.minor < geany_current.minor)
+        return TRUE;
+    if (geany_running.minor > geany_current.minor)
+        return FALSE;
+
+    if (geany_running.mini < geany_current.mini)
+        return TRUE;
+    if (geany_running.mini > geany_current.mini)
+        return FALSE;
+
+    return FALSE;
 }
 
 


b4n (approved):
LGTM (not tested)

Indeed the check is wrong in many aspects -- event 0.0.2 would be < then 1.38.1
Also, it uses `GEANY_VERSION`, which is a compile-time thing, so better update your update checker!  It could benefit from Geany exposing `main_get_version_string()` or similar (but probably not under that name).

techee:
> Indeed the check is wrong in many aspects -- event 0.0.2 would be < then 1.38.1

Here you'll be saved by the copy-paste error of the current code where the last digit is ignored ;-).

The current code won't behave as bad as I thought originally because for releases the value on the server is the highest number. It would be just in the meantime, like now, where we already have 2.0 and server says 1.38 and for the code 1.38 > 2.0. Anyway, better to fix and also handle the last digit of the version.

> Also, it uses GEANY_VERSION, which is a compile-time thing, so better update your update checker! It could benefit from Geany exposing main_get_version_string() or similar (but probably not under that name).

Yeah, but hopefully most of the time users update geany+plugins at the same time.

b4n:
> Here you'll be saved by the copy-paste error of the current code where the last digit is ignored ;-).

Ah indeed, missed that typo :)

> The current code won't behave as bad as I thought originally because for releases the value on the server is the highest number. It would be just in the meantime, like now, where we already have 2.0 and server says 1.38 and for the code 1.38 > 2.0. 

Well, one issue is that people using the plugin won't be notified about 2.0 anyway, as it'll make it older than basically whatever their version is… but we can't do anything about that.

techee:
> Well, one issue is that people using the plugin won't be notified about 2.0 anyway, as it'll make it older than basically whatever their version is… but we can't do anything about that.

They should get notified because of
```
geany_running.major < geany_current.major
```
The problem was rather the potential "false positive" update notifications of the plugin. (I think the biggest problem of the code was to reason about what actually it would do.)

b4n:
Ah right, it was notifying when *either* `major` or `minor` was higher.  So yeah, only false positives if the local version is newer than the release.

PR #1286, opened 2023-10-18
lineoperations: Fix translation of some strings

b4n:
@smostertdev @frlan This makes translations actually work in several places (prefs dialog, keybindings list, Nth line dialog)


b4n:
@frlan Ideally this could make it to 2.0 as it would extend translation and is fairly trivial, but if it doesn't because it's way too late for this nothing new breaks.

smostertdev (approved):
👍

Issue #1289, opened 2023-11-03
ProjectOrganizer: possible parsing optimization


dolik-rce:
I suspect that project organizer plugin parses some files twice, when opening a project with "Index all project files" set o Yes or Auto.

The reason for this is that the plugin parses all files but then Geany itself parses all opened files (or maybe the other way around?). So if you have a project that has many of its files already opened, it might make the opening of project needlessly slower. I believe that the plugin might safely skip files that are opened in current session (assuming it can get this information)

@techee: If you have no time to look into this, I might try to submit a PR myself. I'd just like to confirm first that this is not intentional and that there is no obvious blocker that would make this impossible.

techee:
Oh and by the way, I'm now playing with LSP support for Geany, see https://github.com/techee/geany-lsp (the API is not yet part of Geany itself so now it's an all-in-one Geany+plugin repo). Since you wanted an improved kotlin support in Geany, you might want to try https://github.com/fwcd/kotlin-language-server together together with the plugin. I haven't tried this particular server so if you run into issues, please tell me.

dolik-rce:
Well, I did a simple test: 
1. I have added  `printf` call to a kotlin parser, that prints first few bytes of the file.
2. I have recompiled geany.
3. Then I created a simple project containing single kotlin file.
4. Run Geany and opened the simple project.
5. The log shows that the parser really runs twice. 

I'll try to figure out why it happens.

I've seen your PR about LSP just today and I have looked at the kotlin server you link. I'm just little worried that it is written by single guy who doesn't even use Kotlin, so he has no real reason to maintain it :slightly_smiling_face: But I might try to look into it, just to test the LSP integration.

techee:
> The log shows that the parser really runs twice.

Then I'd suggest having a look if files are already open when the "project-open" signal runs - if not, it would explain this problem (but I'm not sure what to do about it).

dolik-rce:
Yes, it is exactly as you guessed. The signal is emitted before the files are opened:
```c
gboolean project_load_file_with_session(const gchar *locale_file_name)
{
	if (project_load_file(locale_file_name))  // project-open signal is emitted in this function 
	{
		configuration_open_files(app->project->priv->session_files); // files are opened here
		app->project->priv->session_files = NULL;
		document_new_file_if_non_open();
		ui_focus_current_document();
		return TRUE;
	}
	return FALSE;
}
```

I have found a (little ugly) way to workaround this, see #1290...

[...]

PR #1290, opened 2023-11-03
avoid parsing opened files when loading project #1289

dolik-rce:
Crude but working solution to #1289. 

Disclaimer: I'm not really a C person, so any suggestion on how to improve the code quality are most welcome.

### DIFF
--- a/projectorganizer/src/prjorg-project.c
+++ b/projectorganizer/src/prjorg-project.c
@@ -282,7 +282,9 @@ static void regenerate_tags(PrjOrgRoot *root, gpointer user_data)
 	gpointer key, value;
 	GPtrArray *source_files;
 	GHashTable *file_table;
+	const gchar **session_files;
 
+	session_files = (const gchar **) user_data;
 	source_files = g_ptr_array_new();
 	file_table = g_hash_table_new_full(g_str_hash, g_str_equal, g_free, (GFreeFunc)tm_source_file_free);
 	g_hash_table_iter_init(&iter, root->file_table);
@@ -292,10 +294,11 @@ static void regenerate_tags(PrjOrgRoot *root, gpointer user_data)
 		gchar *utf8_path = key;
 		gchar *locale_path = utils_get_locale_from_utf8(utf8_path);
 		gchar *basename = g_path_get_basename(locale_path);
+		gboolean will_open = session_files && g_strv_contains(session_files, utf8_path);
 
 		if (g_strcmp0(PROJORG_DIR_ENTRY, basename) != 0)
 			sf = tm_source_file_new(locale_path, filetypes_detect(utf8_path)->name);
-		if (sf && !document_find_by_filename(utf8_path)  )
+		if (sf && !will_open && !document_find_by_filename(utf8_path))
 			g_ptr_array_add(source_files, sf);
 
 		g_hash_table_insert(file_table, g_strdup(utf8_path), sf);
@@ -310,7 +313,7 @@ static void regenerate_tags(PrjOrgRoot *root, gpointer user_data)
 }
 
 
-void prjorg_project_rescan(void)
+void rescan_project(gchar **session_files)
 {
 	GSList *elem;
 	gint filenum = 0;
@@ -325,9 +328,13 @@ void prjorg_project_rescan(void)
 		filenum += prjorg_project_rescan_root(elem->data);
 
 	if (prj_org->generate_tag_prefs == PrjOrgTagYes || (prj_org->generate_tag_prefs == PrjOrgTagAuto && filenum < 1000))
-		g_slist_foreach(prj_org->roots, (GFunc)regenerate_tags, NULL);
+		g_slist_foreach(prj_org->roots, (GFunc)regenerate_tags, session_files);
 }
 
+
+void prjorg_project_rescan() {
+    rescan_project(NULL);
+}
 
 static PrjOrgRoot *create_root(const gchar *utf8_base_dir)
 {
@@ -343,6 +350,7 @@ static void update_project(
 	gchar **header_patterns,
 	gchar **ignored_dirs_patterns,
 	gchar **ignored_file_patterns,
+	gchar **session_files,
 	PrjOrgTagPrefs generate_tag_prefs,
 	gboolean show_empty_dirs)
 {
@@ -374,7 +382,7 @@ static void update_project(
 	prj_org->roots = g_slist_prepend(prj_org->roots, create_root(utf8_base_path));
 	g_free(utf8_base_path);
 
-	prjorg_project_rescan();
+	rescan_project(session_files);
 }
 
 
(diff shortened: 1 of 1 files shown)


projectorganizer/src/prjorg-project.c
```diff
@@ -276,13 +276,24 @@ static GeanyFiletype *filetypes_detect(const gchar *utf8_filename)
 }
 
 
+static gboolean string_exists_in_array(gchar **array, gchar *str) {
```
dolik-rce, line 279: I'm pretty sure this function already exists in glib or somewhere... I just couldn't find it :roll_eyes:

projectorganizer/src/prjorg-project.c
```diff
@@ -497,13 +509,51 @@ gchar **prjorg_project_load_expanded_paths(GKeyFile * key_file)
 }
 
 
+static GPtrArray *get_session_files(GKeyFile *config)
```
dolik-rce, line 512: This is mostly ~~stolen from~~ inspired by `configuration_load_session_files` in Geany. The only difference is that this one only reads the filename.

projectorganizer/src/prjorg-project.c
```diff
@@ -497,13 +509,51 @@ gchar **prjorg_project_load_expanded_paths(GKeyFile * key_file)
...
+		g_snprintf(entry, sizeof(entry), "FILE_NAME_%d", i);
+		tmp_array = g_key_file_get_string_list(config, "files", entry, NULL, &error);
+		if (! tmp_array || error)
+		{
+			g_error_free(error);
+			error = NULL;
+			have_session_files = FALSE;
+		} else {
+			unescaped_filename = g_uri_unescape_string(tmp_array[7], NULL);
+			locale_filename = utils_get_locale_from_utf8(unescaped_filename);
```
dolik-rce, line 536: Hm, thinking about it now: This should probably not be here. I think conversion to locale should only be used for display purposes, but all internal filenames should be handled in UTF-8, right?

dolik-rce (commented):
I just hope I didn't introduce any stupid memory leak. It's been years since I had to manage memory manually.

projectorganizer/src/prjorg-project.c
rdipardo: > I'm pretty sure this function already exists in glib or somewhere...

Perhaps you mean [g_strv_contains](https://developer-old.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strv-contains)?

projectorganizer/src/prjorg-project.c
```diff
@@ -275,14 +275,15 @@ static GeanyFiletype *filetypes_detect(const gchar *utf8_filename)
...
 }
 
-
 static void regenerate_tags(PrjOrgRoot *root, gpointer user_data)
 {
 	GHashTableIter iter;
 	gpointer key, value;
 	GPtrArray *source_files;
 	GHashTable *file_table;
+	gchar **open_files;
```
techee, line 284: I'd suggest renaming it to `session_files` because these aren't open yet and may be confusing.

projectorganizer/src/prjorg-project.c
```diff
@@ -292,10 +293,11 @@ static void regenerate_tags(PrjOrgRoot *root, gpointer user_data)
 		gchar *utf8_path = key;
 		gchar *locale_path = utils_get_locale_from_utf8(utf8_path);
 		gchar *basename = g_path_get_basename(locale_path);
+		gboolean is_open = open_files && g_strv_contains(open_files, utf8_path);
```
techee, line 296: Maybe call it `will_open`.

projectorganizer/src/prjorg-project.c
```diff
@@ -310,7 +312,7 @@ static void regenerate_tags(PrjOrgRoot *root, gpointer user_data)
 }
 
 
-void prjorg_project_rescan(void)
+void prjorg_project_rescan(gchar **open_files)
```
techee, line 315: Rename to `session_files`

projectorganizer/src/prjorg-project.c
```diff
@@ -497,13 +500,49 @@ gchar **prjorg_project_load_expanded_paths(GKeyFile * key_file)
...
+
+
 void prjorg_project_open(GKeyFile * key_file)
 {
 	gchar **source_patterns, **header_patterns, **ignored_dirs_patterns, **ignored_file_patterns, **external_dirs, **dir_ptr, *last_name;
 	gint generate_tag_prefs;
 	gboolean show_empty_dirs;
 	GSList *elem = NULL, *ext_list = NULL;
 	gchar *utf8_base_path;
+	GPtrArray *open_files;
```
techee, line 545: `session_files`

projectorganizer/src/prjorg-project.c
```diff
@@ -275,14 +275,15 @@ static GeanyFiletype *filetypes_detect(const gchar *utf8_filename)
 	return ft;
 }
 
-
```
techee, line 278: keep two empty lines before functions

projectorganizer/src/prjorg-project.c
techee: In addition, I'd suggest converting it to a static function called `rescan_project` and having a separate `prjorg_project_rescan()` without any arguments and just calling `rescan_project(NULL);` inside it. This way you avoid the diffs adding the `NULL` parameter across the project.

projectorganizer/src/prjorg-project.c
```diff
@@ -497,13 +500,49 @@ gchar **prjorg_project_load_expanded_paths(GKeyFile * key_file)
...
+		} else {
+			unescaped_filename = g_uri_unescape_string(tmp_array[7], NULL);
+			g_ptr_array_add(files, g_strdup(unescaped_filename));
+			g_free(unescaped_filename);
+		}
+		i++;
+	}
+	g_ptr_array_add(files, NULL);
+
+	return files;
```
techee, line 534: change to `return g_ptr_array_free(files, FALSE);` and return `gchar **` from the function.

projectorganizer/src/prjorg-project.c
techee: In addition, with the suggested change in `get_session_files`, change to `gchar **`.

projectorganizer/src/prjorg-project.c
```diff
@@ -563,6 +604,7 @@ void prjorg_project_open(GKeyFile * key_file)
 	g_strfreev(ignored_dirs_patterns);
 	g_strfreev(ignored_file_patterns);
 	g_strfreev(external_dirs);
+	g_ptr_array_free(open_files, TRUE);
```
techee, line 607: Use `g_strfreev` with the other changes above.

projectorganizer/src/prjorg-project.c
techee: Cast `open_files` to `(const gchar **)` to avoid warnings.

techee (changes requested):
@dolik-rce Thanks! I was hesitating whether this patch doesn't complicate things but it's not too bad so let's do that!

There are some minor comments in the review, nothing major, mostly some renames and minor cleanups.

dolik-rce:
> I was hesitating whether this patch doesn't complicate things but it's not too bad so let's do that!

It is even less complicated now, after I applied your suggestions :-)

techee (approved):

techee:
Thanks! Merged now.

PR #1291, opened 2023-11-04
load files when idle

dolik-rce:
Another attempt at solving #1289, by deferring the parsing to idle time after all files are open. It is far from elegant, but works.


dolik-rce:
Closing after discussion in #1289.

PR #1294, opened 2023-11-21
Fix handling of locals

b4n:
This is a two-parts fix required for usage with 2.0:

8cad88d72c343fd1091b6dab85ecb226c062a9da: Fix a crash when encountering an unknown tag type (e.g. locals)
3af0bf5117a1143b20286f5ee9fb0a2c35916d6b: Add support for locals, and do something useful with them by default.

Without the first one, triggering documentation generation when a local is the "current" symbol crashes the plugin (and thus Geany).
Without the second one in addition, it's annoying because locals "catch" the generation request, although in most cases they should be skipped.

Both of these are particularly important with 2.0 because we now generate locals for purpose of improved scope completion, and thus we have a *lot* of them for parsers supporting it (e.g. the C and C++ one, which is the main default target of the plugin).

### DIFF
--- a/geanygendoc/data/filetypes/c.conf
+++ b/geanygendoc/data/filetypes/c.conf
@@ -41,6 +41,8 @@ doctypes = {
     struct.prototype.policy = FORWARD;
     member.policy           = FORWARD;
     enumval.policy          = FORWARD;
+    # usually, locals are just not documented and get in the way
+    local.policy            = FORWARD;
     
     function = {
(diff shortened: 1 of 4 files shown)


elextr:
LGBI.

b4n:
@frlan note that this might warrant a point release, because basically using this plugin with 2.0 will crash geany (it might not, but it's effectively more likely to crash than not in real use)

Sorry I didn't catch that in time 😕

frlan:
I did not yet plan any – maybe a point-release of Geany would be useful, too – but we should do it.

PR #1295, opened 2023-11-22
Various small WebHelper improvements

b4n:
* Improve documentation
* Add support for webkit2gtk-4.1 (no changes)
* Add a key binding & menu item for loading the current file into the web view (for e.g. static HTML pages)


elextr:
Documentation looks ok, rest went zooming over my head.

jbicha:
Switching from webkit2gtk 4.0 to 4.1 is important. Fedora has [stopped](https://fedoraproject.org/wiki/Changes/Remove_webkit2gtk-4.0_API_Version) building the 4.0 API in preparation for their Fedora 40 release in a few months.

Debian and Ubuntu are preparing to do the same thing.

Ideally, all the plugins would be ported at the same time.

- markdown (uses webkitgtk 4.0
- webhelper (uses webkitgtk 4.0
- geniuspaste (uses libsoup2)
- updatechecker (uses libsoup2)

However, it is not possible for a process to load both libsoup2 and libsoup3 at the same time. I think this could be possible if an app uses plugins. I tested this with the markdown plugin (which I switched to use webkitgtk 4.1). With the markdown plugin enabled and a Markdown document open, many times (but not every time), I was unable to use the _Check for Updates_ button or _Paste It!_ without Geany hanging.

elextr:
Maintainers are:

- Webhelper - @b4n, this PR
- Markdown - in theory @codebrainz, but hasn't contributed to it for 8 years, the markdown plugin died for a while until others ported to GTK3, maybe @b4n could port the (small) webkit4.1 change to markdown, or it can die again until a new maintainer or contributor steps up
- Updatechecker - @frlan 
- Geniuspaste - in theory @Enrix835, but hasn't contributed to it for 12 years, maybe if the libsoup3 changes are small enough @frlan may port them to Geniuspaste, or wait for a maintainer or contributor.

xiota:
I don't use this plugin, but can test it to help facilitate its update if provided a list of things to check.

elextr:
Which plugin did you mean? There are four listed in the previous two comments.  

Since I would think that it is not possible to load old and new versions of webkit2gtk and libsoup at the same time probably the version needs to be selected once and builds of all four need to disable them if the new version is found.  Then they can be switched to re-enable as they are modified to use the new libraries, ping @b4n as the plugin build system expert.

Webhelper is done here by itself and if the comments are correct will select webkit2gtk-4.1 if its available, so @xiota if you build it with this PR applied it should build with the new version if its on your system.  Then just try it to see that it doesn't crash when you use it a bit.

Then the other plugins need to be adapted to the new libraries by "somebody", since the maintainers seem to have departed.

b4n:
@jbicha could you make a PR with your Markdown changes, assuming they are not too Ubunut/Debian-specific?
@elextr @jbicha I'll merge this (which at least *allows* building webhelper with 4.1, although not forcing it), and try to update the rest to match, as indeed it will cause issues if various plugins load various versions of those libraries.

b4n:
@xiota if you'd like to try, just look at the PR's commit messages.  Basically, try building against WebKit2GTK 4.1, enable the plugin and try loading pages, editing an HTML displayed there, etc.  But all should be good in theory :crossed_fingers:

xiota:
* Loads and seems to function when built with either webkit2gtk 4.0 or 4.1.
* Key bindings all work when conditions for their function are met.
* The reload on save function works.

elextr:
Maybe should update the requirements section of README as well.

b4n:
@elextr I just did the bare minimum, see update

webhelper/README
```diff
@@ -26,9 +26,9 @@ Prominent features
 Requirements
 ============
 
-This plugin requires GTK+ (>= 2.16), GLib (>= 2.16), GIO (>= 2.18),
-GdkPixbuf (>= 2.0), WebKitGTK (>= 1.1.18), and obviously Geany (>= 0.20) to
-work. If you intend to build it yourself, you will need to get the development
+This plugin requires GTK3, GLib (>= 2.22), GIO (>= 2.18), GdkPixbuf (>= 2.0),
+WebKit2GTK (either API 4.0 or 4.1), and obviously Geany (>= 0.20) to work.
```
elextr, line 30: Geany 0.20 or 2.0?

b4n: Who knows?  As mentioned, I didn't properly review those numbers just yet.  Geany dep is likely to be closer to 0.20 than 2.0 though.  Others like GTK3 or GLib might be higher.
But I don't plan in fiddling with that for the moment, because it's less trivial than what I feel it gains, and can be done at a later point.

elextr: I thought it might have been a typo 0.2 vs 2.0 ;-) 

Don't forget that the plugin READMEs go to the plugins.geany.org site, probably doesn't have to be 2.0 but pretending it will work with any Geany pre-GTK3 is wrong.

b4n: > but pretending it will work with any Geany pre-GTK3 is wrong.

Good point.  Well, what about you opening a "update plugins requirements" PR after having checked and updated them all? :grin: 
Though, if I find the time (maybe tonight? who knows) I'll try and do that for this one.  But I don't love seeing this endlessly delayed for basically unrelated stuff… although I laid that on myself given the PR title, didn't I?

elextr: Well, you updated requirements in README as part of #1336, which is what prompted me to mention it here and on the markdown PR, so its all your own fault :grin: 

And you had already made changes to README here so it seemed pretty simple to add.

b4n: Damned :smile: 

Simple to add, yes, but simple to figure out, not so much.
I sometimes wish there was a generic tool that can *reliably* scan the code an tell me "ok, you need libfoo >= 1.2.3 (because of $reason), libbar 42.0 (because of $othereason), …", but well.  I had a script that did something like that ad-hoc grepping GTK documentation and the code to try and get an idea, but not sure if it still works, and it never was truly reliable.

xiota: What was the first version of Geany supporting GTK 2.x?  Is it necessary to mention Geany version?

elextr: Well, do we need to list all those things individually?  I would expect a new enough webkit2gtk will require newer versions of Glib, GTK3, etc than the plugin.  So wherever a builder gets their webkit2gtk-dev from should have the right parts automatically ... won't it?

b4n: Probably

b4n:
And voila, I got plagued with another round of cleanups :)

webhelper/src/gwh-browser.c
```diff
@@ -950,6 +1005,23 @@ gwh_browser_set_uri (GwhBrowser  *self,
...
+gwh_browser_set_uri_from_document (GwhBrowser    *self,
+                                   GeanyDocument *doc)
+{
+  gchar *uri;
+
+  /* document must exist on disk */
+  if (! doc || ! doc->real_path)
+    return FALSE;
+
+  uri = g_strconcat ("file://", doc->file_name, NULL);
```
elextr, line 1018: Why test doc->real_path above and use untested doc->file_name here?  Shouldn't this use the path?

elextr (approved):
LGBI except for one query.

webhelper/src/gwh-browser.c
b4n: No, `doc->real_path` is used to verify the document is actually saved to a file, but it's in locale encoding, whereas `doc->file_name` is the expected UTF-8.  Admittedly, it might not behave properly if the locale is non-UTF-8 and the file path contains non-ASCII characters; but at least the worse that would happen would be the file not being found.

elextr: Ok, I forgot the inconsistency that sometimes paths are called paths and sometimes filenames in C land, ... sigh @elextr goes back to his nice consistent C++ `std::path` :grin:

PR #1298, opened 2023-12-09
projectorganizer: prjorg-sidebar.c: Cast g_ptr_array_free to gchar**

### LINKED ISSUE #1297: projectorganizer: prjorg-sidebar.c: returning 'void **' from a function with incompatible return type 'gchar **'
When compiling on Linux with GCC 13 and `-Wincompatible-pointer-types`:

```
prjorg-sidebar.c: In function 'prjorg_sidebar_get_expanded_paths':
prjorg-sidebar.c:1565:16: warning: returning 'void **' from a function with incompatible return type 'gchar **' {aka 'char **'} [-Wincompatible-pointer-types[https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wincompatible-pointer-types]]
 1565 |         return g_ptr_array_free(expanded_paths, FALSE);
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Apparently this will become an error by default with GCC 14 in 2024:

https://gcc.gnu.org/pipermail/gcc-cvs/2023-December/394351.html

https://github.com/gcc-mirror/gcc/commit/9715c545d33b3a32ddc1ae817ba9356ade1fb9df

csfore:
Grepping the source code shows this is already done in pohelper/src/gph-plugin.c at line 903.

Tests run: Compiles on GCC 14

Fixes #1297

Bug report from Gentoo: https://bugs.gentoo.org/919446

### DIFF
--- a/projectorganizer/src/prjorg-sidebar.c
+++ b/projectorganizer/src/prjorg-sidebar.c
@@ -1562,7 +1562,7 @@ gchar **prjorg_sidebar_get_expanded_paths(void)
 		(GtkTreeViewMappingFunc)on_map_expanded, expanded_paths);
 	g_ptr_array_add(expanded_paths, NULL);
 
-	return g_ptr_array_free(expanded_paths, FALSE);
+	return (gchar **) g_ptr_array_free(expanded_paths, FALSE);
 }
 
 


elextr:
Sigh C!!! [exasperated tone]  Surely it can trace the type of the pointers that were put in the array as `gchar*` where they were implicitly converted to `void*` and so allow the therefore legal implicit conversion from `void*` to `gchar*` in the return :stuck_out_tongue_winking_eye: 

Anyway LGBI since I don't have gcc 14.  @b4n will probably want to add this in his --b4n_pedantic :grin:

csfore:
Heh they're slowly making GCC more strict it seems :) in 14 they're also no longer including `<algorithm>` by default so it's fun having to add a single line import to projects

elextr:
The `<algorithm>` thing (which is C++ not C) is within the C++ standard that allows any header to include any other header (or not, which is what they changed, some headers that used to use it no longer do).  Users should always include what they use irrespective of "automagic" inclusions in headers since they can change at any time.  There is a [tool](https://github.com/include-what-you-use/include-what-you-use) released by Google that leeches off clang to check for it.

techee:
Ah, OK, I noticed this PR only after merging https://github.com/geany/geany-plugins/pull/1315 which eliminates some more warnings including this one.

So I'm closing this and https://github.com/geany/geany-plugins/issues/1297.

PR #1309, opened 2024-02-13
Add missing sign for cppcheck --template argument

eht16:
Closes #1304.

PR #1310, opened 2024-02-17
CI: Download, compile and use latest cppcheck version

eht16:
Instead of using a dated version of "cppcheck" from the runner image, we download the latest release of "cppcheck" and build it.

As building it takes a few minutes, the build result is cached to be re-used in succeeding build runs.

This should make us aware of failed checks in recent "cppcheck" versions quicker. Currently we notice these mostly only when running "make check" locally or when the nightly builds for Debian Unstable fail because they also use recent "cppcheck" versions.

Currently the build fails because of the pending changes in #1309, once this is merged, the build should succeed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,6 +36,7 @@ jobs:
     env:
       CC:                       ccache gcc
       CXX:                      ccache g++
+      CPPCHECK_CACHE_PATH:      ${{ github.workspace }}/.cppcheck_cache
       GEANY_SOURCE_PATH:        ${{ github.workspace }}/.geany_source
       GEANY_CACHE_PATH:         ${{ github.workspace }}/.geany_cache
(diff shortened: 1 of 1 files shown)


andy5995:
Have you ever considered adding a [CodeQL check](https://github.com/github/codeql-action) here or on Geany itself?

elextr:
> Have you ever considered adding a [CodeQL check](https://github.com/github/codeql-action) here or on Geany itself?

Not relevant to this PR, if "somebody" ran it on the source and posted the results in a separate issue it might be interesting, but please do not pollute PRs with other topics.

eht16:
If there are no objections or improvement suggestions, I would like to merge to this in a week.

frlan:
If it's working -- why not. I have absolutely no experience with this part of the universe and therefor no objective opinion.

PR #1311, opened 2024-02-23
Devhelpplugin

liomarhora:
Insert devhelp plugin


elextr:
It still contains build artifacts that should not be saved in the repository.

liomarhora:
> artefatos de construção que não devem ser salvos no repositório

deletes the old build that appeared in the workflows.
I generated a more current one.
does this solve it?

liomarhora:
> It still contains build artifacts that should not be saved in the repository.

Hello elextr.
Thank you for the informations!
I managed to make the necessary adjustments so that the devehelp plugin works in the new version of Geany.
Could you verify?

liomarhora:
Hello elextr.
How are you?
I've made some changes to the DevHelp plugin and it's working fine on Geany. No Action, compilation normally takes place in the Linux environment. But in Windows there is an error. Would you know how to let me know what's going on.

elextr:
@liomarhora probably not, I don't have much time and I don't use devhelp so it may be a long time.

PR #1317, opened 2024-03-05
geanyctags: eliminate some warnings and update the readtags library

techee:


eht16:
LGBI.

I didn't test the changes, @techee I guess you did. If so, I think we can merge this.

PR #1323, opened 2024-04-08
lineoperations: Removed debug code (fixes #1322)

### LINKED ISSUE #1322: Line Operations plugin outputs "Value: 0" to terminal when Geany launches
When running Geany (v1.38) from the terminal with the ***Line Operations*** (AKA `lineoperations`) plugin enabled, Geany issues the string `Value: 0` to the terminal. This almost looks like a return value, but it appears as soon as Geany is launched (not when it terminates).

I searched the Geany and the Geany Plugins documentation and did not find an explanation for the meaning of this value or an official way to instruct Geany or the Plugin to suppress it. I did find that redirecting Geany's standard output to `/dev/null` did hide the seemingly undocumented message: `geany > /dev/null`

Is this message intentional or a bug? If it's intentional, what is the meaning? Can it be added to the documentation (or did my search simply not find it)? Is there an option to suppress it? Can one be added?

See also: https://github.com/geany/geany/issues/3822

Gitoffthelawn:
Fixes #1322

### DIFF
--- a/lineoperations/src/lo_prefs.c
+++ b/lineoperations/src/lo_prefs.c
@@ -117,8 +117,6 @@ lo_init_prefs(GeanyPlugin *plugin)
 	lo_info->use_collation_compare = utils_get_setting_boolean(config,
 		"general", "use_collation_compare", FALSE);
 
-	/* printf("VALUE: %d\n", lo_info->use_collation_compare); */
-
 	g_key_file_free(config);
 }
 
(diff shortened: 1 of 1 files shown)


lineoperations/src/lo_prefs.c
```diff
@@ -117,7 +117,7 @@ lo_init_prefs(GeanyPlugin *plugin)
 	lo_info->use_collation_compare = utils_get_setting_boolean(config,
 		"general", "use_collation_compare", FALSE);
 
-	printf("VALUE: %d\n", lo_info->use_collation_compare);
+	/* printf("VALUE: %d\n", lo_info->use_collation_compare); */
```
eht16, line 120: I would drop it completely. There is no point in having commented debug flying around. If needed, it can be restored from the GIT history.

Gitoffthelawn (commented):
Completely removed the debug statement

eht16:
Thanks!

PR #1324, opened 2024-04-14
Windows installer: Prefer Geany installation directory over user's HOME

### LINKED ISSUE #1321: Default plugins folder incorrect / not found
Fresh install on Windows of Geany and Geany plugins with default settings will install Geany to "C:\Users\Username\Geany  2.0" and the plugins to "C:\Users\Username\Geany-Plugins 2.0". In this configuration Geany does not see the plugins. I have to add "C:\Users\Username\Geany-Plugins 2.0\lib\geany" to the "Extra Plugins Path" in Geany preferences, a step which I do not see documented anywhere.

eht16:
If no explicit installation directory was specified and an existing, Geany installation was found, then use its installation directory.

Closes #1321.

### DIFF
--- a/build/geany-plugins.nsi
+++ b/build/geany-plugins.nsi
@@ -73,7 +73,7 @@ ManifestSupportedOS all
 
 OutFile "${GEANY_PLUGINS_INSTALLER_NAME}"
 
-Var Answer
+Var UserIsAdmin
 Var UserName
 Var GEANY_INSTDIR
 Var UNINSTDIR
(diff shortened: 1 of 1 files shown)


build/geany-plugins.nsi
```diff
@@ -401,6 +398,12 @@ Function .onInit
 		StrCpy $INSTDIR "$GEANY_INSTDIR"
 	${EndIf}
 
+	; if $INSTDIR has not been set yet above, set it to the profile directory for non-admin users
+	${If} $INSTDIR == ""
```
eht16, line 402: The diff is a bit cluttered because of the variable rename (I don't know what was wrong with me naming it "Answer" :( ), the relevant part is here: we set `$INSTDIR` only to the user's profile directory if we didn't find a Geany installation before. Before the Geany installation was ignored if the installer was executed without administrative privileges.

PR #1327, opened 2024-04-16
Vimode: Added folding support

scresto09:
Supported commands are:
za / zo / zc      toggle / open / close fold on one level of folding
zA / zO / zC    toggle / open / close fold on all folding levels
zR / zM           open / close all folds

### DIFF
--- a/vimode/src/Makefile.am
+++ b/vimode/src/Makefile.am
@@ -35,7 +35,9 @@ vi_srcs = \
 	cmds/edit.h \
 	cmds/edit.c \
 	cmds/excmds.h \
-	cmds/excmds.c
+	cmds/excmds.c \
+	cmds/undo.h \
+	cmds/undo.c
 
 vimode_la_SOURCES = \
 	backends/backend-geany.c \
--- a/vimode/src/cmds/edit.c
+++ b/vimode/src/cmds/edit.c
@@ -17,6 +17,7 @@
  */
 
 #include "cmds/edit.h"
+#include "cmds/undo.h"
 #include "utils.h"
 
 
@@ -161,13 +162,7 @@ void cmd_del_word_left(CmdContext *c, CmdParams *p)
 
 void cmd_undo(CmdContext *c, CmdParams *p)
 {
-	gint i;
-	for (i = 0; i < p->num; i++)
-	{
-		if (!SSM(p->sci, SCI_CANUNDO, 0, 0))
-			break;
-		SSM(p->sci, SCI_UNDO, 0, 0);
-	}
+	undo_apply(c, p->num);
 }
 
 
--- a/vimode/src/cmds/excmds.c
+++ b/vimode/src/cmds/excmds.c
@@ -18,6 +18,7 @@
 
 #include "cmds/excmds.h"
 #include "cmds/edit.h"
+#include "cmds/undo.h"
 #include "utils.h"
 
 void excmd_save(CmdContext *c, ExCmdParams *p)
@@ -93,7 +94,7 @@ void excmd_put(CmdContext *c, ExCmdParams *p)
 
 void excmd_undo(CmdContext *c, ExCmdParams *p)
 {
-	SSM(c->sci, SCI_UNDO, 0, 0);
+	undo_apply(c, 1);
 }
 
 
--- a/vimode/src/cmds/undo.c
+++ b/vimode/src/cmds/undo.c
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2024 Sylvain Cresto <scresto@gmail.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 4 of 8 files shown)


techee:
Looks good in general, but I think folding should work when the cursor is anywhere in the block, not just on the fold point (vim seems to work this way). I played with it a little and came up with something like this:
```C
static gint prepare_fold(CmdParams *p)
{
	/* foldparent of the next line */
	gint line = SSM(p->sci, SCI_GETFOLDPARENT, p->line + 1, 0);

	if (p->line == line)
		;  /* we are already on the fold point line */
	else
	{
		/* foldparent of the current line */
		line = SSM(p->sci, SCI_GETFOLDPARENT, p->line, 0);
	}

	if (line != -1)
	{
		/* move the cursor on the visible line before the fold */
		gint pos = SSM(p->sci, SCI_POSITIONFROMLINE, line, 0);
		SET_POS_NOX(p->sci, pos, TRUE);
	}

	return line;
}


void cmd_toggle_fold(CmdContext *c, CmdParams *p)
{
	gint line = prepare_fold(p);
	if (line != -1)
		SSM(p->sci, SCI_FOLDLINE, (uptr_t) line, SC_FOLDACTION_TOGGLE);
}
```

If the code looks OK to you and works as expected, could you update all the folding functions similarly to `cmd_toggle_fold()` above?

In addition, these aren't exactly "edit" commands and should be placed in a separate file. Could you create `fold.c/h`, add them to `Makefile.am` and move all these commands there?

techee:
If you are planning to continue working on this PR, there's one catch. Because of

https://github.com/geany/geany-plugins/commit/33984abc9bd24bbf03db3ac517eb45ca2c7e6b7f

you have to first move the cursor on the visible line before performing a fold - otherwise folding won't work because the above code will expand the fold (I experienced this problem myself when testing the code above and it took me a while to realize what was going on).

scresto09:
OK, I tried to make some changes to improve fold support and react more like vim.
With za/zo/zc we fold the current block as before.
With zA we change the fold depending on the state of the parent, if there is one that is contracted we expand it otherwise we contract it.
I was unable to reproduce the issue you are talking about with commit 33984ab, let me know if the issue persists.
And another thing, sorry but I'm not sure I did the right things with git/merge, let me know if I need to fix anything on this.
Thanks.

vimode/src/cmds/fold.c
```diff
@@ -0,0 +1,130 @@
...
+
+#include "cmds/fold.h"
+#include "utils.h"
+
+/* fold command does not depend on state of parents */
+#define FOLD_NONE		0
+/* fold command depend on state of parents */
+#define FOLD_PARENT		1
+/* fold command depend on state of contracted parent if it exists */
+#define FOLD_CONTRACTED		2
```
techee, line 27: I was super-confused by the naming. I's suggest the following:
* `FOLD_NONE` -> `GOTO_NEAREST_PARENT`
* `FOLD_PARENT` -> `GOTO_TOPMOST_PARENT`
* `FOLD_CONTRACTED` -> `GOTO_CONTRACTED_PARENT`

plus the corresponding  comment changes.

vimode/src/cmds/fold.c
```diff
@@ -0,0 +1,130 @@
...
+
+static gint prepare_fold(CmdParams *p, gint filter)
```
techee, line 29: Rename to `goto_above_fold()` and the `filter` parameter to `type`.

vimode/src/cmds/fold.c
```diff
@@ -0,0 +1,130 @@
...
+
+
+void cmd_open_fold_all(CmdContext *c, CmdParams *p)
+{
+	SSM(p->sci, SCI_FOLDALL, SC_FOLDACTION_EXPAND | SC_FOLDACTION_CONTRACT_EVERY_LEVEL, 0);
+}
+
+
+void cmd_close_fold_all(CmdContext *c, CmdParams *p)
+{
```
techee, line 128: This will need `goto_above_fold(p, GOTO_TOPMOST_PARENT);` otherwise the fold where the cursor is won't get contracted because of what I mentioned before.

vimode/src/cmds/fold.c
```diff
@@ -0,0 +1,130 @@
...
+			prev_line = SSM(p->sci, SCI_GETFOLDPARENT, prev_line, 0);
+			if (prev_line != -1)
+			{
+				line = prev_line;
+				if (filter == FOLD_CONTRACTED && ! SSM(p->sci, SCI_GETFOLDEXPANDED, line, 0))
+					break;
+			}
+		}
+	}
+
```
techee, line 61: Remove extra empty line.

techee (changes requested):
I think it looks good functionality-wise, I have just suggested to rename some things to make them clearer (at least to me).

> I was unable to reproduce the issue you are talking about with commit https://github.com/geany/geany-plugins/commit/33984abc9bd24bbf03db3ac517eb45ca2c7e6b7f, let me know if the issue persists.

When you call `prepare_fold()` the cursor is moved above it so you don't see it. But you can reproduce it with `cmd_close_fold_all()` which currently doesn't call `prepare_fold()` (but it should) and it doesn't collapse the fold where the cursor is placed - see the comment below.

It would be best if you rebased this PR on top of current geany-plugins master as there's a merge conflict in the Makefile.

techee:
One more thing, the README file contains all the supported vi commands - would you copy the added commands from `index.txt` in the vimode directory into README?

techee:
@scresto09 You seem to have pushed an unmodified branch and the pull request closed as a result. There are no commits in this pull request now.

scresto09:
Sorry, I made a mistake and this closed this pull request...

I just created a new pull request [#1350](https://github.com/geany/geany-plugins/pull/1350)  with all these changes.. sorry!

PR #1329, opened 2024-04-17
codenav: Keep "Go to file" dialog open after denying creation of file that wasn't found

gkatev:
Previously, if the file wasn't found and you replied 'no' to the prompt for creating it, all dialogs closed. This can be annoying in situations where the name/path was entered wrong, and you have to start all over again after receiving the file not found error. This commit allows one to amend the mistake from right where they left off.

Demo before:

https://github.com/geany/geany-plugins/assets/31203804/590aee45-ccd4-4996-841e-77464c7a6afa

Demo after:

https://github.com/geany/geany-plugins/assets/31203804/ff4a9203-6a61-4057-9650-24580bf173e3


gkatev:
Superseded by #1390.

PR #1335, opened 2024-04-24
Markdown: Add support for webkit2gtk-4.1

xiota:
Add support for webkit2gtk-4.1.  Drop support for webkitgtk.

Related: #1295


elextr:
LGBI, the removals look good but did not check thats all that needs removing.

elextr:
Needs to update README

b4n (commented):
Looks good, works fine, but the 2 first commits don't really work one without the other, so they should be merged: the first one actually make it use webkit2gtk only, but by removes the conditional that enables the code to use it (and it probably also breaks the build system by having an undefined conditional).

So fine by me once those 2 commits get squashed together.

xiota:
@b4n Thanks... I've squashed the commits.

b4n (approved):
LGTM and seems to work fine here.

sergiomb2:
thank you !  I will check this on Fedora package

PR #1336, opened 2024-04-24
updatechecker: Port to libsoup3 and small other fixes

b4n:
Port to libsoup3 (see https://github.com/geany/geany-plugins/pull/1295#issuecomment-1906999722 @jbicha) plus a couple other fixes I just couldn't leave there after seeing them :slightly_smiling_face: 

**:warning: DISCLAIMER:** I don't know neither libsoup2.4 nor libsoup3.  This is a uneducated guess by looking at [the scarce documentation](https://libsoup.org/libsoup-3.0/migrating-from-libsoup-2.html) and the API reference.  And it seems to work with limited testing (I tested manually only, for the normal case, the 404 case and the incorrect domain case -- all of which got the expected result).

### DIFF
--- a/updatechecker/src/updatechecker.c
+++ b/updatechecker/src/updatechecker.c
@@ -49,9 +49,11 @@ enum {
 
 #define UPDATE_CHECK_URL "https://geany.org/service/version/"
 
+#define UPDATECHECKER_TYPE_KEY "updatechecker-type"
+
 static GtkWidget *main_menu_item = NULL;
-static void update_check_result_cb(SoupSession *session,
-    SoupMessage *msg, gpointer user_data);
+static void update_check_result_cb(GObject *session,
+    GAsyncResult *result, gpointer user_data);
 
 static gboolean check_on_startup = FALSE;
 
@@ -84,15 +86,17 @@ static void update_check(gint type)
 
     g_message("Checking for updates (querying URL \"%s\")", UPDATE_CHECK_URL);
     soup = soup_session_new_with_options(
-            SOUP_SESSION_USER_AGENT, user_agent,
-            SOUP_SESSION_TIMEOUT, 10,
+            "user-agent", user_agent,
+            "timeout", 10,
             NULL);
 
     g_free(user_agent);
 
     msg = soup_message_new ("GET", UPDATE_CHECK_URL);
+    g_object_set_data(G_OBJECT(msg), UPDATECHECKER_TYPE_KEY, GINT_TO_POINTER(type));
 
-    soup_session_queue_message (soup, msg, update_check_result_cb, GINT_TO_POINTER(type));
+    soup_session_send_and_read_async(soup, msg, G_PRIORITY_DEFAULT, NULL,
+                                     update_check_result_cb, msg);
 }
 
 
@@ -185,15 +189,30 @@ version_compare(const gchar *current_version)
 }
 
 
-static void update_check_result_cb(SoupSession *session,
-    SoupMessage *msg, gpointer user_data)
+static gchar *bytes_to_string(GBytes *bytes)
 {
-    gint type = GPOINTER_TO_INT(user_data);
+    gsize bytes_size = g_bytes_get_size(bytes);
+    gchar *str = g_malloc(bytes_size + 1);
+    memcpy(str, g_bytes_get_data(bytes, NULL), bytes_size);
+    str[bytes_size] = 0;
+    return str;
+}
+
+
+static void update_check_result_cb(GObject *session,
+    GAsyncResult *result, gpointer user_data)
+{
+    SoupMessage *msg = user_data;
+    gpointer type_ptr = g_object_get_data(G_OBJECT(msg), UPDATECHECKER_TYPE_KEY);
+    gint type = GPOINTER_TO_INT(type_ptr);
+    GError *err = NULL;
+
+    GBytes *bytes = soup_session_send_and_read_finish(SOUP_SESSION(session), result, &err);
 
     /* Checking whether we did get a valid (200) result */
-    if (msg->status_code == 200)
+    if (bytes && soup_message_get_status(msg) == 200)
     {
-        const gchar *remote_version = msg->response_body->data;
+        gchar *remote_version = bytes_to_string(bytes);
         if (version_compare(remote_version) == TRUE)
         {
             gchar *update_msg = g_strdup_printf(
@@ -216,12 +235,14 @@ static void update_check_result_cb(SoupSession *session,
             }
             g_message("%s", no_update_msg);
         }
+        g_free(remote_version);
     }
     else
     {
         gchar *error_message = g_strdup_printf(
             _("Unable to perform version check.\nError code: %d \nError message: »%s«"),
-            msg->status_code, msg->reason_phrase);
+            soup_message_get_status(msg),
+            err ? err->message : soup_message_get_reason_phrase(msg));
         if (type == UPDATECHECK_MANUAL)
         {
             dialogs_show_msgbox(GTK_MESSAGE_ERROR, "%s", error_message);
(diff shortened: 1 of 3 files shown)


updatechecker/src/updatechecker.c
```diff
@@ -185,15 +191,30 @@ version_compare(const gchar *current_version)
 }
 
 
-static void update_check_result_cb(SoupSession *session,
-    SoupMessage *msg, gpointer user_data)
+static gchar *bytes_to_string(GBytes *bytes)
+{
+    gsize bytes_size = g_bytes_get_size(bytes);
+    gchar *str = g_malloc(bytes_size + 1);
+    memcpy(str, g_bytes_get_data(bytes, NULL), bytes_size);
```
xiota, line 198: I don't know any version of libsoup, but wondering why wrap `g_bytes_get_data()` with another function and memcopy that has to be freed later, rather than `g_bytes_get_data()` directly, and letting `g_bytes_unref()` take care of it.

elextr: Perhaps the response body might not be a nul terminated string, so its copied to one.

xiota: I see... thanks.

b4n: Yeah that's exactly the reason: the bytes have no reason to be NUL-terminated, as it's from reading the raw data from the webserver (and does not explicitly add a NUL itself from what I can gather).  So this helper is to make it a C string, not really caring if there's a NUL earlier because a valid response won't have one anyway.

elextr:
I don't know enough about libsoup to sensibly review (and don't have the time to learn), but since Geany uses GIO why not `g_file_new_for_uri()` and `g_file_read()` (or `g_file_read_async()/g_file_read_finish()` to not block the UI) instead of using libsoup directly?

xiota:
> `g_file_new_for_uri()` and `g_file_read()` (or `g_file_read_async()/g_file_read_finish()`

It works...
```C

#include <gio/gio.h>
#include <stdio.h>

int main() {
  GFile *file = g_file_new_for_uri("https://geany.org/service/version/");

  GError *error = NULL;
  GFileInputStream *stream = g_file_read(file, NULL, &error);

  if (!error) {
    printf("No error.\n");

    guint8 *buffer = NULL;
    gsize size = 0;

    buffer = (guint8 *)g_malloc(32);

    gboolean success = g_input_stream_read_all((GInputStream *)stream, buffer,
                                               32, &size, NULL, &error);

    if (success) {
      buffer[size] = '\0';
      printf("Data: %s\n", buffer);
    } else {
      printf("Error reading stream.\n");
    }

  } else {
    printf("Error reading uri.\n");
  }

  return 0;
}
```

```bash
$ gcc test.c -o test $(pkgconf --libs gio-2.0) $(pkgconf --cflags gio-2.0)

$ ./test
No error.
Data: 2.0.0
```

elextr:
@xiota neat, and simple, although it has one theoretical bug (though unlikely in practice) [hint: what if 32 bytes were returned].

Just for general information, Geany used (a __long__ time ago) to avoid GIO, which is probably why the plugin used libsoup directly.  But now GIO is required by Geany so its available to the plugin.

xiota:
Size check is needed if this route is taken for the plugin...  That code was intended to check whether `https://` is supported because I saw only `file://` mentioned in the docs.  Typical version string is <8 bytes.  Even with words like `alpha` or `pre-release`, I figured 32 bytes is enough for the test.

elextr:
Unfortunately the GIO docs don't say when `g_file_new_from_URI()` and the associated stream was supported (that I could see) but it works on the machine I'm on ATM which is pretty olde.

PS I expect not everyone has your alias pkgconf == pkg-config :wink:

xiota:
> Unfortunately the GIO docs don't say when `g_file_new_from_URI()` and the associated stream was supported (that I could see) but it works on the machine I'm on ATM which is pretty olde.

Looks like it depends on `gvfs`...  After uninstalling...
```
$ ./test
Error reading uri.
```

> PS I expect not everyone has your alias pkgconf == pkg-config 😉

On my computer, `pkg-config` is the symlink.  Looks like [pkgconf](https://gitea.treehouse.systems/ariadne/pkgconf) and [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) are different projects.

elextr:
Yeah, `pkg-config` is the Freedesktop original, `pkgconf` would appear to be a re-implementation.  Your distro installs `pkgconf` and links the original command to it so existing scripts will work, if they happen to conform to what `pkgconf` considers "correct".

If an existing script fails because it depends on some edge case in `pkg-config` tough.  And very aggressive about it they are, "if someone insists on fixing such a non-bug, this constitutes a violation of our [Code of Conduct](https://gitea.treehouse.systems/ariadne/pkgconf/src/branch/master/CODE_OF_CONDUCT.md), which may be addressed by banning from this repository".

xiota:
I use whatever is provided by the package manager.  Even on Debian, pkg-config is a dummy package that redirects to pkgconf.

Their stance doesn't seem particularly "aggressive" in context because they call out "passive-aggressive people who try to argue with us".  Seems they probably had some people who opened numerous issues reporting closely related problems that they were told multiple times are not bugs.  It would be like... what if someone opened multiple issues for Geany complaining that it doesn't support various Qt6 features despite being told multiple times that they aren't bugs?  Eventually, those complaints could be considered intentionally disruptive, and people who "insist" on continuing need to be cut off.

elextr:
> Looks like it depends on gvfs... After uninstalling...

Well, IIUC GVFS is _part_ of GIO (not to be confused with GnomeVFS), so if its uninstalled its not surprising parts of GIO don't work.

But that probably means reading URLs won't work on Windows, dunno about Macos, so back to the soup I guess :stuck_out_tongue_winking_eye:

xiota:
It is separate from GIO.  This is the project page: https://gitlab.gnome.org/GNOME/gvfs

Debian package, appears to be `gvfs-backends`

elextr:
Well, I interpret the very first words in the readme "GVfs is a userspace virtual filesystem implementation for GIO" to mean its part of GIO.  IIUC it provides the implementations for the abstract file/URL/dbus stream operations in GIO and thats why you get an error if its not installed and you try to read anything other than a local file.

Never mind, its was worth a try, and thanks for writing that test program.

xiota:
Someone could run the test program on windows or mac to find out.  I no longer have (easy) access to either.

b4n:
Yeah I guess using GIO directly for this kind of super simple GET would make sense Indeed. I also think that gvfs will indeed be part of all reasonable GIO installations, but I could be wrong.

b4n:
Note that writing an async version of the feature, while reasonably easy, is a little bit more work.

b4n:
See #1340 if wanted.

elextr:
> hum… @elextr @xiota so what's your conclusion? Depend on libsoup3 as this PR because it's HTTP, or depend on a GVFS backend for HTTP?

Waiting for https://github.com/geany/geany-plugins/pull/1340#issuecomment-2079741003 

If it works `?` #1340 `:` #1336

b4n:
This PR won (IMO), so it's the one to review @frlan

b4n:
I rebased this on master now #1342 is merged, as it has CI infrastructure changes for building with libsoup3.  Apart from that it's unchanged from the previous version.

b4n:
@frlan I think this should really land (or be rejected if need be) before 2.1 so we have libsoup3 support for every relevant plugin.

eht16:
I'll test the CI installers in the next days when time permits.

eht16:
Tested on Windows and works.

b4n:
@frlan IMO we *need* this for 2.1, see https://github.com/geany/geany-plugins/pull/1336#issuecomment-2171361614.  I'll merge this in a few days unless somebody complains.

PR #1338, opened 2024-04-25
vimode: Handle folded lines correctly

techee:
This PR addresses some issues of the plugin with folded lines - see the individual commits.

I'd also like to automatically set the cursor position when a block gets folded to be set on the visible line (and not hidden somewhere inside the fold) but right now I don't know how to correctly react to the fold events - see https://sourceforge.net/p/scintilla/bugs/2438/

@scresto09 Do the changes look OK to you and does the plugin behave correctly with them?

### DIFF
--- a/vimode/src/cmd-runner.c
+++ b/vimode/src/cmd-runner.c
@@ -704,6 +704,7 @@ static gboolean process_cmd(CmdDef *cmds, CmdContext *ctx, gboolean ins_mode)
 	{
 		if (orig_mode == VI_MODE_COMMAND_SINGLE)
 			vi_set_mode(VI_MODE_INSERT);
+		ensure_current_line_expanded(ctx->sci);
 	}
 	else if (!consumed && ctx->kpl)
 	{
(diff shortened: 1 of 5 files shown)


scresto09:
Hello,

After read the [https://sourceforge.net/p/scintilla/bugs/2438/](https://sourceforge.net/p/scintilla/bugs/2438/) report, I think the best way to automatically set the cursor position on the visible line is surely to track the SCN_MARGINCLICK event.

I did some modification to handle this with the commit [5608b74](https://github.com/geany/geany-plugins/pull/1326/commits/5608b74cba25b571428da9a055e696ddb053a247) in pull request [#1326](https://github.com/geany/geany-plugins/pull/1326)

What do you think about this ?
Thanks

scresto09:
It seems there are some issues with the changes in commit [907b9bc](https://github.com/geany/geany-plugins/pull/1338/commits/907b9bc9830e601119e1d9df9d39848c3f697932), for example the cursor can no longer go up to the first line, or when some lines wrap.

I've made some changes to try and fix this with the commit [274ff97](https://github.com/geany/geany-plugins/pull/1326/commits/274ff97f51e65b02ae86579768ec8d41f5b5c611) . What do you think about this ?
Thanks

scresto09:
OK, it's works fine, thanks

scresto09:
I just create #1349 for SCN_MARGINCLICK  part.

PR #1343, opened 2024-04-28
latex: improve readme

xiota:
Hoping this PR is quick to review and merge...

This moves info from ABOUT to README for the LaTeX plugin.  This should make the [plugin page](https://plugins.geany.org/latex.html) more informative to potential users.

Also, deletes the ABOUT file from the latex, geanymacro, and geanynumberedbookmarks plugins because they duplicate the contents of README.  These three plugins are the only ones with ABOUT files.  According to `grep`, the files are not referenced by any others (eg, Makefile.am).


latex/README
```diff
@@ -1,4 +1,29 @@
...
+  providing templates
+
+* Easy inserting format patterns like ``\texttt{}`` through menu
+
+* Support on inserting environments by offering an dialog and
+  recognising selections
+
+* Shortcuts for inserting ``\item`` and ``\newline``
+
+* Toolbar with often used format options
```
b4n, line 29: I'd remove blank lines between the items (but it works like that)

b4n (commented):
LGTM, indeed the ABOUTs don't seem useful, and for numberedbookmarks it's an old version of README including typos that have been corrected.

latex/README
frlan: Somehow it feels better to read. I'd keep it.

xiota: It was merged before I even saw this... so guess the newlines stay for now.

elextr:
@xiota exact formatting for READMEs doesn't matter so long as they render ok on `plugins.geany.org` which assumes ReST markup IIRC but I don't know how to see what the new version of the page looks like.

xiota:
> which assumes ReST markup

I added the new lines because it was easier for me to read the text version.  I ran it through pandoc to preview, and saw no difference.  But with a different parser or added stylesheet, there could be extra space between the items.  Like the difference between:

```html
<li>...</li>

<li><p>...</p></li>
```

elextr:
Fine, both you and frlan prefer it so no problem, the source is only read by the developers, not the users.

IIRC rEsT (never remember the caPitaLization ;-) spaces list items independent of blank lines in the source, so should not matter.

The plugin pages in `plugins.geany.org` are made from the READMEs of the individual plugins.  So these get CSS and surrounding content from the process that generates the pages.  

So long as the ResT is valid and it looks ok in https://plugins.geany.org is all that matters.  But as I said, I don't know how to check that.

PR #1348, opened 2024-05-05
geanyctags, projectorganizer, vimode: Some README and URL updates

techee:
1. Suggest using Github for reporting bugs and pull requests instead of sending them by email.
2. Use https:// instead of http://
3. Mention universal-ctags as a maintained version of ctags in geanyctags README

### DIFF
--- a/geanyctags/README
+++ b/geanyctags/README
@@ -11,7 +11,7 @@ GeanyCtags adds a simple support for generating and querying ctags files for a G
 project. It requires that the ctags command is installed in a system path. On
 unix systems, distributions usually provide the ctags package; on Windows, the
 ctags binary can be found in the zip Windows distribution from the ctags home
-page (http://ctags.sourceforge.net).
+page (https://ctags.sourceforge.net).
 
 Even though Geany supports symbol definition searching by itself within the open files
 (and with a plugin support within the whole project), tag regeneration can become
@@ -97,7 +97,7 @@ Downloads
 
 GeanyCtags is part of the combined Geany Plugins release.
 For more information and downloads, please visit
-http://plugins.geany.org/geany-plugins/
+https://plugins.geany.org/geany-plugins/
 
 Development Code
 ================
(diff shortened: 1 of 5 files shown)


geanyctags/README
```diff
@@ -11,7 +11,9 @@ GeanyCtags adds a simple support for generating and querying ctags files for a G
 project. It requires that the ctags command is installed in a system path. On
 unix systems, distributions usually provide the ctags package; on Windows, the
 ctags binary can be found in the zip Windows distribution from the ctags home
-page (http://ctags.sourceforge.net).
+page (https://ctags.sourceforge.net). There is also a new, maintained implementation
```
elextr, line 14: Given the last release of ectags was 2009 it seems better to not mention it. Uctags has 15 years of improvement and makes Windows and mac binaries.

geanyctags/README
```diff
@@ -109,9 +111,10 @@ Get the code from::
 Ideas, questions, patches and bug reports
 =========================================
 
-Please direct all questions, bug reports and patches to the plugin author using the
-email address listed below or to the Geany mailing list to get some help from other
-Geany users.
+Please direct all questions, bug reports and patches to the combined geany-plugins
+project at https://github.com/geany/geany-plugins and open the corresponding
+bug report or pull request there. To notify the author of this plugin about
+your post, mention him using his github user name (@techee).
```
elextr, line 117: Maybe mention that the github username can usually be found in the MAINTAINERS file.

elextr (commented):
Couple of suggestions, otherwise looks good.

geanyctags/README
techee: Good point, I've updated this section. I kept the information about ctags though so people know what the difference between ctags and universal ctags is (I think distributions still package the unmaintained ctags).

geanyctags/README
techee: I'll probably skip this level of indirection and mention it directly in the documentation. I normally check geany-plugins only from time to time but I have an email notification enabled when someone mentions me.

elextr (approved):

PR #1350, opened 2024-05-24
Vimode: implement fold in vimode plugin

scresto09:
Supported commands are:
za / zo / zc toggle / open / close fold on one level of folding
zA / zO / zC toggle / open / close fold on all folding levels
zR / zM open / close all folds


scresto09:
Previous pull request #1327

vimode/README
```diff
@@ -413,6 +413,15 @@ a new command, please do not forget to update the table below.::
...
     yy            ["x]yy             yank N lines [into register x]
+    zA            zA                 open a closed fold or close an open fold
+                                       recursively
+    zC            zC                 close folds recursively
+    zM            zM                 set 'foldlevel' to zero
+    zO            zO                 open folds recursively
+    zR            zR                 set 'foldlevel' to the deepest fold
+    za            za                 open a closed fold, close an open fold
+    zc            zc                 close a fold
+    zo            zo                 open fold
```
techee, line 424: Could you move this block to the same place as in index.txt, i.e. under section "2.5 commands starting with z"? And also within this section to the same place as in index.txt? (of course taking into account that some commands are missing)

vimode/README
```diff
@@ -96,7 +96,7 @@ This is an incomplete list of known limitations of the plugin:
 * named registers and related commands are not implemented
 * Ctrl+X mode is not implemented
 * marks are not implemented
-* fold commands are not implemented
+* most fold commands are not implemented
```
techee, line 99: I'd say "many" would be better as the most important ones are implemented now.

vimode/src/cmds/fold.c
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+#include "cmds/fold.h"
+#include "utils.h"
+
+/* fold command does not depend on state of parents */
```
techee, line 22: I'd say the comments above this define and the 2 other below is useless now and can be removed.

techee (changes requested):
Looks good apart from the several documentation-related changes below.

Once fixed, you can squash the result to a single commit. (And be careful about what git does ;-)

scresto09:
OK I did it

techee:
Looks good, merged. Thanks a lot for all your contributions!!!

scresto09:
Thanks

PR #1351, opened 2024-05-24
New IncDec plugin for Geany

scresto09:
IncDec is a plugin for Geany that provides shortcuts to increment an decrement
number at the cursor, or to the right of the cursor (on the same line).

By default, typing **Shift-+** will increment the next number, and typing **Shift--** will decrement the next number.
You can also increment and decrement number many times by typing **Shift-***. The number can be at the cursor, or to the right of the cursor (on the same line). Numbers can be negative and positive, and can be in decimal or hexadecimal format (hexadecimal number start with **0x**).
You can customize those defaults keys on the preferences, under the Keybindings tab by setting the _IncDec_ keybinding.
An item _Increment or Decrement number_ is also added in the Editor Popup Menu, it can be hidden by changing the Plugins Preferences.


scresto09:
This pull request replace previous #1337

incdec/src/incdec-plugin.c
```diff
@@ -0,0 +1,592 @@
...
+
+        g_signal_connect(plugin_data._menu_item_change_number, "activate", G_CALLBACK(on_change_number_x), NULL);
+}
+
+
+void plugin_cleanup (void)
+{
+	if (plugin_data._dialog)
+	{
+		gtk_widget_destroy (GTK_WIDGET (plugin_data._dialog));
```
eht16, line 582: Here and in some other places you use a space between identifier and opening braces, in other places not.
I suggest to use a consistent style.

incdec/src/incdec-plugin.c
```diff
@@ -0,0 +1,592 @@
...
+
+	plugin_data._menu_item_sep = gtk_separator_menu_item_new();
+	gtk_container_add(GTK_CONTAINER(geany->main_widgets->editor_menu), plugin_data._menu_item_sep);
+
+	plugin_data._menu_item_change_number = gtk_menu_item_new_with_mnemonic(_("_Increment or Decrement number"));
+        gtk_container_add(GTK_CONTAINER(geany->main_widgets->editor_menu), plugin_data._menu_item_change_number);
+
+	configuration_apply();
+
+        g_signal_connect(plugin_data._menu_item_change_number, "activate", G_CALLBACK(on_change_number_x), NULL);
```
eht16, line 574: Here and in some other places, there is mixed indentation, sometimes spaces sometimes tabs.
I suggest to use a consistent style.

incdec/src/incdec-plugin.c
```diff
@@ -0,0 +1,592 @@
...
+		else
+		{
+			format_length = digit_end - digit_start;
+			if (format_length > 12)
+				format_length = 0;
+		}
+
+		g_snprintf(format_buf, sizeof(format_buf)-1, "%%0%d%c", format_length, use_hexa ? ( hexaCase == HEXA_CASE_UPPER ? 'X' : 'x' ) : 'd');
+
+		if ((buf = g_strdup_printf(format_buf, guessed_number)))
```
eht16, line 374: gcc warns here:
```
incdec-plugin.c: In function 'on_change_number':
incdec-plugin.c:374:44: warning: format not a string literal, argument types not checked [-Wformat-nonliteral]
  374 |                 if ((buf = g_strdup_printf(format_buf, guessed_number)))
      |                                            ^~~~~~~~~~
```
Maybe this can be fixed.

eht16 (changes requested):
Nice, thanks!

I added a few minor remarks.
Additionally, could you please update the `MAINTAINERS` file in the root directory?

Thanks.

incdec/src/incdec-plugin.c
```diff
@@ -0,0 +1,592 @@
...
+
+
+void plugin_init (GeanyData *data)
+{
+	GeanyKeyGroup *key_group;
+
+	load_config();
+
+	key_group = plugin_set_key_group (geany_plugin, "incdec", KB_COUNT, on_change_number_callback);
+	keybindings_set_item (key_group, KB_INCREMENT_NUMBER, NULL, GDK_KEY_KP_Add, GDK_SHIFT_MASK,
```
eht16, line 556: GDK_KEY_KP_Add is specific for the + key on the numpad, I think.

The ReadMe mentions Shift-+ and this does not work on my system (German keyboard).
I think this won't work on most keyboard layouts where +, - and * are interpreted with Shift to another key.
So, maybe just make the README (and possibly other references) more explicit.

configure.ac
```diff
@@ -72,6 +72,7 @@ GP_CHECK_VIMODE
 GP_CHECK_WEBHELPER
 GP_CHECK_WORKBENCH
 GP_CHECK_XMLSNIPPETS
+GP_CHECK_INCDEC
```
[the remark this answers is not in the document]
scresto09: Ok I did it

incdec/src/incdec-plugin.c
scresto09: Ok I did it

incdec/src/incdec-plugin.c
scresto09: Yes, you're right, these are not the best keyboard shortcuts.

I have modified the default keyboard shortcuts and updated the files. Now we have to use Shift+A, Shift+X, and Shift+M.

incdec/src/incdec-plugin.c
scresto09: Hum, the code seems correct to me, and it’s really useful for me to use a format that isn’t a "string literal".
I don’t think this is a problem. If you'd like, maybe I could add a DEFINE or a CFLAGS to disable this warning ?

eht16: The best I found is:
```c
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wformat-nonliteral"
		if ((buf = g_strdup_printf(format_buf, guessed_number)))
#pragma GCC diagnostic pop
```

configure.ac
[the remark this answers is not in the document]
eht16: Are you sure you pushed all changes?

The order in `configure.ac` is still wrong, also the inconsistent spaces between function name and opening braces.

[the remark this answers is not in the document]
scresto09: My bad, it should be better now

eht16:
Thanks. I think it's ok to merge.

@scresto09 if you change the code the next time, I suggest to remove the mixed indentation (spaces and tabs). It's not utterly important but makes the code a bit nicer and easier to read.

eht16 (approved):

scresto09:
OK, thanks !

scresto31:
Hi, any update on merging this plugin?  Thanks

b4n (changes requested):

PR #1352, opened 2024-05-25
Windows: Compile GLib schemas after installation

eht16:


build/geany-plugins.nsi
```diff
@@ -152,6 +152,8 @@ Section "Dependencies" SEC04
 SectionEnd
 
 Section -Post
+	Exec '"$INSTDIR\bin\glib-compile-schemas.exe" "$INSTDIR\share\glib-2.0\schemas"'
```
b4n, line 155: How does this behave if the executable is missing?  If it is mandatory, then maybe mark the dependency somehow?
Also, has `$INSTDIR` has to be the same as Geany's?  If not, we won't find the executable here, will we?

b4n: Do we need to re-run this on uninstall to "de-compile" the uninstalled schemas?

eht16: > How does this behave if the executable is missing? If it is mandatory, then maybe mark the dependency somehow? 

Not sure, I'll test this.
We could add the GLib package to the plugins bundle, it would be redundant to the one in the Geany bundle but probably won't hurt.

> Also, has `$INSTDIR` has to be the same as Geany's? If not, we won't find the executable here, will we?

True. But much more things won't work then, the compiled schemas will probably only searched in Geany's installation directory. Using another installation directory for G-P doesn't make much sense at all.

> Do we need to re-run this on uninstall to "de-compile" the uninstalled schemas?

In theory, yes. In reality, we do not uninstall the installed schema files. To do so, we had to track each installed file to know what to uninstall.
This would be nice but also quite some work.

b4n: OK.  If `$INSTDIR` ought to be the same, maybe just make the 2.1 installer depend on Geany 2.1's one and call it a day.

eht16: > OK. If `$INSTDIR` ought to be the same, maybe just make the 2.1 installer depend on Geany 2.1's one and call it a day.

Good point, done.

> > How does this behave if the executable is missing? If it is mandatory, then maybe mark the dependency somehow?
> 
> Not sure, I'll test this. We could add the GLib package to the 

Tested: the installer silently ignores a missing "glib-compile-schemas". In this case, this is good I'd say.

b4n: Ok yeah it's probably at least good enough.

b4n (approved):
LGTM

PR #1353, opened 2024-05-26
ProjectOrganizer, Workbench: Use GStatBuf for "g_stat()" calls

eht16:
For some reason, since the latest MSYS2 update on my Windows system, I got compiler errors that `g_stat()` requires a `GStatBuf` type as second argument.

According to the docs, https://docs.gtk.org/glib/func.stat.html, this is correct and in Geany we also use `GStatBuf` in all places.


techee (approved):
LGTM, thanks!

PR #1355, opened 2024-06-08
projectorganizer: Use GtkFileChooserNative when configured in Geany

techee:
This is a preparation in case https://github.com/geany/geany/pull/3861 gets merged in Geany.


b4n (commented):
LGBI

PR #1358, opened 2024-06-09
Addons: Remove obsolete GTK_CHECK_VERSION checks

eht16:


addons/src/addons.c
```diff
@@ -652,9 +652,7 @@ static GtkWidget *plugin_addons_configure(G_GNUC_UNUSED GeanyPlugin *plugin, Gtk
 
 	gtk_widget_show_all(vbox);
 
-#if ! GTK_CHECK_VERSION(2, 10, 0)
 	gtk_widget_hide(check_systray);
-#endif
```
b4n, line 657: Looks backwards, should you also remove the body?  I guess the idea was not to add the option when the feature wasn't supported, wasn't it?

b4n (changes requested):
LGTM apart from the comment

addons/src/addons.c
eht16: This is so me...you are completely right.
Thanks for spotting, fixed.

Funny thing, this attempt to hide the option didn't work at all because in Geany itself we call `gtk_widget_show_all()` when creating the plugin preferences dialog and so it was shown again.
Anyway, GTK2 is long time ago.

b4n (approved):
LGBI with a little squashing

PR #1359, opened 2024-06-09
Addons: Hide status icon on startup until Geany has started

eht16:
This avoids showing the status icon without an icon on startup and even if it is disabled while Geany is starting.


frlan:
I cannot reproduce the initial issue (even with a lot of documents etc.) but the fix seems to don't break anything -- LGTM ;)

PR #1372, opened 2024-09-12
Surround word with "\\<" and "\\>" for search_current().

Shooter23:
The crosshatch (#) and asterisk (*) commands in Vim perform exact word searches. For example, with the cursor on "word", the asterisk would perform "/\\<word\\>" to search for the next exact occurrence of "word".


techee:
Seems to work. Thanks!

Issue #1374, opened 2024-09-23
[Geanylua] geany.choose doesn't list all files


Johnmcenroyy:
There is an example:
https://github.com/geany/geany-plugins/blob/master/geanylua/examples/info/list-open-files.lua

It gets info about all opened files in Geany and open this list in new tab of Geany or lists them in choose menu.
List in new tab works as expected but geany.choose function not - it lists only two first opened files.

elextr:
Please note that Geanylua is orphaned and so unmaintained.

Johnmcenroyy:
@elextr yes, I know but maybe someone knows why this happened, will try to find myself also, but how I understand it is better to move to Peasy plugin (just compiled) with Libpeas support. Is Geanylua planned to be removed from Geany plugins ?

Johnmcenroyy:
Overall good simple plugin, but pity seems not all work.

elextr:
Well clearly nobody needs it urgently enough to be willing to support it.

There is no specific policy for removing plugins from the plugins collection, in fact sometimes the threat of removal results in individual PRs to keep an almost dead plugin going (usually something like moving to a new library version as happened to Geanylua and whatever changes that needs).  But that is individual PRs, it often does not result in a rush of offers to take up ownership of the plugin.

I don't know how many geanylua issues there are, but in the past many scripts have worked just fine, guess they didn't use your troublesome function.  So that makes it even harder to let it go, even if its no longer perfect.  I suspect it will limp along until either some major issue occurs or the oldest version of lua that is provided by distros is not compatible with the newest version of lua that geanylua uses.  Or a white knight thunders in on a massive horse and takes over maintenance. :horse_racing: 

Merging Peasy to Geany-plugins is an issue because Peasy uses its own compile and build system, not Autotools.  And also if the creator does not want to include it in Geany-plugins thats entirely their right, and its not unusual.  Of course "somebody" could make a copy of Peasy and include it in Geany-plugins, thats GPL, but having done that for a plugin in the past myself, I would say its not a happy position to be, even if the originator didn't mind.

Unfortunately for Peasy it appears only Arch AUR seems to supply it as a package, not Ubuntu and dependants or Fedora (dunno about Debian, the package search is down, but its likely that Ubuntu would have it if Debian did).  So Peasy is not really an option for those who don't compile stuff even though its a technical improvement.

Johnmcenroyy:
[...]

Issue #690, opened 2018-02-09
Problem with "Lua Script" after upgrade (GTK 3)


lingvoboard:
**Before upgrade**

Geany 1.32, GTK 2.24.32, GLib 2.54.3
geany-1.32.0-1-x86_64.pkg.tar.xz, geany-plugins-1.32-2-x86_64.pkg.tar.xz
Arch Linux, Theme: BlueMenta

Plugin "Lua Script"
`geany.choose("Snippets:", {"bla", "bla", "bla"})`
http://geany-plugins.sourceforge.net/geanylua/geanylua-ref.html#select

![Screenshot1:](https://s10.postimg.org/tbkoqhvex/Screenshot_2018-02-09_17-41-25.png)

**After upgrade**

Geany 1.32, GTK 3.22.26, GLib 2.54.3
geany-1.32.0-2-x86_64.pkg.tar.xz, geany-plugins-1.32-4-x86_64.pkg.tar.xz

![Screenshot2:](https://s10.postimg.org/h9pawejmh/Screenshot_2018-02-09_17-42-16.png)

codebrainz:
GTK+ 3.22 broke all themes again, try an up to date theme like Adwaita or other popular one that has been fixed since.

Skif-off:
GTK2 vs. GTK3:
![geany choose](https://user-images.githubusercontent.com/6333198/98442970-c5236280-2121-11eb-995b-c6a2065680bf.png)
Xubuntu 20.04, Geany 1.37, GTK 3.24.20, GLib 2.64.3.

Is this the same problem? I tried Adwaita, Greybird (my default theme), Numix, but with the same result. Is this exactly a theme problem?
Maybe someone know, how to fix it?

b4n:
Doesn't look like a theme problem from @Skif-off's screenshots (I can't see the first one from the OP BTW), looks like incorrect widget packing with GTK3's defaults.

b4n:
@Skif-off try:
```diff
diff --git a/geanylua/glspi_dlg.c b/geanylua/glspi_dlg.c
index f0553853..c7d9718a 100644
--- a/geanylua/glspi_dlg.c
+++ b/geanylua/glspi_dlg.c
@@ -163,10 +163,14 @@ static gint glspi_choose(lua_State* L)
 	scroll=gtk_scrolled_window_new(NULL, NULL);
 	gtk_scrolled_window_set_policy (GTK_SCROLLED_WINDOW (scroll),
[...]

PR #1376, opened 2024-10-05
FIX: GeanyLua: 'Choose' dialog size/appearance (b4n's patch)

Skif-off:
See issues [690](https://github.com/geany/geany-plugins/issues/690), [1374](https://github.com/geany/geany-plugins/issues/1374).


b4n (approved):
LGTM.  Only thing is that it's probably not necessary to keep GTK2 support given it has been dropped from Geany since 1.38; but it's fine to keep the check and do this later on.

xiota:
I've updated the lua plugin and updated status to "odd fixes".  I've confirmed issue #1374 and that this PR fixes it.

@Skif-off Up to you whether you want to remove the gtk version check before merge.  There are others, so they can be sweeped up together in a future PR.  If you intend to leave the check, please comment, so your intention is clear.

Skif-off:
Removed.

xiota (approved):

xiota:
@b4n I think this is ready to merge.

PR #1381, opened 2024-11-02
Add .editorconfig so github uses the correct tab size (4)

techee:
See the related discussion in https://github.com/geany/geany/pull/1193.

The branch with this commit is here

https://github.com/techee/geany-plugins/tree/editorconfig

so the effect of this PR can be examined by browsing the code there.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+indent_size = 4
+end_of_line = lf
+
+[*.{c,h,cxx,am,m4}]
+indent_style = tab
+
+[*.{py}]
+indent_style = space
+


techee:
Not really sure why the unit test fails but seems to be unrelated to the patch.

eht16:
LGTM

eht16:
> Not really sure why the unit test fails but seems to be unrelated to the patch.

New cppcheck-version-fun again.
https://github.com/geany/geany-plugins/pull/1382 tries to fix it. For some reason, new cppcheck versions error out on the name "true" when used as a function.
Luckily, this function doesn't seem to be used at all, so removing it was an easy solution.

PR #1382, opened 2024-11-03
Geanyprj: remove unused function "true"

eht16:
This also fixes a strange error with recent cppcheck version.


elextr:
"true" is a keyword [since C23](https://en.cppreference.com/w/c/keyword) thats the problem.

b4n:
LGTM, merged.  This function seems to never have been used, possibly a C&P leftover

PR #1389, opened 2024-11-25
git-changebar: Rename bool for C23 compatibility

cjmayo:
C23 is adding a bool keyword:

https://en.cppreference.com/w/c/keyword/bool

With GCC 15 (in development, bug fixing) this causes a failure: `error: expected identifier or '(' before 'bool'`


b4n:
Sounds silly that C23 didn't make it a tad more backward compatible, but the changes look OK from afar, I'll try to test this.

b4n:
Seems to work fine, thanks! (not having GCC 15 at hand, I faked a similar problem with a temporary `#define bool int` which showed issues similar to what you described, and was happy after your patch).

PR #1399, opened 2025-01-02
Windows: Add ".exe" suffix to post install commands

eht16:
Newer Wine versions require the extension.

Relates to
https://github.com/geany/geany/pull/4088 and
https://github.com/geany/infrastructure/pull/12.


techee:
I've just tried the geany-plugins installer from this PR and a few random plugins from it and it seems to work alright. So I think this PR can be merged.

PR #1404, opened 2025-01-07
Remove redundant check

keremet:
The iter variable has already passed the same check in the for statement


b4n (approved):
Not tested, but LGTM

PR #1408, opened 2025-02-02
fixes for building on windows/msys2

giuspen:
- version locally the script geany-plugins-release.py so far hosted in the wiki
- applied fix for building without signing (missing key/certificate in the filesystem)
- applied fix to geanygendoc/docs/Makefile.am
- applied fix to lsp/src/lsp-server.c
- applied fix for languages binaries not being generated
- changed version num to 2.1


b4n:
Coule you please split each item to its own commit? We prefer having smaller more specific commits leading to the final stage. Of course each commit should be complete in regard of what it is meant to achieve, but each of these should try and be small. It also gives you the opportunity to actually explain why you changed something, both to help reviewers and future understanding of why a change has been made.

geanygendoc/docs/Makefile.am
```diff
@@ -15,6 +15,6 @@ dist_pluginhtmldoc_DATA = \
 if BUILD_RST
 manual.html: manual.rst manual.css html4css1.css
 	$(AM_V_GEN) $(RST2HTML) -d --strict \
-		--stylesheet-path $(abs_srcdir)/html4css1.css,$(abs_srcdir)/manual.css \
+		--stylesheet-path $(srcdir)/html4css1.css,$(srcdir)/manual.css \
```
b4n, line 18: Using `abs_srcdir` was done on purpose, see de89404b0edec7cc6a9808e94dd2559ebabdbe10.
If somehow this breaks MSYS2, it can be weighted again as it's not 2016 anymore, but you'd have to verify no relevant version of the tool is still impacted by the issue.

po/LINGUAS
```diff
@@ -0,0 +1 @@
+be ca da de el es fr gl it ja kk nl pt pt_BR ru tr uk zh_CN
```
b4n, line 1: What's up about the i18n changes?  It looks like a partial replication of geany#2938, but without actually switching over to the "modernized" i18n setup I can't see where this is going.  What is this *actually* supposed to fix?  Again, explaining changes it very important :slightly_smiling_face:

geanygendoc/docs/Makefile.am
b4n: What does the command-line look like? would quoting the paths help?

giuspen: I reverted this change as I found that if I build from the root of the repo it works

po/LINGUAS
giuspen: I reverted this change as I found that if I build from the root of the repo it works

geanygendoc/docs/Makefile.am
b4n: That's even more bizarre given they were *absolute* paths… but glad it worked.

We can revisit that in a more specific changest if it's useful, but again it'd need verifying it doesn't "unfix" something else.

po/LINGUAS
b4n: OK cool.  We can see to switch to the same technique as is used in Geany itself if it helps, but it's a tad more involved -- but IIUC that would be a good thing to avoid deprecated stuff.

eht16: Two comments:
- we have #1183 which could make this better, even if not, it is a good change for itself (this is the pendant of geany#2938)
- I vagely remember some issue after the last or second last release when the translation files were not properly updated in the Windows build. Unfortunately, I didn't find any related issue or PR, neither in Geany nor in G-P :(.

lsp/src/lsp-server.c
```diff
@@ -45,6 +45,7 @@
 # include "spawn/lspunixoutputstream.h"
 #endif
 
+#include  <unistd.h>
```
eht16, line 48: Is this change on purpose?

If so, @techee should add it upstream as well.

techee: No problem to add but is it actually needed @giuspen ? The plugin seemed to build without it on my machine at least.

giuspen: Hi @techee have you tried to build in msys2 or you are referring to cross compiling? I couldn't on windows/mays2 without this patch but I can revert and reapply when I build locally anyway, not a big deal

techee: I tried it but it's possible it was before making some changes in `lsp-server.c` that require this header. I was just wondering if you ran into any problems without it.

No problem to add this include and make the corresponding change in the lsp repo.

b4n: FWIW, it's most probably because of the `getpid()` call, whose declaration indeed is in *unistd.h*.

eht16: The include of the `unistd.h` header has been merged via #1421.
This will lead to a merge conflict here, @giuspen maybe you want to rebase or merge the branch on current master.

giuspen: Thanks @eht16 I will rebase shortly resolving the conflict

build/msys2-geany-plugins-release.py
```diff
@@ -0,0 +1,122 @@
...
+# adjust paths to your needs ($HOME is used because expanduser() returns the Windows home directory)
+BASE_DIR = join(os.environ['HOME'], 'geany_build')
+SOURCE_DIR = join(os.environ['HOME'], 'git', 'geany-plugins')
+RELEASE_DIR_ORIG = join(BASE_DIR, 'release', 'geany-plugins-orig')
+RELEASE_DIR = join(BASE_DIR, 'release', 'geany-plugins')
+BUNDLE_BASE_DIR = join(BASE_DIR, 'bundle')
+BUNDLE_GEANY_PLUGINS = join(BASE_DIR, 'bundle', 'geany-plugins-dependencies')
+INSTALLER_NAME = join(BASE_DIR, f'geany-plugins-{VERSION}_setup.exe')
+
+# signing params
```
giuspen, line 29: Similarly to what discussed in https://github.com/geany/geany/pull/4223 as @b4n questioned the signing part of the script, @eht16 then commented that he explained parts of the signing topic in https://github.com/geany/geany/issues/4189#issuecomment-2629001461 and we will probably not sign the binaries any longer and so we could also remove the related code.

giuspen:
Here is the Geany-Plugins updated script and workflow action to build on native windows.
I have been debugging with the help of Google Gemini after an issue with the autotools and it is currently building again https://github.com/giuspen/geany-plugins/actions/runs/15662106128/job/44121339614 (may still fail but I'm doing progress...) but I got to leave now and will be able to try new things only in the evening

configure.ac
```diff
@@ -1,5 +1,5 @@
 AC_PREREQ([2.61])
-dnl Don't forget when updating version number to do GP_CHECK_GEANY also
+dnl Dont forget when updating version number to do GP_CHECK_GEANY also
```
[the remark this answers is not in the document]
giuspen: sorry forgot to revert that... ironically Geany messes up the syntax highlighting with the single quite in the autotools file, a bug of Scintilla I guess... I will revert that anyway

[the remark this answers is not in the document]
Kreijstal: you can use "do not" :)

giuspen:
This is now complete same as https://github.com/geany/geany/pull/4223 but here we are building both installers. As mentioned in the other PR, I will need both merged before I can proceed writing to SignPath as I will need to indicate the workflow that we intend to use for signing. I think it will be this one, but this also fetches from the other repo... in fact this points to giuspen/geany I will have to change that, but only after you merge the other PR first on Geany.

.github/workflows/windows-msys2-build.yml
```diff
@@ -65,6 +65,7 @@ jobs:
           rsync -a --delete ${DESTINATON}/build/geany/ ${DESTINATON}/release/geany-orig
       - name: CI-Build-Geany-Plugins
         run: |
+          find . -type f \( -name "configure.ac" -o -name "Makefile.am" -o -name "*.m4" -o -name "*.in" -o -name "*.pc.in" \) -exec dos2unix {} \;
```
b4n, line 68:
```suggestion
          find . -type f \( -name "configure.ac" -o -name "Makefile.am" -o -name "*.m4" -o -name "*.in" \) -exec dos2unix {} \;
```

giuspen:
> > It was all because of the Autotools don't like CRLF, apparently considered CR as an empty line, I sorted out with a dos2unix call in the workflow
> 
> I'm a bit puzzled by why you'd get CRLFs in the checkout (we don't have attributes… but then, maybe we should have a `*.m4 -text`?), even more why it might not affect _all_ files… but well, if it works I guess.

I was surprised as well as cloning a repo from the msys2 shell it doesn't happen, I believe it is the GitHub checkout action on a native windows builder that behaves like that.

.github/workflows/windows-msys2-build.yml
```diff
@@ -0,0 +1,98 @@
...
+          mkdir -p geany_build/bundle/geany-plugins-dependencies
+          cd geany_build/bundle/geany-plugins-dependencies
+          bash $(cygpath -u "${GITHUB_WORKSPACE}")/build/gtk-bundle-from-msys2.sh -3
+      - name: CI-Build-Geany
+        run: |
+          DESTINATON=$(cygpath -u "${GITHUB_WORKSPACE}")/geany_build
+          cd .geany_source
+          VERSION=$(autom4te --no-cache --language=Autoconf-without-aclocal-m4 --trace AC_INIT:\$2 configure.ac)
+          NOCONFIGURE=1 ./autogen.sh
+          export lt_cv_deplibs_check_method=${lt_cv_deplibs_check_method='pass_all'}
```
b4n, line 58: what's this about?

giuspen: > what's this about?

I remember I was finding an error, googled for it and found this as the solution.
I later noticed that Enrico did the same, see https://github.com/geany/geany-plugins/blob/b95b09c124c24c12046dd2660b53821affd6b73d/build/ci_mingw64_geany_plugins.sh#L79

b4n:
Actually, maybe
```diff
diff --git a/.gitattributes b/.gitattributes
new file mode 100644
index 00000000..bf98e73e
--- /dev/null
+++ b/.gitattributes
@@ -0,0 +1,2 @@
+*.m4 text eol=lf
+*.ac text eol=lf
```
would be better, not sure.

giuspen:
> ```diff
> eol=lf
> ```

This solution works. Pushed to this PR.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+*.m4 -text eol=lf
+*.ac -text eol=lf
```
b4n, line 2: I don't think `eol` has any effect without `text`, so it should either be `-text` *or* `text eol=lf`

giuspen: Ok I'll try that

giuspen: After removing the dashes, workflow still working for me

eht16:
I'm sorry @b4n that now you also dive so deep into all this :(.

eht16:
Then we don't need to install the "dos2unix" tool any longer?

giuspen:
No we still use it on the python script!

b4n:
Is this ready as well, e.g. regarding what generate which installer(s)?  I'm afraid I kind of lost track here, and am too lazy (and don't have time right now actually) to re-walk the whole history.
If all is well, we can squash and merge here as well.

giuspen:
Actually no wait @b4n I need to edit the second repo name it is still looking for giuspen, will do in a moment

giuspen:
Done

b4n:
@eht16 if the workflow rum succeeds, would you squash & merge? I probably won't be able to do any of this this WE, maaayybe tomorrow but no guarantee

eht16:
Here we go: https://github.com/geany/geany-plugins/actions/runs/15912668412 :pray: :crossed_fingers:

PR #1412, opened 2025-02-16
CI: Update to Ubuntu 22.04 runners as 20.04 are getting deprecated

eht16:
See https://github.com/actions/runner-images/issues/11101


b4n (approved):
LGTM

Issue #1218, opened 2023-01-21
Numbered Bookmarks - Modified file Notification-Dialog may be Excessive (and not necessary)


neoh4x0r:
I've noticed that the accept/reject notification dialog is displayed for files which have been modified outside of geany (which may or may not have been changed), at least I don't remember having made changes, outside of geany, every time it is displayed. 

I find my self just clicking **ignore** (which seems to be often) and just moving on.

I'm wondering if:
1. Is the notification dialog really necessary? (eg. could it just force the rejection and log a status message)
2. or maybe the display of the notification dialog could be a plugin configuration option (if it is disabled -- then maybe it could be auto-rejected like in the first part and perhaps that could be configured as well)

Long story short -- I really don't think displaying a modal notification, when a file is loaded, is really necessary when all someone wants to do (is ignore/reject) and load the standard bookmarks (which from my experience aren't saved by geany).

----

This following code blob in **geanynumberedbookmarks.c** is responsible for showing the dialog:

https://github.com/geany/geany-plugins/blob/ddf06e8a1ce945a9d3dbdf482dc9aeb8478f080c/geanynumberedbookmarks/src/geanynumberedbookmarks.c#L967-L981

elextr:
> I'm wondering if: The notification dialog is really necessary? (eg. could it just force the rejection and log a status message)

Well, only you know if its an error with the timestamp, or something "touch"ed the file without changing it, or you edited the file with __ANOTHER__ editor (gasps of horror at such infidelity ;-), the plugin can't tell, so it asks you.

IIRC the discussion at the time saving bookmarks and folds was contentious because if the file was changed they won't be in the same place.  So the modification check was used to reduce that risk.  

Better that you find why you get timestamp mismatches.  If something touches the file just don't do that.  If its an error in the timestamp there are several possible causes.  Depending on the OS and the filesystem the timestamp may not be updated immediately after the file is saved, so the value the plugin reads is the previous time it was saved.  This is particularly likely for remote filesystems.

PR #1413, opened 2025-02-18
numberedbookmarks additional options

caclark:
Additional options for numberedbookmarks
1. Always save bookmarks, even if file has not changed
2. Always try to load bookmarks, even if file has changed

### DIFF
--- a/geanynumberedbookmarks/README
+++ b/geanynumberedbookmarks/README
@@ -78,7 +78,7 @@ Save file settings...
 Remember normal Bookmarks
     If this is set then the plugin will remember standard non-numbered
     bookmarks, and restore them when the file is next loaded.
-Try to load bookmarks
+Force loading bookmarks if file has changed
     Always try to load markers, even if the file has changed since last save.
 Always save bookmarks
     Always save markers, even if the file has not changed.
--- a/geanynumberedbookmarks/README
+++ b/geanynumberedbookmarks/README
@@ -78,6 +78,10 @@ Save file settings...
 Remember normal Bookmarks
     If this is set then the plugin will remember standard non-numbered
     bookmarks, and restore them when the file is next loaded.
+Try to load bookmarks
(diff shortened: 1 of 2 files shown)


eht16:
Looks good to me and works as promised.
Thanks.

Maybe the title of the option "try to load bookmarks" could be rephrased to better explain what this option does. Maybe something like "Force loading bookmarks if file has changed".

@WilliamFraser are you still around and what to review this change as well?

eht16:
I think https://github.com/geany/geany-plugins/issues/1218 can be closed by this change as well.

WilliamFraser:
Life is crazy busy, but I'm still around.
It's a good upgrade that makes it clearer what it does, and gives the user more options.

caclark:
I am unfamiliar with making pull requests to GitHub.
Can anyone see why the PR has failed the Linux Build check?

b4n:
@caclark it's a bit weird, it fails on tests unrelated to your changes, but a PR with the same parent commit from today succeeded.  I tried re-running the test just in case, we'll see what happens.

b4n:
OK it seems fine this time, might have been a glitch of some sort. @eht16 any reason why cppcheck would give different results between seemingly identical runs? Otherwise I might try to have a look off I find time (i.e. not in this lifetime 😁)

b4n:
OK I just had a quick look and we build upstream releases of cppcheck, and there was a new one a couple days ago. I thought other builds succeeded before that, but maybe not and the new release fixed it.

eht16:
> OK I just had a quick look and we build upstream releases of cppcheck, and there was a new one a couple days ago. I thought other builds succeeded before that, but maybe not and the new release fixed it.

Now, the master branch failed, it seems in the same way as before. cppcheck. Sigh.

I might have a look at this next week. But as you said, it seems unrelated to this PR.

PR #1416, opened 2025-03-03
Fix inconsistent icon sizes in treebrowser

chrisgoldberg1:
The treebrowser plugin displayed content-type specific icons at their native resolution, causing some file types (like CSS) to appear with oversized icons. This patch ensures all icons are consistently sized.


b4n (changes requested):
You probably should rather use the `GTK_ICON_LOOKUP_FORCE_SIZE` flag for the `gtk_icon_theme_lookup_by_gicon()` calls than reimplementing this manually. Unless it doesn't work somehow?

PR #1417, opened 2025-03-16
Fix nullPointerOutOfMemory warnings of cppcheck 2.17.0

eht16:
The latest cppcheck version (2.17.0) introduces a new check to check for properly handling errors on memory allocation failures (https://sourceforge.net/p/cppcheck/news/2025/02/cppcheck-2170/).

Multiple plugins are affected, some more, others less.

I wonder if it is worth to update all occurrences or ignore this check completely since at least the GLib memory allocation functions seems to terminate the whole application on errors (at the beginning of https://docs.gtk.org/glib/memory.html) and so our code to handle the errors probably will never be called.

On the opther hand, it might be still useful if any memory allocation returns NULL because of invalid input or other reasons without terminating the whole application.

I made a start in this PR but it's probably only about 5-10% of the necessary changes and in many places I just added TODOs to properly handle errors.

What do you think? @b4n @techee

### DIFF
--- a/addons/src/ao_xmltagging.c
+++ b/addons/src/ao_xmltagging.c
@@ -93,7 +93,7 @@ void ao_xmltagging(void)
 			/* Getting the tag */
 			tag = g_string_new(gtk_entry_get_text(GTK_ENTRY(textbox)));
 
-			if (tag->len > 0)
+			if (tag && tag->len > 0)
 			{
 				gsize end = 0;
 				gchar *end_tag;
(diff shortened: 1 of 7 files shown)


b4n:
As you said glib terminates on allocation failure (but for explicit cases like `g_try_malloc()` & co). IMO it doesn't make sense to riddle the code with checks for allocation failures that can't happen, it just adds noise. There is a reason why glib decided to terminate the program for most allocation failures: it's a lot of work to handle it properly everywhere, and many times there's not much of anything sensible to do anyway, so usually trying to recover will just lead to a chain of failures, probably ending up in terminating the program anyway as nothing is working anymore anyway.

So I'd say we should ignore those, and probably check our setup or tell cppcheck these just doesn't make sense.

techee:
I agree with Colomban - if there's a way to suppress these warnings, it would probably be the best way to handle it.

eht16:
Closed in favor of #1419.

PR #1419, opened 2025-03-23
cppcheck: Disable nullPointerOutOfMemory check

b4n:
This produces too many false-positives with GLib code.

See https://github.com/danmar/cppcheck/pull/7068 and https://github.com/danmar/cppcheck/pull/7390.

Closes #1417.


eht16 (approved):
Great. :+1: 

Thanks for doing this and also for reporting/changing it on cppcheck!

PR #1427, opened 2025-04-26
geanyctags: Fix navigation queue

keremet:
When activating the "Find Tag Declaration" menu item the line number in a new document was saved as a position to navigate back. We should save the same line number in the old document.


techee:
Thanks for noticing, merged.

PR #1428, opened 2025-05-04
Add Github action workflow to create Windows installer for releases

eht16:
This adds a new Github workflow which is only triggered after pushing new tags.

So we could easily create a Windows release installer by just pushing a new tag and no manual release build is required any more. The release process will get a bit simpler and less dependent on single persons.

To easily test this workflow I created a temporary repository and the resulting installer can be found there: https://github.com/eht16/geany-plugins-tmp/actions/runs/14820385881

See also https://github.com/geany/geany/pull/4238.

### DIFF
--- a/.github/workflows/release-windows.yml
+++ b/.github/workflows/release-windows.yml
@@ -1,0 +1,107 @@
+#
+# Copyright:	2022, The Geany contributors
+# License:		GNU GPL v2 or later
+
+name: Release Windows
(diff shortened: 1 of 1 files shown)


eht16:
Closing in favor of https://github.com/geany/geany-plugins/commit/2107bbdf69266987a79a661a49bc8d9ecc48c57d.

PR #1434, opened 2025-05-23
Use standard gettext, following Geany core - continued

eht16:
Based on @kugel-'s work in https://github.com/geany/geany-plugins/pull/1183, rebased on current master and updated gettext to 0.23.


configure.ac
```diff
@@ -16,9 +16,11 @@ AM_PROG_CC_C_O
 AC_DISABLE_STATIC
 AC_PROG_LIBTOOL
 
-dnl i18n
-IT_PROG_INTLTOOL([0.35.0])
-GP_I18N
+dnl i18n, inline so autoreconf can detect gettext usage
+AM_GNU_GETTEXT_VERSION([0.23.1])
```
eht16, line 20: I hope this not too new and only relevant when running "autogen.sh".

With the old version, I got errors when running "make" like:
```
*** error: gettext infrastructure mismatch: using a Makefile.in.in from gettext version 0.19 but the autoconf macros are from gettext version 0.22
```

I hope this is OK.

eht16: I reverted this to the version 0.19 as in Geany, this makes the CI happy.

b4n: Maybe that's the kind of problem newer versions tried to fix, but if we don't have the issue in Geany we should be able to figure something out.

eht16: It seems we need to define `AM_GNU_GETTEXT_REQUIRE_VERSION`.

From my tests, it doesn't make a difference if we define it instead of or additionally to `AM_GNU_GETTEXT_VERSION`, Windows build succeeds once it is set.

Yet, I still have no clue why, what's the difference to Geany and what exactly these two options really do :(.

@b4n @techee do you have any clue or do we want to use it this way? I would cleanup the commits then.

techee: Oficially, I have no clue about anything related to autotools. And whatever anyone does to make it work is more than enough for getting it merged. :-)

autogen.sh
```diff
@@ -1,11 +1,16 @@
 #!/bin/sh -e
 
+srcdir=`dirname $0`
+test -z "$srcdir" && srcdir=.
+
 mkdir -p build/cache
-intltoolize -c -f
-autoreconf -vfi
+
+(cd $srcdir; autoreconf --install --verbose)
```
b4n, line 8: Why is `--force` (`-f`) dropped here?  Without it I'd have to cleanup some auto-copied files manually before I can use this:
```console
$ NOCONFIGURE=1 ./autogen.sh 
autoreconf: export WARNINGS=
autoreconf: Entering directory '.'
autoreconf: running: autopoint
autopoint: File build/cache/nls.m4 has been locally modified.
autopoint: File po/Makefile.in.in has been locally modified.
autopoint: *** Some files have been locally modified. Not overwriting them because --force has not been specified. For your convenience, you find the local modifications in the file '/tmp/artiVqY7/gtX8ppkg/autopoint.diff'.
autopoint: *** Stop.
autoreconf: error: autopoint failed with exit status: 1
```

With it, it happily replaces the files it wants, which non are tracked and we rely on autoreconf to copy over.  So, I'm wondering if there's a reason *not* to force-overwrite them, or if it just makes our lives that much harder if we change anything that affects it.  I see it's the same in Geany now, but I'm still puzzled.

b4n: (BTW -- and I'm not complaining, it's something probably unavoidable -- but one will *have* to re-run autogen, just `make` won't cut because of the changed build system dependencies, Automake can't magically fix it for us)

configure.ac
```diff
@@ -1,13 +1,14 @@
...
 dnl Don't forget when updating version number to do GP_CHECK_GEANY also
 AC_INIT([geany-plugins], [2.1])
-AM_INIT_AUTOMAKE([1.8 foreign dist-bzip2 tar-ustar subdir-objects])
-
-m4_ifdef([AM_SILENT_RULES],[AM_SILENT_RULES([yes])])
 
 AC_CONFIG_SRCDIR([po/POTFILES.in])
 AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_MACRO_DIR([build/cache])
+AC_CONFIG_AUX_DIR([build/aux])
```
eht16, line 8: This is not strictly related to the PR but I stole it from Geany to reduce the clutter of generated files if using in-source builds.

autogen.sh
eht16: > Why is `--force` (`-f`) dropped here? Without it I'd have to cleanup some auto-copied files manually before I can use this:
> 
> ```
> $ NOCONFIGURE=1 ./autogen.sh 
> autoreconf: export WARNINGS=
> autoreconf: Entering directory '.'
> autoreconf: running: autopoint
> autopoint: File build/cache/nls.m4 has been locally modified.
> autopoint: File po/Makefile.in.in has been locally modified.
> autopoint: *** Some files have been locally modified. Not overwriting them because --force has not been specified. For your convenience, you find the local modifications in the file '/tmp/artiVqY7/gtX8ppkg/autopoint.diff'.
> autopoint: *** Stop.
> autoreconf: error: autopoint failed with exit status: 1
> ```
> With it, it happily replaces the files it wants, which non are tracked and we rely on autoreconf to copy over. So, I'm wondering if there's a reason _not_ to force-overwrite them, or if it just makes our lives that much harder if we change anything that affects it. I see it's the same in Geany now, but I'm still puzzled.

I had this as well a few times but not always. This is probably part of the puzzle what's different here from Geany. 

Thomas removed the `--force` in https://github.com/geany/geany/pull/3092 but I guess you already have seen this.

I don't mind much, if it works better with `--force` and it affects only files not checked in in GIT, then it's probably fine to use.

build/i18n.m4
```diff
@@ -1,16 +0,0 @@
-AC_DEFUN([GP_I18N],
-[
-    if test -z "$conf_dir" ; then
-        conf_dir="."
-    fi
-    ALL_LINGUAS=`cd "$conf_dir/po" 2>/dev/null && ls *.po 2>/dev/null | $AWK 'BEGIN { FS="."; ORS=" " } { print $[1] }'`
```
b4n, line 6: Funnily enough, this looks entirely broken for me on my fresh-ish Debian Bookworm install: I'm not sure where `conf_dir` is supposed to come from, but it's nowhere to be found here, and thus just results in no linguas for out-of-tree builds.  Oopsie, right?  `s/conf_dir/srcdir/` works though, and that's a documented Autoconf variable, whereas I couldn't find references to `conf_dir` in anything current.

configure.ac
b4n: Fair enough, but don't use `aux` as a directory name, it can [break under Windows](https://www.gnu.org/software/autoconf/manual/autoconf-2.60/autoconf.html#File-System-Conventions):

> * Invalid names (lfn)
>   *   Some DOS file names are reserved, and cause problems if you try to use files with those names. These names include `CON`, `AUX`, `COM1`, `COM2`, `COM3`, `COM4`, `LPT1`, `LPT2`, `LPT3`, `NUL`, and `PRN`. File names are case insensitive, so even names like `aux/config.guess` are disallowed.

autogen.sh
b4n: I didn't find that one yet, I was not digging deep yesterday, but found it today before seeing your message… so I'd say we should *probably* keep `--force` and drop any possibly updated files, those are useless in Git (e.g. INSTALL if it's not from us)

b4n: @eht16 looking at it I don't think it's gonna cause any issue.  I'm not sure why we track `INSTALL` in Geany, but we don't here, and I don't see any other problematic tracked files.
So are you OK for adding back the `--force` option and simplify upgrades?

autogen.sh
```diff
@@ -1,11 +1,16 @@
...
 
 if [ "$NOCONFIGURE" = 1 ]; then
     echo "Done. configure skipped."
     exit 0;
 fi
-exec ./configure "$@"
+
+echo "Running $srcdir/configure $@ ..."
+$srcdir/configure "$@" && echo "Now type 'make' to compile." || exit 1
```
b4n, line 16:
```suggestion
"$srcdir/configure" "$@" && echo "Now type 'make' to compile." || exit 1
```

autogen.sh
b4n:
```suggestion
(cd "$srcdir"; autoreconf --force --install --verbose)
```

configure.ac
```diff
@@ -16,9 +17,12 @@ AM_PROG_CC_C_O
 AC_DISABLE_STATIC
 AC_PROG_LIBTOOL
 
-dnl i18n
-IT_PROG_INTLTOOL([0.35.0])
-GP_I18N
+dnl i18n, inline so autoreconf can detect gettext usage
+#AM_GNU_GETTEXT_VERSION([0.19.8])
+AM_GNU_GETTEXT_REQUIRE_VERSION([0.19.8])
```
b4n, line 22: what's up with this in the end?

po/Makevars
```diff
@@ -0,0 +1,82 @@
...
+
+# This is the copyright holder that gets inserted into the header of the
+# $(DOMAIN).pot file.  Set this to the copyright holder of the surrounding
+# package.  (Note that the msgstr strings, extracted from the package's
+# sources, belong to the copyright holder of the package.)  Translators are
+# expected to transfer the copyright for their translations to this person
+# or entity, or to disclaim their copyright.  The empty string stands for
+# the public domain; in this case the translators are expected to disclaim
+# their copyright.
+COPYRIGHT_HOLDER = The Geany contributors
```
b4n, line 25:
```suggestion
COPYRIGHT_HOLDER = The Geany-Plugins contributors
```

maybe?  Not sure if this matters or what though

configure.ac
eht16: Did I ever mention how much I love Windows?

Anyway, I will rename it to "build/build-aux". A little redundant but still just for generated files.

po/Makevars
```diff
@@ -0,0 +1,82 @@
...
+
+# Usually the message domain is the same as the package name.
+DOMAIN = $(PACKAGE)
+
+# These two variables depend on the location of this directory.
+subdir = po
+top_builddir = ..
+
+# These options get passed to xgettext.
+XGETTEXT_OPTIONS = --keyword=_ --keyword=N_ --keyword=C_:1c,2 --keyword=NC_:1c,2 --from-code=UTF-8
```
b4n, line 15: We should extract comments for translators, there's at least one properly marked (others didn't look intentionally for translators but just extracted automatically)
```suggestion
XGETTEXT_OPTIONS = --keyword=_ --keyword=N_ --keyword=C_:1c,2 --keyword=NC_:1c,2 --from-code=UTF-8 -c"Translators:"```

File: build/i18n.m4 (hunk 4 above)
Comment (eht16) [reply build/i18n.m4:6]: But this whole file is gone and `LINGUAS` is now checked in, to be updated manually.

File: configure.ac (hunk 6 above)
Comment (eht16) [reply configure.ac:22]: See my comment in https://github.com/geany/geany-plugins/pull/1434#discussion_r2105801339

Comment (b4n) [reply configure.ac:22]: Oops, my bad, I got lost with all these.
OK, looking at [the docs](https://www.gnu.org/software/gettext/manual/html_node/autopoint-Invocation.html) it seems like using the latter is fine: AFAIK we don't really care which version is used so long as it works as expected.
Also, this macro seems [old enough](https://lists.gnu.org/archive/html/autoconf-patches/2015-10/msg00000.html) for our requirements I guess

Comment (b4n) [reply configure.ac:22]: So I'd say just drop the commented-out `AM_GNU_GETTEXT_VERSION` and call it good

File: po/Makevars [hunk 9]
```diff
@@ -0,0 +1,82 @@
...
+
+# This is the list of locale categories, beyond LC_MESSAGES, for which the
+# message catalogs shall be used.  It is usually empty.
+EXTRA_LOCALE_CATEGORIES =
+
+# This tells whether the $(DOMAIN).pot file contains messages with an 'msgctxt'
+# context.  Possible values are "yes" and "no".  Set this to yes if the
+# package uses functions taking also a message context, like pgettext(), or
+# if in $(XGETTEXT_OPTIONS) you define keywords with a context argument.
+USE_MSGCTXT = no
```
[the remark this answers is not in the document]
b4n: Nah, leave it for now I'll try and see the incidence it might have

po/Makevars
```diff
@@ -12,7 +12,7 @@ subdir = po
 top_builddir = ..
 
 # These options get passed to xgettext.
-XGETTEXT_OPTIONS = --keyword=_ --keyword=N_ --keyword=C_:1c,2 --keyword=NC_:1c,2 --from-code=UTF-8
+XGETTEXT_OPTIONS = --keyword=_ --keyword=N_ --keyword=C_:1c,2 --keyword=NC_:1c,2 --from-code=UTF-8 -c"Translators:"```
```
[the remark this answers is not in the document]
eht16: Oopsie, four eyes were not enough :).

configure.ac
eht16: Alright, done.

po/Makevars
```diff
@@ -0,0 +1,82 @@
...
+
+# Usually the message domain is the same as the package name.
+DOMAIN = $(PACKAGE)
+
+# These two variables depend on the location of this directory.
+subdir = po
+top_builddir = ..
+
+# These options get passed to xgettext.
+XGETTEXT_OPTIONS = --keyword=_ --keyword=N_ --keyword=C_:1c,2 --keyword=NC_:1c,2 --from-code=UTF-8 -c"Translators:"
```
b4n, line 15: Beware: as we changed this since last re-generation, running update-po might loos the `#|` entries.  We might then want to revert the PO files to their previous state, and only *then* run update-po to get as much info as possible.  Not sure if the loosing is expected though, but that's what I noticed.

eht16: If I understand you correctly, we could drop the commit e86ae47a64ac257ac09d98b2ad81a335b4cf9f1f and regenerate the files again?

b4n: Yes that's my idea

po/Makevars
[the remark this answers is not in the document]
b4n: Digging a tad, it looks like it doesn't change anything with msgfmt and xgettext >= 0.15, and given we ask for a newer version anyway it seems irrelevant.  So no need to worry about this, *but* the right value might still be `yes`.  But it doesn't matter.

b4n:
OK now it seems to actually work.  Sadly enough, it seems nobody actually tested this before, so no translation were actually loaded :)

Anyway, it looks mostly decent now, I'll try and rebase+squash it a bit now.

configure.ac
```diff
@@ -1,13 +1,14 @@
 AC_PREREQ([2.61])
 dnl Don't forget when updating version number to do GP_CHECK_GEANY also
 AC_INIT([geany-plugins], [2.1])
-AM_INIT_AUTOMAKE([1.8 foreign dist-bzip2 tar-ustar subdir-objects])
-
-m4_ifdef([AM_SILENT_RULES],[AM_SILENT_RULES([yes])])
```
b4n, line 6: @eht16 can you tell whether moving this was needed or not in the end?

b4n: I'm testing without, we'll see

b4n: (it's actually required to have `AM_INIT_AUTOMAKE` *after* `AC_CONFIG_AUX_DIR` on all platforms… so if I had tested before asking I'd have known).

geanylua/autogen.sh
b4n, line 1: Removing this file is unrelated to this PR, and I just noticed it now by grepping for intltool.  If we don't want to have this change here I can pull the commit out and do this separately, otherwise having it in a series that touch the build system is not making me scream either.

b4n (approved):
LGTM now (but given I have changed it quite some, another review would be good :))

eht16:
I have removed the commit 2f192292606ff213d4a1d1ec7e03dd229fc8a596 ("Fix Windows build").
It contained the change from `AM_GNU_GETTEXT_REQUIRE_VERSION` to `AM_GNU_GETTEXT_VERSION` and some debug output.
The debug output was dropped and the macro change was moved to the very first commit.

eht16:
> OK now it seems to actually work. Sadly enough, it seems nobody actually tested this before, so no translation were actually loaded :)
> 
> Anyway, it looks mostly decent now, I'll try and rebase+squash it a bit now.

According to https://github.com/geany/geany-plugins/pull/1183#issuecomment-2395400026, I tested it "roughly". Either it worked at this time or my testing was too lax :(.

Tested this time more thorough and it works for me, on Linux.

@giuspen or @techee could you maybe test the Windows installer from the CI build?

b4n:
> @giuspen or @techee could you maybe test the Windows installer from the CI build?

That would be great -- and don't forget to try non-C (and non-English would be easier) locale!

giuspen:
I will have a go tomorrow and report the result

geanylua/autogen.sh
b4n: Thank you for revamping this, debugging it and getting the ball rolling :)

b4n:
I just added a small fix to simplify our lives, same as Geany's geany/geany@35043402562a8ba8566f0cd4d7be6a1c94f006ba.
Otherwise, looks good to merge :+1:

techee:
> Unless there are objections, I think I'll merge this in a couple days™ 🙂

TM is mine :-)

eht16:
> …and it's fixed, unveiling what broke use of `AM_GNU_GETTEXT_VERSION` in the Windows build: the `--install` flag in `ACLOCAL_AMFLAGS`. Don't ask me why, but I guess it leads to copying over copies of the gettext macros. I like it better that way for 3 reasons:

I'm all fine with this.
The only reason I added `AM_GNU_GETTEXT_REQUIRE_VERSION` was that it fixed the Windows build. Though I did not understand the difference and I probably had never found the `--install` flag in `ACLOCAL_AMFLAGS`.

Thank you!

> Unless there are objections, I think I'll merge this in a couple days™ 🙂

Yes, please.

eht16:
P.S.: I will probably miss this commit which made me laughing a lot but it's ok :).

Screenshot_2025-05-28_22-08-26

eht16:
> Maybe I will setup a CI build using a windows-latest runner and msys2 similar to what I am using ( https://github.com/giuspen/cherrytree/blob/master/.github/workflows/windows-msys2.yml ) 

Sounds good. Fine for me as long as I do not have to maintain it :). And I see that the current Docker-MSYS2-Crossbuild-Setup is not very tempting to dig in.

> Note that "MSYS2 recommends UCRT64 nowadays as the default instead of MINGW64" ( from https://github.com/msys2/setup-msys2?tab=readme-ov-file#options ) so I migrated to it recently, maybe we should do the same in Geany.

I don't know any blockers regarding UCRT64. When I started with Mingw and later MSYS2, I think there was no UCRT64 yet or I wasn't aware of it and then never migrated the build setup.

PR #1438, opened 2025-05-26
geanygendoc: Use one single stylesheet for compatibility

b4n:
Some rst2html versions do not handle more than one relative-path stylesheet, and some Windows setup do not like `$(abs_srcdir)`.

So, workaround all this by manually combining the stylesheets in one single generate one, which makes it both only one, and relative to the build directory, both of which should make this simpler and more compatible.

@giuspen, does this fix your Windows setup?


giuspen:
Builds OK now also on Windows/MSYS2, thanks @b4n

b4n:
@giuspen cool, thanks for the test!

PR #1444, opened 2025-06-09
SpellCheck: Update Scintilla styles for new filetypes in Geany 2.1

eht16:
The usual update of Scintilla styles for which spell checking should be performed.

I don't know the new filetypes well enough but guessed according to previous mappings :smile: .

### DIFF
--- a/spellcheck/src/speller.c
+++ b/spellcheck/src/speller.c
@@ -879,6 +879,28 @@ gboolean sc_speller_is_text(GeanyDocument *doc, gint pos)
 			}
 			break;
 		}
+		case SCLEX_DART:
+		{
+			switch (style)
+			{
+				case SCE_DART_COMMENTLINE:
+				case SCE_DART_COMMENTLINEDOC:
+				case SCE_DART_COMMENTBLOCK:
+				case SCE_DART_COMMENTBLOCKDOC:
+				case SCE_DART_STRING_SQ:
+				case SCE_DART_STRING_DQ:
+				case SCE_DART_TRIPLE_STRING_SQ:
+				case SCE_DART_TRIPLE_STRING_DQ:
+				case SCE_DART_RAWSTRING_SQ:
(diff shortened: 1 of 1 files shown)


b4n (changes requested):
What about COBOL?  It seems to have new specific styles now.

I'd also consider `SCE_HJ_TEMPLATELITERAL` and `SCE_HJA_TEMPLATELITERAL`, which IIUC are f-string-like stuff.

And you need to add the brand new `SCE_DART_STRINGEOL`, `SCE_ZIG_STRINGEOL`, `SCE_NIX_STRINGEOL` and `SCE_TOML_STRINGEOL` :slightly_smiling_face:

techee:
> @techee OK. I still believe it'd be more future-proof to have them listed

OK, in that case, should I update Geany style mappings with these?

b4n:
@techee Don't you think it makes it easier whenever the COBOL lexer against incompatible styles?  Which IIUC is the reason Scintilla started having its own styles.

If you do, yes, that'd be neat.  But it can wait after 2.1 as well, whichever.

eht16:
> What about COBOL? It seems to have new specific styles now.

I added the specific COBOL styles, seems to work as expected.

> I'd also consider `SCE_HJ_TEMPLATELITERAL` and `SCE_HJA_TEMPLATELITERAL`, which IIUC are f-string-like stuff.

Ah thanks, added.

> And you need to add the brand new `SCE_DART_STRINGEOL`, `SCE_ZIG_STRINGEOL`, `SCE_NIX_STRINGEOL` and `SCE_TOML_STRINGEOL` 🙂

Done.
I'm not too sure if unterminated strings are worth for spell checking but since for other languages, they were already present, it's fine to add these too.

techee:
> @techee Don't you think it makes it easier whenever the COBOL lexer against incompatible styles? Which IIUC is the reason Scintilla started having its own styles.

Agree, see https://github.com/geany/geany/pull/4337.

PR #1448, opened 2025-06-16
Updated Spanish translation for Geany 2.1

lucasvieites:


b4n:
> Unsure this will break anything

It won't break anything at least

PR #1449, opened 2025-06-17
French translation update

b4n:
@techee The endless LSP menu is a nightmare regarding mnemonics… 😅

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -9,8 +9,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Geany-Plugins 2.0\n"
 "Report-Msgid-Bugs-To: https://github.com/geany/geany-plugins/issues\n"
-"POT-Creation-Date: 2025-06-15 20:12+0200\n"
-"PO-Revision-Date: 2018-02-19 17:07-0600\n"
+"POT-Creation-Date: 2025-06-18 00:03+0200\n"
+"PO-Revision-Date: 2025-06-18 00:05+0200\n"
 "Last-Translator: Colomban Wendling <ban@herbesfolles.org>\n"
 "Language-Team: French <geany-i18n@uvena.de>\n"
 "Language: fr\n"
@@ -3710,12 +3710,11 @@ msgstr "Enregistrer les marque-pages normaux"
 
 #: geanynumberedbookmarks/src/geanynumberedbookmarks.c:1284
 msgid "force loading bookmarks if file has changed"
-msgstr ""
+msgstr "forcer le chargement des marque-pages si le fichier a changé"
 
 #: geanynumberedbookmarks/src/geanynumberedbookmarks.c:1289
-#, fuzzy
 msgid "always save bookmarks"
-msgstr "Liste des marque-pages"
+msgstr "toujours enregistrer les marque-pages"
 
 #: geanynumberedbookmarks/src/geanynumberedbookmarks.c:1308
 msgid "Numbered Bookmarks help"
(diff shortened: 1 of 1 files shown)


techee:
> @techee The endless LSP menu is a nightmare regarding mnemonics… 😅

Frankly, I'm not sure all of them are unique even in the English version. It was like "I'll use this letter as it seems it hasn't been used anywhere else" without making really sure ;-)

b4n:
> Frankly, I'm not sure all of them are unique even in the English version. It was like "I'll use this letter as it seems it hasn't been used anywhere else" without making really sure ;-)

I bet :smile: -- and no, that didn't work.

<details><summary>FWIW, issues reported in the LSP menu for C language by <a href="https://github.com/b4n/geany-plugin-internal-mnemonics-check">geany-plugin-internal-mnemonics-check</a></summary>

### Conflicting mnemonics inside "_LSP Client":

* a
  * Show _All Diagnostics
    * candidates: o
  * Go to _Anywhere...
    * candidates: o y
* c
  * _Clear Highlighted
    * candidates: none. An option might be to change another item's mnemonic to free an option for this one, but remember that changing existing mnemonics should be avoided when possible not to break user habits.  Another option might be rewording this item, or, for non-latin languages, add a new mnemonic in parentheses.
  * Show _Current File Diagnostics
    * candidates: o
* d
  * Go to _Document Symbol...
    * candidates: o m y b
  * Go to _Definition
    * candidates: o
* h
  * Show _Hover Popup
    * candidates: o v
  * _Highlight Symbol Occurrences
    * candidates: y m b o
* p
  * _Project Configuration
    * candidates: o j
  * Rename in _Project...
    * candidates: m o j
  * Go to _Previous Diagnostic
    * candidates: o v
* r
  * _Restart All Servers
    * candidates: v
  * _Rename Highlighted
    * candidates: m
  * Find _References
    * candidates: none. An option might be to change another item's mnemonic to free an option for this one, but remember that changing existing mnemonics should be avoided when possible not to break user habits.  Another option might be rewording this item, or, for non-latin languages, add a new mnemonic in parentheses.
* s
  * _Server Initialize Responses
    * candidates: v z o
  * _Show Current Diagnostic
    * candidates: o
+ this menu already uses the following mnemonics:
  * a (See above)
  * c (See above)
  * d (See above)
  * e (Go to D_eclaration)
  * f (_Format Code)
  * g (_Global Configuration)
  * h (See above)
  * i (Find _Implementations)
  * l (Go to _Line...)
  * n (Go to _Next Diagnostic)
  * p (See above)
  * r (See above)
  * s (See above)
  * t (Go to _Type Definition)
  * u (_User Configuration)
  * w (Go to _Workspace Symbol...)
</details>

frlan:
I wouldn't mind to change them also in 2.1 ;) 

However, this is another topic

b4n:
> I wouldn't mind to change them also in 2.1 ;)

It'll change quite a few strings in LSP, are you sure its OK? We can manually fix it for already updated translations as it doesn't really affected the translated parts, but it's a bit of a tedious job.

frlan:
> > I wouldn't mind to change them also in 2.1 ;)
> 
> It'll change quite a few strings in LSP, are you sure its OK? We can manually fix it for already updated translations as it doesn't really affected the translated parts, but it's a bit of a tedious job.

It's good question always. But maybe you are right. I though that we most likely will not get too many complete translations on plugins anyway as it'S really a huge amount of work. This change would add more work onto pile of transslators, but users might not even see the translated strings anyway.

techee:
I would personally leave the change after the release. I think the mnemonics aren't useful that much as the items in the menu (maybe apart from the configuration) aren't meant to be used normally and users should rather assign keybindings to the individual actions they want to use.

The primary functions of the menu are:
1. Improved discoverability of individual features the plugin offers.
2. A way to test what the given thing does without having to assign a keybinding first.
3. A way to see what features the used LSP server offers; features not offered by the server are insensitive and clearly visible. This gives users some feedback so they won't be surprised why e.g. symbol goto doesn't work when the server doesn't support it.

PR #1450, opened 2025-06-17
lsp, projectorganizer: Fix translator comments

b4n:


b4n:
I can be merged now, no worries, it only gives additional context for translators. But it will appear in .po files only after next update -- which we might want to do, @frlan ?  Its not critical though.

Issue #1447, opened 2025-06-16
Crash with unloaded webkit-using plugins


b4n:
Not sure what's going on, but I got a crash 2 times trying out LSP this morning:

<details><summary>(uninformative) GDB output</summary>

```gdb
$ gdbr /tmp/_install_gtk3/bin/geany -vc /tmp/_conf /home/ban/progs/ctpl/ctpl.geany 
GNU gdb (Debian 13.1-3) 13.1
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /tmp/_install_gtk3/bin/geany...
Starting program: /tmp/_install_gtk3/bin/geany -vc /tmp/_conf /home/ban/progs/ctpl/ctpl.geany
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Geany-INFO: 11:43:29.949: Using alternate configuration directory
[New Thread 0x7ffff53ff6c0 (LWP 1915023)]
Geany-INFO: 11:43:29.956: Geany 2.1 (git >= 0e11e4be7), unknown
Geany-INFO: 11:43:29.956: GTK 3.24.38, GLib 2.74.6
Geany-INFO: 11:43:29.956: OS: Debian GNU/Linux 12 (bookworm) (bookworm)
Geany-INFO: 11:43:29.956: System data dir: /tmp/_install_gtk3/share/geany
Geany-INFO: 11:43:29.956: User config dir: /tmp/_conf
(geany:1915020): GLib-GIO-DEBUG: 11:43:29.962: Using cross-namespace EXTERNAL authentication (this will deadlock if server is GDBus < 2.73.3)
[New Thread 0x7ffff4b856c0 (LWP 1915024)]
[...]

PR #1452, opened 2025-06-19
LSP: Update to the latest version of jsonrpc-glib

techee:
@b4n Not sure if it could possibly fix https://github.com/geany/geany-plugins/issues/1447 but might be worth trying if you are able to reproduce the problem.


techee:
I think it's better to wait for an official jsonrpc-glib release as the problem is somewhere else. Closing now.

PR #1453, opened 2025-06-20
LSP: Remove keybindings for goto definition/declaration

techee:
There are already keybindings for standard Geany goto definition/declaration that will get propagated to the LSP server so there is not a real need for another pair of keybindings doing nearly the same. This is confusing for users (see https://github.com/techee/geany-lsp/discussions/79).

The only exception when this could get useful is when users configure the plugin not to use LSP for goto in which case standard Geany goto is used when using Geany goto definition/declaration. In this case, the extra keybinding would still use LSP features and users could choose between the two by using two different keybindings. However, such as feature is hard to discover and probably not used by many and better to make the UI less confusing for users.

@frlan @b4n This patch removes 2 translatable strings. Probably not a problem, just can the patch be merged now or should it wait until all translators submit their updates?


b4n:
> @frlan @b4n This patch removes 2 translatable strings. Probably not a problem, just can the patch be merged now or should it wait until all translators submit their updates?

It's not a problem, just don't run `update-po` to avoid merge conflicts.

frlan:
LGTM

PR #1454, opened 2025-06-22
codenav: Fix crash caused by missing NULL check

### LINKED ISSUE #1406: codenav plugin makes geany crash when adding new file types
Geany crashes when adding new file type to the Code Navigator plugin.

1.) install Code navigator plugin, and select it in the plugins list
2.) go to Preferences
3.) press Add
4.) add new extension
5.) press Apply

Result: Geany quits immediately

techee:
Fixes #1406

### DIFF
--- a/codenav/src/codenavigation.c
+++ b/codenav/src/codenavigation.c
@@ -261,7 +261,7 @@ on_configure_response(GtkDialog* dialog, gint response, gpointer user_data)
 									
 			/* If one field is empty, ignore this line (it will be replaces
 			   at next execution) */ 
-			if ( strlen(impl_list[i])==0 || strlen(head_list[i])==0 )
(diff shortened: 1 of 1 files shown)


codenav/src/codenavigation.c
```diff
@@ -261,7 +261,8 @@ on_configure_response(GtkDialog* dialog, gint response, gpointer user_data)
 									
 			/* If one field is empty, ignore this line (it will be replaces
 			   at next execution) */ 
-			if ( strlen(impl_list[i])==0 || strlen(head_list[i])==0 )
+			if ( !impl_list[i] || strlen(impl_list[i])==0 ||
```
b4n, line 264: BTW, it's not this patche's fault, and many compilers are likely gonna optimize some cases, but it's a bit silly to use `strlen(x) == 0`, `! *x` is gonna be quite a tad faster.

PR #1459, opened 2025-06-27
overview: disable focus of the overview widget

### LINKED ISSUE #731: Overview "breaks" focus cycle
When tabbing through the interface to change the currently focused widget, one gets stuck when reaching the overview.

Steps to reproduce:
1. Enable Overview plugin
2. Get keyboard focus to the message window (e.g. click on one of the status lines)
3. hit <kbd>Shift+Tab</kbd> twice: first one will reach the msgwindow's tab itself (e.g. "Status"), second will reach overview

At this point, it's not possible to move anymore using <kbd>Tab</kbd> or <kbd>Shift+Tab</kbd>.  I don't believe this is on purpose, or has any usefulness as the overview isn't editable.

Proposed fix: mark the overview itself as not focusable (`gtk_widget_set_can_focus(self, FALSE)`) so it's skipped over.  Alternatively, if you want to e.g. allow arrows to move in the overview, make sure to left focus cycling through.

techee:
Thanks to Colomban Wendling for reporting the issue and fixing it 7 years ago ;-)

Fixes #731.

### DIFF
--- a/overview/overview/overviewscintilla.c
+++ b/overview/overview/overviewscintilla.c
@@ -691,6 +691,8 @@ overview_scintilla_init (OverviewScintilla *self)
   memcpy (&self->overlay_color, &def_overlay_color, sizeof (OverviewColor));
   memcpy (&self->overlay_outline_color, &def_overlay_outline_color, sizeof (OverviewColor));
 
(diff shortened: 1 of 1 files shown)


eht16:
LGTM and works.

PR #1461, opened 2025-06-29
geanynumberedbookmarks: make the plugin work when numlock is on

techee:
Thanks to Colomban Wendling for the patch.

A lower-risk pre-release variant of https://github.com/geany/geany-plugins/pull/1458.

### DIFF
--- a/geanynumberedbookmarks/src/geanynumberedbookmarks.c
+++ b/geanynumberedbookmarks/src/geanynumberedbookmarks.c
@@ -1482,6 +1482,7 @@ static gboolean Key_Released_CallBack(GtkWidget *widget, GdkEventKey *ev, gpoint
 {
 	GeanyDocument *doc;
 	gint i;
+	GdkModifierType state = keybindings_get_modifiers(ev->state);
 
 	doc=document_get_current();
 	if(doc==NULL)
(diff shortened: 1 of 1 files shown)


eht16:
LGTM though I have surprising test results:

Either I did something wrong when testing or it is like this on a German keyboard layout:
- without the patch Ctrl-Shift-<anynumer> does not set a bookmark at all, neither with top row number keys nor with numpad keys, regardless of numlock setting
- with the patch: Ctrl-Shift-<anynumer> *does* set bookmarks but only with top row number key, not with numpad keys, regardless of numlock setting

So, for me it is better than before but the numpad keys still don't work.

I'll try to do more testing tomorrow.

techee:
> Either I did something wrong when testing or it is like this on a German keyboard layout:

Yes, it's the German (Czech, French, etc.) keyboard.

The real full fix with international keyboard support is https://github.com/geany/geany-plugins/pull/1458 - it's just biggish so I wasn't sure if OK for 2.1.

b4n (approved):
LGTM (obviously), that's a minimal fix to get the NumLock issue sorted.

b4n:
> So, for me it is better than before but the numpad keys still don't work.

This isn't supposed to work 🙂

b4n:
#1458 is merged, closing this.

Issue #1460, opened 2025-06-27
CI Windows build: unnecessary jobs?


eht16:
Feel free to ignore this until 2.1 is released, there is no rush need for this.

I didn't notice earlier and I'm not yet fully sure but wondered if the jobs
- GTK-Bundle-Geany
- Compile NSIS-Geany
are still necessary?

Is the Geany installer relevant for the code signing process?
Or is it maybe some sort of left over from previous attempts?

b4n:
@giuspen probably something you might be able to help on :)

giuspen:
@eht16 is right, as we do not intend to make use of the geany installer generated from the geany-plugins workflow, it makes sense to remove unneeded steps GTK-Bundle-Geany and Compile NSIS-Geany
I can apply those changes when (hopefully) SignPath replies to my email as they should generate geany and geany-plugin projects and I will need to update the workflows on both geany and geany-plugins to include the signing of the installers

PR #1470, opened 2025-07-05
Gp msys2 native workflow remove geany installer

giuspen:
Addresses https://github.com/geany/geany-plugins/issues/1460
Should run the workflow automatically, so if we get a green light I will remove the auto trigger and then can hopefully be merged by Monday


b4n:
Merged the first commit as 35e4e27042e96ebfec4fc7abd2ce71b094a8fa65

PR #1471, opened 2025-07-06
geanylua: Fix a few warnings and potential crashes

b4n:
:warning: Beware, this is entirely untested, beside compiling OK.


xiota (approved):

xiota:
Looks good to merge.  Tested with [examples/work/01.edit-test-script.lua](https://github.com/geany/geany-plugins/blob/master/geanylua/examples/work/01.edit-test-script.lua) and following script:
```lua
local files={}

for file in geany.documents()
do
  table.insert(files,file)
end

if ( #files > 0 )
then
  file=geany.choose("Choose a tab to activate:", files)
  if (file)
  then
    geany.save(file)
    geany.activate(file)
  end
else
  geany.message("No files!")
end
```

xiota:
@b4n I think this is good to merge.

PR #1476, opened 2025-07-09
lsp: Fix build with GLib 2.58

b4n:
@techee First 2.1 issue breakage :)  Current LSP doesn't build on GLib 2.58 (GP only depend on 2.56 through Geany 2.1).

We have 2 options:
* Make LSP depend on 2.68 (for `g_string_replace()`)
* Do something similar to this PR to support 2.56 (I don't have any 2.56 so I can't verify for sure, but could check the new API in 2.58 to manually check).
  * The most annoying piece might be the json-glib change, which might need to either be a patch, or conditionally defining `G_GNUC_FALLTHROUGH` from the build system (on GLib < 2.60)

I know you want the patches against the LSP repo, but I'm starting it here because it's mostly concerning the GP integration. I can PR against that repo when we know what to do -- or you can take whatever you like and make your own, I don't mind :)

### DIFF
--- a/lsp/src/lsp-autocomplete.c
+++ b/lsp/src/lsp-autocomplete.c
@@ -315,6 +315,51 @@ LspAutocompleteSymbol *find_symbol(GeanyDocument *doc, const gchar *text)
 }
 
 
+#if ! GLIB_CHECK_VERSION (2, 68, 0)
+static guint lsp_g_string_replace(GString *string, const gchar *find, const gchar *replace, guint limit)
+{
+	guint replacements = 0;
+	gchar *p = string->str;
+	gsize find_len = strlen (find);
+	gsize replace_len = strlen (replace);
+
+	while ((p = strstr (p, find)))
+	{
+		gsize i = p - string->str;
+
+		if (replace_len <= find_len)
+		{
+			memcpy (p, replace, replace_len);
+			if (replace_len < find_len)
+				g_string_erase (string, i + replace_len, find_len - replace_len);
+		}
+		else
+		{
+			memcpy (p, replace, find_len);
+			g_string_insert_len (string, i + find_len, replace + find_len, replace_len - find_len);
+		}
+
+		replacements++;
+		if (limit && replacements >= limit)
+			break;
+
+		p = string->str + i + replace_len;
+		/* special case for empty search: advance not to match forever */
+		if (! find_len)
+		{
+			if (*p)
+				p++;
+			else
+				break;
+		}
+	}
+
+	return replacements;
+}
+#define g_string_replace lsp_g_string_replace
+#endif
+
+
 void lsp_autocomplete_selection_changed(GeanyDocument *doc, const gchar *text)
 {
 	LspAutocompleteSymbol *sym = find_symbol(doc, text);
--- a/lsp/deps/json-glib/json-scanner.c
+++ b/lsp/deps/json-glib/json-scanner.c
@@ -1078,7 +1078,10 @@ json_scanner_get_token_ll (JsonScanner    *scanner,
               ch = 0;
               break;
             }
+#ifdef G_GNUC_FALLTHROUGH /* GLib >= 2.60 */
             G_GNUC_FALLTHROUGH;
+#endif
+            /* Fallthrough */
 
 	case '0':
 	case '1':
--- a/lsp/src/spawn/lspunixoutputstream.c
+++ b/lsp/src/spawn/lspunixoutputstream.c
@@ -93,7 +93,9 @@ lsp_unix_output_stream_class_init (LspUnixOutputStreamClass *klass)
   gobject_class->set_property = lsp_unix_output_stream_set_property;
 
   stream_class->write_fn = lsp_unix_output_stream_write;
+#if GLIB_CHECK_VERSION (2, 60, 0)
   stream_class->writev_fn = NULL;
+#endif
   stream_class->close_fn = lsp_unix_output_stream_close;
 
    /**


lsp/src/lsp-autocomplete.c
```diff
@@ -315,6 +315,51 @@ LspAutocompleteSymbol *find_symbol(GeanyDocument *doc, const gchar *text)
 }
 
 
+#if ! GLIB_CHECK_VERSION (2, 68, 0)
+static guint lsp_g_string_replace(GString *string, const gchar *find, const gchar *replace, guint limit)
```
b4n, line 319: Yes, I re-implemented that manually out of sheer 1am :laughing: You might want to use a copy of the GLib implementation, although this one should work fine.

techee:
> @techee First 2.1 issue breakage :)

Nice, caused by me ;-)

> Current LSP doesn't build on GLib 2.58 (GP only depend on 2.56 through Geany 2.1).

Out of curiosity - is it something you noticed by yourself or did someone else report this issue? Isn't there some compile-time glib macro we could use to always report issues like this? I've made this mistake several times already and I suspect it's not the last time.

> The most annoying piece might be the json-glib change, which might need to either be a patch, or conditionally defining G_GNUC_FALLTHROUGH from the build system (on GLib < 2.60)

Would it be hard to do using the build system? If it's not a big problem, I'd prefer that to modifying jsonrpc-glib. But if the result is something too complicated, I can e.g. save this diff to a patch file and apply it when updating to a newer version of jsonrpc-glib (which I think is pretty stable and not evolving too much so it might not be necessary to do normally).

> I know you want the patches against the LSP repo, but I'm starting it here because it's mostly concerning the GP integration. I can PR against that repo when we know what to do -- or you can take whatever you like and make your own, I don't mind :)

Don't worry, it doesn't really matter - I have to apply the change to the other repo anyway.

b4n:
> I'm not sure if we have the same thing in mind - I thought about adding `-DG_GNUC_FALLTHROUGH` to gcc arguments on glib < 2.60 so the macro would get defined and it wouldn't require source code modification.

Oh, my bad.  Still untested, but something like this:
```diff
diff --git a/build/lsp.m4 b/build/lsp.m4
index 2aaa6da3..11098887 100644
--- a/build/lsp.m4
+++ b/build/lsp.m4
@@ -12,6 +12,10 @@ AC_DEFUN([GP_CHECK_LSP],
             [Force using system json-glib and jsonrpc-glib libraries for the LSP plugin. [[default=no]]]),,
         enable_system_jsonrpc=no)
 
+    PKG_CHECK_MODULES([GLIB_2_60], [glib-2.0 >= 2.60],
+                      [have_glib_2_60=yes], [have_glib_2_60=no])
+    AM_CONDITIONAL([HAVE_GLIB_2_60], [test "x$have_glib_2_60" = "xyes"])
+
     PKG_CHECK_MODULES([SYSTEM_JSONRPC],
         [${JSON_GLIB_PACKAGE_NAME} >= ${JSON_GLIB_VERSION}
          ${JSONRPC_GLIB_PACKAGE_NAME} >= ${JSONRPC_GLIB_VERSION}],
diff --git a/lsp/deps/Makefile.am b/lsp/deps/Makefile.am
index f3e03c22..c766d705 100644
--- a/lsp/deps/Makefile.am
+++ b/lsp/deps/Makefile.am
@@ -70,6 +70,11 @@ libjsonrpc_la_CPPFLAGS = $(AM_CPPFLAGS) \
 	-I$(top_srcdir)/lsp/deps/jsonrpc-glib
 libjsonrpc_la_CFLAGS = $(AM_CFLAGS)
 
+if ! HAVE_GLIB_2_60
+# compatibility for GLib < 2.60
+libjsonrpc_la_CPPFLAGS += -DG_GNUC_FALLTHROUGH
+endif HAVE_GLIB_2_60
+
 # do not enable cppcheck for json-glib or jsonrpc-glib as it generates lots of
 # false positives and it's not "our" code anyway
 #
```

Or `AM_SUBST` some variable that contain `-DG_GNUC_FALLTHROUGH` from `lsp.m4`, and use it directly in `libjsonrpc_la_CPPFLAGS`, which would be effectively the same.

techee:
>> I'm not sure if we have the same thing in mind - I thought about adding -DG_GNUC_FALLTHROUGH to gcc arguments on glib < 2.60 so the macro would get defined and it wouldn't require source code modification.

> Oh, my bad. Still untested, but something like this: ...

Unfortunately, with `-DGLIB_VERSION_MAX_ALLOWED=GLIB_VERSION_2_56` it produces `warning: Not available before 2.60` on newer glib versions (which is what we told it to do with GLIB_VERSION_MAX_ALLOWED) so I think the only option here is to really patch the library if we want to use `-DGLIB_VERSION_MAX_ALLOWED=GLIB_VERSION_2_56` (which I think is a good idea).

techee:
@b4n I created #1479 for the LSP part. I left `g_uri_to_string_partial()` to you :-P.

I kind of hoped to define `-DGLIB_VERSION_MAX_ALLOWED=GLIB_VERSION_2_56` globally for all plugins and Geany because it would be useful but this produces warnings for functions provided by newer glib versions even if they are guarded by the corresponding `#if GLIB_CHECK_VERSION(...)` check. I'm not sure if there's a solution for this.

b4n:
> I left `g_uri_to_string_partial()` to you :-P.

Hum, actually this looks fine.  Let me explain:
* the whole [`GUri` API](https://docs.gtk.org/glib/struct.Uri.html) is new in 2.66, not just this function
* libsoup3's [soup_message_get_uri()](https://libsoup.gnome.org/libsoup-3.0/method.Message.get_uri.html) returns a `GUri`, so it ought to mean libsoup3 depends on GLib >= 2.66
* the plugin properly depend on libsoup3 already, so cannot be built otherwise.

So I don't see anything else I could do, but add a explicit dependency on GLib 2.66, although it's moot because the libsoup3 one ought to bring it.

Opinions?

techee:
> So I don't see anything else I could do, but add a explicit dependency on GLib 2.66, although it's moot because the libsoup3 one ought to bring it.

But then geany-plugins as a whole should depend on glib 2.66, right?

techee:
Is there actually anything like glib or gtk version check for g-p? There seems to be just gtk 3.x check but I kind of suspect the compilation would fail with some older 3.x versions. There doesn't seem to be any explicit glib check and g-p just use whatever is available.

b4n:
> > So I don't see anything else I could do, but add a explicit dependency on GLib 2.66, although it's moot because the libsoup3 one ought to bring it.
> 
> But then geany-plugins as a whole should depend on glib 2.66, right?

No, genuispate will just not get enabled if you don't have libsoup3

> Is there actually anything like glib or gtk version check for g-p?

I don't think so, it just requires whatever is needed by its Geany dep.

techee:
Closing, replaced by https://github.com/geany/geany-plugins/pull/1479.

PR #1479, opened 2025-07-11
lsp: Ensure glib 2.56 compatibility

techee:
Replaces #1476.

### DIFF
--- a/lsp/deps/json-glib/json-scanner.c
+++ b/lsp/deps/json-glib/json-scanner.c
@@ -38,6 +38,10 @@
 #include <glib.h>
 #include <glib/gprintf.h>
 
+#ifndef G_GNUC_FALLTHROUGH
+# define G_GNUC_FALLTHROUGH  /* GLib < 2.60 */
+#endif
+
 typedef enum
 {
   JSON_ERROR_TYPE_UNKNOWN,
--- a/lsp/src/lsp-autocomplete.c
+++ b/lsp/src/lsp-autocomplete.c
@@ -338,9 +338,9 @@ void lsp_autocomplete_selection_changed(GeanyDocument *doc, const gchar *text)
 
 		g_strstrip(sym->documentation);
 		str = g_string_new(sym->documentation);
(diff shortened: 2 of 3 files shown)


lsp/deps/json-glib/json-scanner.c
```diff
@@ -38,6 +38,10 @@
 #include <glib.h>
 #include <glib/gprintf.h>
 
+#ifndef G_GNUC_FALLTHROUGH
+# define G_GNUC_FALLTHROUGH  /* GLib < 2.60 */
+#endif
```
techee, line 43: FYI: the upstream json-glib uses G_GNUC_FALLTHROUGH at more places in json-scanner.c defining it once for all the uses is a better solution.

techee: > If you do it like that you could prevent the warning with MAX_ALLOWED by undefining the GLib one.

I'm lost - what is MAX_ALLOWED and what warning are you talking about?

> Not that it's necessarily perfect either given you'll get GCC to warn about the implicit fallthroughs…

This will be on older glib versions so I think we don't have to care too much, do we?

b4n: I meant what you mentioned in https://github.com/geany/geany-plugins/pull/1476#issuecomment-3063207264 (although I didn't see that one, so that you were maybe giving up on it).

My point is just that: if you don't want to get a warning when setting `-DGLIB_VERSION_MAX_ALLOWED=GLIB_VERSION_2_56` you probably need to:
```c
#undef G_GNUC_FALLTHROUGH
#define G_GNUC_FALLTHROUGH /* Fallthrough */
```

but then GCC might emit the implicit fallthrough warnings.

b4n (approved):
LGTM

lsp/deps/json-glib/json-scanner.c
techee: Ah, I've just read https://github.com/geany/geany-plugins/pull/1476#issuecomment-3063809906 so forget the above.

Issue #1483, opened 2025-07-16
vimode: f/F/t/T not working with modifier keys on nonstandard keyboard layout


Schyrsivochter:
The NEO 2 German ergonomic keyboard layout ([Wikipedia](https://en.wikipedia.org/wiki/Neo_(keyboard_layout)), [official website (in German)](https://neo-layout.org)) has two modifier keys in addition to Shift – called Mod3 and Mod4 – which give access to four more layers of characters. Layer 3, for instance, contains most punctuation aside from comma and full stop, so it is an absolutely essential feature of the NEO 2 layout.

Using NEO 2 in Geany vimode generally works fine – e.g. there are no problems using the `^`, `$`, `?`, `/`, `:`, `;` commands/command introducers, nor text objects like `a(`.

However, if I try to search for punctuation in the current line using f/F/t/T, the behaviour is strange: it’s as if _the Mod3 keypress_ specifies a character to search for, and the character it finds is:
- the character under the cursor (for backwards search with F/T),
- the character to the right (for forwards search with f/t).

That is, effectively, `f<Mod3>` and `T<Mod3>` perform an `l` or `<Right>`, and `F<Mod3>` and `t<Mod3>` are no-ops.

After Mod3 is pressed, and the cursor has gone one character to the right (or not), any punctuation typed will be interpreted as a new command; e.g. trying to search for `^`, the cursor instead jumps to the start of text in the line; attempting to go to the next `:` opens the ex command popup, etc.

Furthermore, after such an action, the `;` command repeats the motion (or lack thereof); and `,`:
- does nothing if the command was `f`,
- goes right (`l`) if it was `F`,
- goes left (`h`) if it was `t`,
- goes two characters to the right (`2l`) if it was `T`.

The behaviour is identical no matter which of the modifier keys I use, Mod3 or Mod4, left or right.

All this was observed under Ubuntu 24.04.2 LTS, with an X11-based desktop.

```
$ geany --version
geany 2.0 (built on 2024-03-31 with GTK 3.24.41, GLib 2.80.0)
```

Schyrsivochter:
Reporting that `r` also behaves strangely with Mod3 and Mod4 – namely, it toggles the case of the letter under the cursor.

PR #1486, opened 2025-07-16
vimode: Possibly improve behavior on some strange keyboard layouts

techee:
There seems to be a problem when using commands such as f<key> on German Neo keyboard where punctuation characters are accessed using MOD3 keys. This is possibly mapped to GDK_MOD1_MASK or GDK_META_MASK state which leads to ignoring such key events.

Attempt to use gdk_keymap_translate_keyboard_state() which returns potentially consumed modifiers and remove those from ev->state. The hope is that when a visible character is typed using MOD3+key, the modifier corresponding to MOD3 will get reported in consumed and removed so it won't affect further event processing.

Possibly fixes https://github.com/geany/geany-plugins/issues/1483


techee:
@Schyrsivochter Since I don't have the Neo German keyboard, I have no way to test if this fixes the problem or not. I _believe_ the problem here really is keyboard event handling but I'm not sure how `gdk_keymap_translate_keyboard_state()` behaves exactly. And of course, the problem might be somewhere else so it would be nice if you could test this PR.

Schyrsivochter:
> Since I don't have the Neo German keyboard, I have no way to test

The Neo 2 layout is, in fact, in FreeDesktop’s XKeyboardConfig database and thus preinstalled in basically any Linux distro.

> it would be nice if you could test this PR

Cherry-picked on 2.0, builds successfully but no change in behaviour, unfortunately.

Schyrsivochter:
Looks promising. I’ll try it when I’m back at work on Monday.

Schyrsivochter:
Works like a charm. Fixes `r` too. Thank you very much!

PR #1493, opened 2025-07-27
From Test to Release signing workflow

giuspen:
This should give us the geany-plugins 2.1 signed installer.
No further changes should be necessary, only merge and kick the workflow.
Only please double check that this branch was branched correctly before the post release version increase.


giuspen:
Mistake will open again

Issue #1492, opened 2025-07-25
Spell Check plugin alters Geany behaviour with forwards/backwards word delete after being activated


ghost:
By default, pressing Ctrl+Backspace or Ctrl+Delete removes words backwards or forwards in Geany, regardless of whether the previous word is followed/preceded with punctuation characters or not. The command also removes all non letter characters (that includes punctuation and spaces) until it hits the word to be deleted.

After activating/enabling the spell checker, this is no longer the case and the cursor stops at non-word symbols, such as '*', '<', '>'... The default behaviour is not restored after deactivating the plugin.

Only the current document is affected, i.e. the document which is active when the plugin is activated. Closing and reopening the document restores said default behaviour. I have noticed this altered behaviour for as long as I can remember using Geany so it's not new (at least > 5 years old, I would say).

Using Geany 2.0-1 and Geany plugins 2.0-6.
Operating system: Manjaro Linux

PR #1495, opened 2025-07-27
spellcheck: Fix wordchars restoration

b4n:
See #1492.

The second commit can be removed if wanted, it's not currently necessary.


eht16:
Thanks!

LGTM, tested and works.

I'm fine with merging also the second commit.

PR #1498, opened 2025-08-08
commander: settings dialog, project files, symbols, icons

memstream:
Added settings to the plugin configuration dialog, allowing you to control the display of project files, symbols in the command panel and the size of the command panel. Added the ability to recursively search for project files using the installed file pattern. Added the ability to search for symbols. Added display of icons, for files and menus the icons are fixed, for symbols (tags) they are different depending on the type of symbol.

I'm not a C programmer, if someone does a review I'll be grateful. I haven't noticed any problems yet.

dialog_settings
project_files
symbols


memstream:
> I probably won't have time to review in coming week, but I'll try to look at this, thanks.
> 
> One thing that'll I surely ask is whether it could be split into smaller commits, one for each feature. If you don't know how to do that, either learn :) or I'll try to do it when reviewing. But it's usually easiest when doing the changes themselves at first (for the future). Basically to split it, you uncommit (but keep the changes themselves, `git reset --soft @^`) and then commit partial pieces (`git add -p`).

I divided the commit into several others; in addition to new features, there were several fixes, I moved them into separate commits.

memstream:
I moved "file search" to GTask. Because the search was previously in the same thread as gtk, it could block gtk.

memstream:
In addition to fixing leaks and style, I added GtkSpinner (at the bottom of the screenshot), without it it was not clear whether the search was still in progress or had already been completed.

spinner

PR #1533, opened 2026-01-21
GeanyLua: Update to match latest Geany's Scintilla API

Skif-off:


Skif-off:
@xiota, could you look at these changes? I started to suspect that new types should be added to `glspi_scintilla_param`...
(I'm guessing they're numbers, but I'm not sure.)

xiota:
@b4n I think this should be fine to merge, but I can't really test because my local builds are against stable geany.  (I regenerate the symbols before build).

PR #1586, opened 2026-07-03
GeanyLua: Add geany.range.

Skif-off:
Maybe remove
```c
if (start < 0) { return FAIL_UNSIGNED_ARG(1); }
if (stop < 0) { return FAIL_UNSIGNED_ARG(2); }
```
?


Skif-off:
@xiota, @b4n, @eht16, @elextr , Winnie-the-Pooh and all-all-all :)

Scintilla uses `SCI_GETTEXTRANGE` with the `Sci_TextRange` structure, so we have no way to access the text without using the selection functions, but that can be inconvenient in some cases.

PR #1608, opened 2026-09-17
vimode: Don't clamp the cursor while the mouse button is pressed

### LINKED ISSUE #1606: Vim Mode deletes the last character of a line when pressing "a" to go into insert mode and then pushing enter
I have been trying to use Geany with Vim Mode and I keep running into this obnoxious bug where if I press "a" key at the end of a line and then press enter, half the time it will highlight the last character and then delete it when I push enter to create a new line. The expected behavior of pushing "a" should be to not highlight the last character and simply move the cursor to the right of it. I do appreciate that Geany has a vim mode and I hope this issue can be easily fixed to make the plugin more usable.

techee:
In command mode the cursor is clamped onto the last character of the line when a mouse click places it behind the end of line. This is done from the SCN_UPDATEUI notification which is delivered asynchronously and usually arrives while the mouse button is still held. Scintilla then re-applies the position under the pointer as the caret on button release but keeps the anchor we moved, which turns the clamp into a selection of the last character. Entering insert mode with "a" and pressing Enter deletes that character, and mouse selections started behind the end of line lose the last character.

Skip the clamp while Scintilla holds its mouse grab and perform it from a "button-release-event" handler connected after Scintilla's own handler instead. The handler is managed by vi_set_active_sci() so the backends don't need any changes.

Fixes #1606

Claude-Session: https://claude.ai/code/session_012BUmiLaxo5vQVRGYWGEEzc


techee:
The Claude message may seem a bit cryptic so in plain words what the problem was and what the fix does. In the command mode in vim, there's never caret behind the last character, it's ON the character. Scintilla places the caret behind the last character when e.g. clicking behind the end of the line. To avoid this, the code moved the caret one character to the left in this case. This was done in `nt->nmhdr.code == SCN_UPDATEUI && nt->updated == SC_UPDATE_SELECTION` event coming from Scintilla.

What happened, however, is that the Scintilla event happened right at the moment the mouse button was still depressed so the move of the caret to the left caused selection instead of just move.

This patch defers the move of the caret until after the mouse is released. This leads to temporary appearance of the caret behind the end of the line while the mouse is depressed but it's just a small visual problem.

